PR #102, opened 2026-06-01
feat(peer): SwarmPeer + handshake 15.0.0 + hive 2.0.0 (bee-mainnet wire)

mfw78:
## Summary

Cuts vertex's handshake and hive protocols over to the bee-mainnet
wire format, deletes the legacy `SwarmPeer` rust struct (replaced
by the signed peer record below), and replaces every caller with
the new `SwarmPeer` type. Absorbs the scope previously tracked as
#109 (handshake 15.0.0) and #111 (hive 2.0.0).

### Naming

- Rust type is `SwarmPeer` (the old `BzzAddress` name clashed with
  `bzz://` content URIs; `SwarmPeer` lives in the peering context).
  Proto message also renamed to `SwarmPeer` since proto3 keys on
  field numbers, not names.
- Multiaddr set is named `multiaddrs` throughout (Rust struct field,
  accessor, proto field). Bee calls this `underlay`; the rename
  prefers the literal protocol term. Field numbers in the proto are
  unchanged so vertex stays bee-wire-compatible.

### Wire / protocol

- `/swarm/handshake/14.0.0/handshake` -> `/swarm/handshake/15.0.0/handshake`
- `/swarm/hive/1.1.0/peers` -> `/swarm/hive/2.0.0/peers`
- Both protos carry the nested `SwarmPeer` (multiaddrs, signature,
  overlay, nonce, timestamp, chequebook_address).
- Sign-data layout comes from `nectar_primitives::signing::sign_data`;
  this crate owns the libp2p `Vec<Multiaddr>` so nectar stays
  libp2p-free.

### API

- `SwarmPeer::sign` signs with EIP-191 and caches `signer.address()`.
- `SwarmPeer::parse` recovers the signer, recomputes the overlay,
  optionally enforces clock-skew tolerance.
- `ethereum_address()` accessor (no re-recovery on every call).
- Scope helpers (`ip_capability`, `addrs_by_scope`, `has_scope`,
  `max_scope`) collapse the old `SwarmPeer` surface.
- `#[non_exhaustive]` on `SwarmPeer` and the error enum.
- Test-only `SwarmPeer::from_parts` under
  `cfg(any(test, feature = "test-utils"))`.

### Wall-clock timestamp

- `prepare_local_peer` and codec test helpers now call
  `Timestamp::now()` (exposed by nectar_primitives) instead of
  hand-rolling `SystemTime::now().duration_since(UNIX_EPOCH)`.

### Defences against review findings

- Wire malleability: bee's `generateSignData` produces byte-identical
  sign-data for `chequebook = None` and `chequebook = Some(0u20)`.
  Both `sign` and `parse` collapse the zero-address chequebook into
  `None`.
- Timestamp contract: split `TimestampOutsideSkewWindow` into
  `InvalidTimestamp` (structural `<= 0`, bee's `ErrTimestampInvalid`)
  and the existing skew-window variant.

Nectar follow-up for the `PeerRecord` split is tracked at
ethersphere/nectar#43.

## Test plan

- [x] `cargo clippy --lib --all-features -- -D warnings`
- [x] `cargo clippy --tests --benches --all-features -- -D warnings -A clippy::unwrap_used -A clippy::expect_used -A clippy::indexing_slicing`
- [x] `cargo test --workspace --all-features --no-fail-fast` (603 passed, 0 failed)
- [x] `cargo fmt --all --check`
- [ ] bee-interop golden vectors (deferred to a follow-up unit)

Closes #109
Closes #111


mfw78:
## Code review findings

Posting here because the in-skill `--comment` flag was blocked by the orchestration sandbox; findings produced by an automated reviewer pass with default effort.

### 1. `crates/swarm/peers/peer/src/bzz.rs:228` — chequebook wire malleability

`Some(Address::ZERO)` and `None` produce **byte-identical sign-data** (both pad 20 zero bytes per nectar's `sign_data`) but parse to **distinct `BzzAddress` values**. The intentional test at L358 confirms the sign-data equivalence; the consequence is that an on-path attacker can rewrite a peer's wire bytes from "empty chequebook field" to "all-zero chequebook" without breaking signature verification. Downstream code that branches on `addr.chequebook().is_some()` (e.g. "use SWAP if Some, pseudosettle otherwise") will route the rewritten record via SWAP for a peer that never advertised one.

Fix: normalize on parse — treat all-zero chequebook bytes as `None`, or reject non-zero-length chequebook with all-zero content.

### 2. `crates/swarm/peers/peer/src/bzz.rs:135` — `parse()` violates its "opt out skew" contract for `timestamp <= 0`

The docs say `skew: None` opts out of the skew check entirely. The implementation still unconditionally rejects `timestamp <= 0` with `TimestampOutsideSkewWindow` even when skew is `None`. Either tighten the doc to "timestamp must be strictly positive regardless of skew" or move the `<=0` check inside the `skew.is_some()` branch.

### 3. `crates/swarm/peers/peer/src/error.rs:34` — error variant overloaded

`TimestampOutsideSkewWindow` is reused for the distinct "timestamp <= 0" protocol-violation case. Bee uses two sentinels (`ErrTimestampInvalid` vs `ErrTimestampOutsideSkew`). Operator triage is misled: an alert on `TimestampOutsideSkewWindow` reads as a clock-drift / NTP issue when the actual cause is a malformed peer record. Split into `InvalidTimestamp` + `OutsideSkewWindow`.

### 4. `crates/swarm/peers/peer/src/error.rs:21` — adding `#[non_exhaustive]` is source-breaking

Public `SwarmPeerError` gains `#[non_exhaustive]`. Any external (or in-tree) exhaustive `match` arm without a catch-all will fail to compile. Worth a CHANGELOG note and a minor version bump.

### 5. `Cargo.toml:210` — workspace alloy-signer 1.0 → 2.0 bump is out of scope

This PR's title is peer-only, but the workspace `alloy-signer` bump (and the cascade to alloy-signer-local, alloy-primitives 1.6, strum 0.28, alloy-sol-types 1.6) was required by the new nectar pin and now touches every workspace crate that uses alloy-signer (handshake, identity, chequebook, api, test-utils). Mixing wire-protocol semantics with an unrelated dep bump widens the bisect surface for any regression in chequebook signing or identity keystore. Recommend splitting into two PRs, or calling out the bump prominently in the PR title/description/changelog.

PR #103, opened 2026-06-01
feat(net-core): introduce SwarmProtocol trait abstraction

mfw78:
## Summary

Unit 3 of the bee-mainnet bootnode interop plan.

Introduces a uniform `SwarmProtocol` trait so handshake, hive, pingpong and future protocols share one shape. Pure additive surface; no runtime behaviour change.

- New `vertex-swarm-net-core` crate (`crates/swarm/net/core/`) with:
  - `SwarmProtocol` trait (`NAME`, `VERSION`, `STREAM_NAME`, `Codec`, `Message`, `full_protocol_id()`).
  - `SemanticVersion(u16, u16, u16)` with libp2p-compatible matching (same major, server-minor ≥ client-minor).
  - `swarm_protocol_id!(name, version, stream)` compile-time macro composing `/swarm/{name}/{version}/{stream}`.
- Metadata-only impls on the existing handshake, hive, pingpong modules.
- Existing `PROTOCOL` constants are preserved unchanged (verified by tests).

## Test plan

- [x] `cargo test -p vertex-swarm-net-core --no-fail-fast` (8 unit + 1 doctest passing)
- [x] `cargo build -p vertex-swarm-net-{core,handshake,hive,pingpong}`
- [ ] downstream regression once dependent units land

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8109,6 +8109,13 @@ dependencies = [
 ]
 
 [[package]]
+name = "vertex-swarm-net-core"
+version = "0.1.0"
+dependencies = [
+ "libp2p",
+]
+
+[[package]]
 name = "vertex-swarm-net-handler-core"
 version = "0.1.0"
 dependencies = [
@@ -8149,6 +8156,7 @@ dependencies = [
  "vertex-observability",
  "vertex-swarm-api",
  "vertex-swarm-identity",
+ "vertex-swarm-net-core",
  "vertex-swarm-net-proto",
  "vertex-swarm-peer",
  "vertex-swarm-spec",
@@ -8198,6 +8206,7 @@ dependencies = [
  "vertex-net-utils",
  "vertex-observability",
  "vertex-swarm-api",
+ "vertex-swarm-net-core",
  "vertex-swarm-net-handler-core",
  "vertex-swarm-net-headers",
  "vertex-swarm-net-proto",
@@ -8247,6 +8256,7 @@ dependencies = [
  "vertex-net-codec",
  "vertex-net-ratelimiter",
  "vertex-observability",
+ "vertex-swarm-net-core",
  "vertex-swarm-net-handler-core",
  "vertex-swarm-net-headers",
  "vertex-swarm-net-proto",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,6 +33,7 @@ members = [
     "crates/net/utils",
 
     # Swarm Network Protocols (crates/swarm/net/)
+    "crates/swarm/net/core",
     "crates/swarm/net/proto",
     "crates/swarm/net/handler-core",
     "crates/swarm/net/handshake",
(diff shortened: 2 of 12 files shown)


mfw78:
## Code review findings

Posting these here because the in-skill `--comment` flag was blocked by the orchestration sandbox; findings below were produced by an automated reviewer pass with default effort.

### 1. `crates/swarm/net/core/src/protocol.rs:56` — `full_protocol_id()` can drift from `NAME`/`VERSION`/`STREAM_NAME`

The associated `full_protocol_id()` is hand-supplied per impl rather than derived from the trait constants. A future bump of e.g. `Pingpong::VERSION` to `(1,1,0)` that forgets to update the protocol-ID string compiles cleanly but breaks libp2p negotiation. The `Dummy` test impl at L78-80 already hard-codes the full string, demonstrating the divergence the trait was meant to prevent.

Suggest: provide a default body that composes `NAME`/`VERSION`/`STREAM_NAME` and forbid overriding.

### 2. `crates/swarm/net/core/src/version.rs:72` — `swarm_protocol_id!` mishandles typed integer literals

The macro uses `stringify!` on `$major/$minor/$patch :literal`, so a typed literal like `1u16` produces `/swarm/foo/1u16.0.0/foo` — a malformed protocol ID that no longer matches bee's `/swarm/foo/1.0.0/foo`. Restrict the macro fragment to bare integer literals or format the numeric value instead of stringifying the token.

### 3. `crates/swarm/net/core/src/version.rs:42` — `SemanticVersion::matches(client, server)` is asymmetric but reads symmetric

Argument swap silently flips the compat decision. Either rename to `matches(advertised, supported)` so the asymmetry is in the name, or expose a method form `self.is_compatible_with(other)`.

### 4. `crates/swarm/net/core/src/version.rs:13` — `Hash`/`Eq` disagree with `matches()`

Derived `Eq`/`Hash` include patch, but `matches()` ignores it. Using `SemanticVersion` as a `HashMap` key for compatibility lookup silently treats `1.0.0` and `1.0.1` as distinct even though `matches()` calls them compatible — cache misses, repeated negotiation, broken dedup. Either drop patch from `Hash`/`Eq` or add a `CompatKey` newtype that hashes only `(major, minor)`.

mfw78:
Closing as scope-creep. PR adds a SwarmProtocol trait + ProtoCodec marker that nothing in the codebase consumes; the per-protocol PROTOCOL constants are sufficient on their own. The HandshakeCodec/HiveCodec/PingpongCodec marker structs only exist to fill an unused type slot, and HandshakeCodec::Message = HandshakeInfo would be incorrect anyway (HandshakeInfo is the result of a completed handshake, not a wire message; the handshake exchanges Syn/SynAck/Ack). Per the project CLAUDE.md, we should not design for hypothetical future requirements. SemanticVersion and the swarm_protocol_id! macro might earn their keep later, but only when a real consumer (light-node retrieval, GSOC) needs them. Revisit then.

PR #104, opened 2026-06-01
feat(node)!: enforce persistent identity for bootnodes

mfw78:
## Summary

Unit 11 of the bee-mainnet bootnode interop plan.

Bootnodes must have a stable overlay address across restarts — that's the whole contract for a well-known peer. Ephemeral identity is now rejected at build time.

- `SwarmNodeType::Bootnode::requires_persistent_identity() / _nonce()` flipped to `true`.
- New trait method `SwarmIdentityConfig::assert_persistent_identity()`.
- `BootNodeBuilder::build` asserts before constructing the swarm.
- `IdentityArgs` routes to keystore when the node type requires persistence.
- New errors `#[non_exhaustive] IdentityError::EphemeralBootnode`.

## Test plan

- [x] `cargo test -p vertex-swarm-primitives -p vertex-swarm-identity -p vertex-swarm-node -p vertex-swarm-api`
- [x] new tests: `ephemeral_bootnode_is_rejected_at_build`, `bootnode_requires_persistent_identity_and_nonce`

### DIFF
--- a/crates/swarm/api/src/config.rs
+++ b/crates/swarm/api/src/config.rs
@@ -148,9 +148,32 @@ pub trait SwarmNetworkConfig {
 }
 
 /// Configuration for Swarm node identity.
+#[auto_impl::auto_impl(&, Arc, Box)]
 pub trait SwarmIdentityConfig {
     /// Whether to use ephemeral identity (random key, not persisted).
     fn ephemeral(&self) -> bool;
+
+    /// Assert that this configuration satisfies the persistence requirement
+    /// of `node_type`.
+    ///
+    /// Called by node builders (e.g. `BootNodeBuilder`) that require a
+    /// stable overlay address across restarts. Returns
+    /// [`crate::error::IdentityError::EphemeralWhenPersistent`] if `node_type`
+    /// requires persistence (see
+    /// [`crate::SwarmNodeType::requires_persistent_identity`]) but this
+    /// configuration would produce an ephemeral identity.
+    ///
+    /// The default implementation derives the answer from [`Self::ephemeral`].
+    /// Implementations whose persistence semantics differ may override.
+    fn assert_persistent_identity(
+        &self,
+        node_type: crate::SwarmNodeType,
+    ) -> Result<(), crate::error::IdentityError> {
+        if self.ephemeral() && node_type.requires_persistent_identity() {
+            return Err(crate::error::IdentityError::EphemeralWhenPersistent { node_type });
+        }
+        Ok(())
+    }
 }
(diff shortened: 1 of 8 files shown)


mfw78:
## Code review findings

Posting here because the in-skill `--comment` flag was sandbox-blocked; findings produced by an automated reviewer pass with default effort.

### 1. `crates/swarm/node/src/node/storer.rs:165` — `StorerNodeBuilder::build` missing the persistence assertion

`SwarmNodeType::Storer` now reports `requires_persistent_identity() = true`, but `StorerNodeBuilder::build` does not call `self.identity.assert_persistent_identity()?` the way `BootNodeBuilder` does (mirror of `bootnode.rs:228`). A library/test caller that hands in `Identity::random(spec, Storer)` (ephemeral) gets a storer with an ephemeral signing key, breaking staking/reservation contracts. The CLI path catches this via `IdentityArgs::identity`, but the build-time guard is missing for non-CLI callers.

### 2. `crates/swarm/api/src/error.rs:236` — `IdentityError::EphemeralBootnode` is emitted for Storer too

`Identity::load_persistent_or_error` and `IdentityArgs::identity` both return `EphemeralBootnode` even when the failing node type is `Storer`. With `strum::IntoStaticStr`, this serializes to `"ephemeral_bootnode"` in logs/metrics for storer failures, misleading operators. Rename to `EphemeralIdentity` (or split per node type) before `#[non_exhaustive]` ossifies the surface.

### 3. `crates/swarm/identity/src/lib.rs:102` — `Identity::load_persistent_or_error` is dead public API

The method is exported and documented as the canonical persistence check but has zero call sites in the workspace. `BootNodeBuilder` uses `SwarmIdentityConfig::assert_persistent_identity`; `IdentityArgs::identity` open-codes the same check. Either wire it through the builders/args (consolidating the three-place duplicate check) or drop it before it becomes a permanent public API.

### 4. `crates/swarm/identity/src/keystore.rs:24` — empty-password keystore is silently written

Real security concern. Pre-PR, `vertex bootnode` (no password args) produced an ephemeral identity, weak but contained. Post-PR, the same invocation falls through to `resolve_password` returning `String::new()` and `create_and_save_signer` writes a **keystore protected by an empty password** to `network_dir/keystore/swarm`. Operators following the new "persistent by default" UX get a stable overlay but a zero-password keystore on disk — exfiltrating the file is enough to control the bootnode's overlay forever.

Fix: require a non-empty password (or explicit `--insecure-empty-password`) when `node_type.requires_persistent_identity()` and the keystore is being created.

PR #105, opened 2026-06-01
feat(net-hive): per-peer rate limiting and bootnode discard gate

mfw78:
## Summary

- Add bee-compatible per-peer hive rate limiters (refill 1/min, burst 128 = `4 * MaxBins`) for both inbound and outbound peer batches; consumes `len(peers)` tokens per batch to match bee's `Limiter.Allow(addr, count)`.
- Introduce a `HivePeerHandler` trait with two impls: `LearnAndDial` (default, surfaces peers to the topology) and `DiscardSilently` (bootnodes — drops the batch and bumps a metric, mirroring bee's `BootnodeMode` early return in `pkg/hive/hive.go:283-284`).
- Wire `ProtocolBehaviours::new` to pick the impl from `SwarmIdentity::node_type` (`Bootnode → DiscardSilently`, else `LearnAndDial`); outbound broadcasting still runs for bootnodes — only ingestion is gated.
- Add `hive_peers_discarded_total{reason="bootnode_mode" | "rate_limited" | "verifier_rejected"}` so all dropped-peer reasons are observable in one place.

## Implementation notes

- `vertex-net-ratelimiter` gains `KeyedRateLimiter<K>`, `RateLimitExceeded`, and `try_consume_n` / `try_acquire_n` (atomic, no partial consumption — matches Go's `rate.Limiter.AllowN`).
- `HiveRateLimit<K = PeerId>` wraps `KeyedRateLimiter` behind a `parking_lot::Mutex` so the behaviour can hold both inbound and outbound limiters.
- Per-peer buckets are evicted on the final `ConnectionClosed` for a peer, mirroring bee's `disconnect` clearing both limiters (`pkg/hive/hive.go:296-300`).

## Test plan

- [x] `cargo build -p vertex-swarm-net-hive -p vertex-swarm-topology`
- [x] `cargo test -p vertex-swarm-net-hive -p vertex-swarm-topology -p vertex-net-ratelimiter` (all unit tests pass, including new ones covering bee-compat defaults, per-peer independence, bucket eviction, atomic N-token consumption, and inbound/outbound policy dispatch)
- [x] `cargo clippy --lib -p vertex-swarm-net-hive -p vertex-swarm-topology -p vertex-net-ratelimiter -- -D warnings`
- [x] `cargo clippy --tests --benches -p vertex-swarm-net-hive -p vertex-swarm-topology -p vertex-net-ratelimiter -- -D warnings -A clippy::unwrap_used -A clippy::expect_used -A clippy::indexing_slicing` (matches CI config in `.github/workflows/unit.yml`)
- [ ] Cluster e2e deferred to Unit 14 per plan.

Refs: bee `pkg/hive/hive.go:54-55,78-79,150,273,283-284,296-300`, `pkg/swarm/swarm.go:29`, `pkg/ratelimit/ratelimit.go`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7866,6 +7866,10 @@ dependencies = [
 [[package]]
 name = "vertex-net-ratelimiter"
 version = "0.1.0"
+dependencies = [
+ "parking_lot",
+ "thiserror 2.0.18",
+]
 
 [[package]]
 name = "vertex-net-utils"
--- a/crates/net/ratelimiter/Cargo.toml
+++ b/crates/net/ratelimiter/Cargo.toml
@@ -7,5 +7,9 @@ license.workspace = true
 homepage.workspace = true
 repository.workspace = true
 
+[dependencies]
+parking_lot.workspace = true
+thiserror.workspace = true
+
 [lints]
 workspace = true
--- a/crates/net/ratelimiter/src/lib.rs
+++ b/crates/net/ratelimiter/src/lib.rs
@@ -1,89 +1,302 @@
-//! Token bucket rate limiter for per-connection stream acceptance.
+//! GCRA token-bucket rate limiting, with a single-bucket form and a per-key
+//! sharded form intended for use across libp2p protocols.
+//!
+//! # Algorithm
+//!
+//! Both [`RateLimiter`] and [`KeyedRateLimiter`] implement the Generic Cell
+//! Rate Algorithm (GCRA). A request asking for `n` tokens is accepted iff
+//! enough time has passed since the bucket's theoretical arrival time (TAT)
+//! to permit `n` token-replenishments. Each bucket needs just one timestamp
+//! (`u64` nanoseconds since the limiter was constructed), so memory per key
+//! is small and a check is a handful of integer ops.
+//!
+//! # Single vs keyed
+//!
+//! - [`RateLimiter`] is one bucket, [`RateLimiter::try_consume_n`] takes
+//!   `&mut self`. Use it when a single owner needs throttling (for example
+//!   a libp2p [`ConnectionHandler`] limiting substream-open rate on the one
+//!   connection it manages).
+//! - [`KeyedRateLimiter`] is `&self`-shareable via an internal mutex and
+//!   maintains an independent bucket per key. Use it when many call sites
+//!   (typically a `NetworkBehaviour` plus the per-connection handler readers
+//!   it spawned) need to charge the same per-peer quota. Disconnect handlers
+//!   should call [`KeyedRateLimiter::clear`] to release the bucket; otherwise
+//!   memory grows with the count of distinct peers seen.
+//!
+//! [`ConnectionHandler`]: https://docs.rs/libp2p/0.56/libp2p/swarm/trait.ConnectionHandler.html
 
+use parking_lot::Mutex;
+use std::collections::HashMap;
+use std::hash::Hash;
+use std::num::NonZeroU32;
 use std::time::{Duration, Instant};
 
-/// Token bucket rate limiter for inbound stream acceptance.
+/// Why a charge against a rate-limited bucket was refused.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, thiserror::Error)]
+pub enum RateLimitedErr {
+    /// The request asks for more tokens than the bucket's burst, so it can
+    /// never be accepted - the caller must reduce the cost or fail
+    /// permanently.
+    #[error("rate limit cost exceeds bucket capacity")]
+    TooLarge,
+    /// The bucket cannot satisfy the request right now; the wrapped duration
+    /// is the earliest moment a retry would succeed.
+    #[error("rate limit exceeded, retry after {0:?}")]
+    TooSoon(Duration),
+}
+
+/// A user-friendly quota: at most `max_tokens` may be consumed in any window
+/// of `replenish_all_every`.
 ///
-/// Prevents peers from rapidly cycling streams even when each stream completes
-/// quickly. Tokens replenish at a steady rate up to a configurable burst limit.
-/// Check this *before* any concurrent-stream bound (e.g. `FuturesSet::try_push`)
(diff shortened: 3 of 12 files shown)


mfw78:
(self-deleted: test)

mfw78 (commented):
(test review)

mfw78 (commented):
Automated code review (high effort). 5 findings posted via separate gh pr comment calls because inline-comment APIs are denied in this sandbox.

mfw78:
**Finding 1/5 — HIGH: DoS regression — rate-limit applied AFTER ECDSA validation**

Location: `crates/swarm/net/hive/src/protocol.rs:89` (`HiveInner::read`, the `validate_batch_blocking` call) and `crates/swarm/net/hive/src/behaviour.rs:180` (where `inbound_limit.try_consume_n` now lives).

`validate_batch_blocking` performs full ECDSA signature recovery on every raw peer in `proto.peers` BEFORE any rate-limit is consulted. The new limiter lives in `HiveBehaviour::on_connection_handler_event`, which only fires AFTER `HiveHandlerEvent::PeersReceived` is emitted by the handler.

Bee `pkg/hive/hive.go:273` calls `s.inLimiter.Allow(...)` BEFORE any per-peer processing precisely so a single peer cannot force unbounded CPU work on signature recovery by spamming oversized batches. With the current ordering, an attacker on one hive substream can send 32 KiB batches (~100 peers each, framed by `MAX_MESSAGE_SIZE`) and force ~100 ECDSA recoveries per batch as fast as they can stream them; the new limiter only suppresses downstream forwarding, not the CPU burn.

Failure scenario: malicious peer keeps one substream open and pushes back-to-back 32 KiB batches of fabricated peers with garbage signatures → CPU stays pinned doing `SwarmPeer::from_signed` on the blocking pool even though every batch is dropped post-validation.

Fix: charge `try_consume_n(peer_id, raw_peers.len())` inside `HiveInner::read` BEFORE `validate_batch_blocking`; drop the duplicate charge at the behaviour layer (or keep it only as a safety net).

mfw78:
**Finding 2/5 — MEDIUM: Inbound rate-limit charges only validated peer count, not raw wire count**

Location: `crates/swarm/net/hive/src/behaviour.rs:170-189`.

`HiveHandlerEvent::PeersReceived(peers)` delivers the POST-validation peer set — `validate_batch_blocking` in `crates/swarm/net/hive/src/protocol.rs:115-154` drops invalid entries via `filter_map`. The cost handed to `try_consume_n` is therefore `validated_count`, but bee `pkg/hive/hive.go:273` charges `len(peersReq.Peers)` — the RAW wire count.

Failure scenario: attacker sends batches of 128 raw peers with 127 invalid signatures. Validated count = 1, so each batch only costs 1 token; attacker sustains ~128× bee's intended ingestion rate per peer while still triggering full ECDSA cost (compounds finding #1).

Fix: propagate the raw wire count out of `HiveInner::read` (e.g. add `raw_count: usize` to `ValidatedPeers`) and charge that; cleanest if folded into the fix for #1 by moving the limiter into the protocol layer where the raw count is in scope.

mfw78:
**Finding 3/5 — MEDIUM: Outbound limiter silently throttles legitimate gossip cycles**

Location: `crates/swarm/net/hive/src/behaviour.rs:96-114` together with `crates/swarm/topology/src/behaviour.rs:493-506` (`broadcast_peers` chunks by `MAX_BATCH_SIZE = 30`).

Each chunk consumes `peer_count` tokens against a 128-token bucket refilling at 1/min. Under normal gossip churn the topology may call `broadcast_peers` multiple times per neighbor between ticks (neighborhood change, depth change, periodic refresh in `gossip/tasks.rs`), each chunked into 30-peer batches. After a few cycles the bucket is exhausted and subsequent broadcasts are dropped SILENTLY — only a `debug!` log + counter increment — for minutes of refill.

Failure scenario: a healthy node that just completed a depth change tries to refresh several neighbors in quick succession → second/third broadcast per neighbor silently dropped; downstream peers miss neighborhood updates and the operator only sees `hive_peers_discarded_total{reason=\"rate_limited\"}` climb with no actionable warning. Identical signal to "hostile peer flooding outbound".

Fix: emit at `warn!` rather than `debug!` for the outbound drop path, and/or expose a saturation gauge per peer so operators can tell hostile activity apart from a starved gossip cycle.

mfw78:
**Finding 4/5 — LOW: `bootnode_mode` discard counter is silent precisely when a bootnode is being flooded**

Location: `crates/swarm/net/hive/src/behaviour.rs:180-204`.

For bootnodes, the order is: charge inbound limiter → if exhausted, increment `hive_peers_discarded_total{reason=\"rate_limited\"}` and return. Only batches that pass the limiter then invoke `on_inbound_peers`, which (for `DiscardSilently`) bumps `hive_peers_discarded_total{reason=\"bootnode_mode\"}`. This matches bee's ordering, but the operator-visible side effect is that a bootnode being SPAMMED past its limit reports `rate_limited`, not `bootnode_mode` — making it hard to alert specifically on "bootnodes getting flooded".

Failure scenario: an alert keyed on `bootnode_mode` stays silent during the exact attack the metric was added to surface.

Fix: bump both reasons for bootnodes when rate-limited, or document the precedence explicitly in `crates/swarm/net/hive/src/metrics.rs` so operators understand what `bootnode_mode` does and does not cover.

mfw78:
**Finding 5/5 — LOW: `peer_count as u64` casts are clippy-clean today but fragile**

Locations: `crates/swarm/net/hive/src/behaviour.rs:108`, `:188`; `crates/swarm/net/hive/src/peer_handler.rs:79`.

`peer_count as u64` and `peers.len() as u64` are lossless on 64-bit `usize` but truncating on 32-bit targets (currently allowed under the `-A clippy::cast_possible_truncation` flag used in the PR test plan). With `MAX_BATCH_SIZE = 30` this is cosmetic, but `u64::try_from(peer_count).unwrap_or(u64::MAX)` (or `u64::from(peer_count as u32)` after a try_from) would be more defensible if the lint policy is later tightened.

Failure scenario: none in practice; hygiene only.

PR #106, opened 2026-06-01
feat(topology): per-peer reachability tracker (AutoNAT + pingpong bridge)

mfw78:
## Summary
- New `PeerReachability { Unknown, Public, Private }` and thread-safe `ReachabilityTracker` that fuses AutoNAT, handshake, and stabilization (Unit 9) signals into a single per-peer verdict. Mirrors bee's `p2p.ReachabilityStatus` and `kademlia.UpdatePeerHealth`.
- Wires the tracker through `LocalAddressManager` so AutoNAT integration has a single entry point (`on_autonat_event`), and exposes `reachability()` / `on_autonat_event` / `on_stabilization` on `TopologyBehaviour`. Handshake handlers in `protocol_handlers.rs` update the tracker BEFORE `routing.connected` so Unit 8's eviction policy sees the new peer as `Public` when `trim_overpopulated_bins` runs.
- Adds the consumer surface (`counts_toward_saturation`, `prefer_survivor`) that Unit 8 will call later, without touching topology routing — Unit 8 can land independently.

## Design notes
- AutoNAT `OutboundProbe::Response` and `InboundProbe::Response` are the only AutoNAT variants that promote a peer to `Public`; failures do not demote because they can stem from our own NAT rather than the server's reachability.
- Handshake failures accumulate; the peer flips to `Private` only after `HANDSHAKE_FAILURE_THRESHOLD = 3` consecutive failures, and any success resets the counter.
- Stabilization-locked peers ignore handshake signals until explicitly unlocked. `update_from_stabilization(peer, false)` is a no-op for untracked peers so a spurious negative cannot blacklist a peer we have no evidence about.

## Test plan
- [x] `cargo build -p vertex-swarm-topology`
- [x] `cargo test  -p vertex-swarm-topology --no-fail-fast` (113 tests, including 16 new reachability tests covering default, handshake success/failure, threshold flip, counter reset, stabilization lock, stabilization loss (tracked and untracked), AutoNAT promotion, ranking, eviction surface, shared-state clone, last-updated semantics)
- [x] `cargo clippy --lib -p vertex-swarm-topology -- -D warnings`
- [x] `cargo clippy --tests --benches -p vertex-swarm-topology --all-features -- -D warnings -A clippy::unwrap_used -A clippy::expect_used -A clippy::indexing_slicing`

### DIFF
--- a/crates/swarm/topology/src/behaviour.rs
+++ b/crates/swarm/topology/src/behaviour.rs
@@ -406,6 +406,35 @@ impl<I: SwarmIdentity + Clone> TopologyBehaviour<I> {
         self.nat_discovery.on_observed_addr(addr);
     }
 
+    /// Forward an AutoNAT event into the per-peer reachability tracker.
+    ///
+    /// Wire this from whichever layer owns the `libp2p::autonat::Behaviour`
+    /// (the swarm-builder, typically). `OutboundProbe::Response` and
+    /// `InboundProbe::Response` promote the involved peer to
+    /// [`crate::PeerReachability::Public`]; other variants are ignored
+    /// (see [`crate::ReachabilityTracker::update_from_autonat`]).
+    pub fn on_autonat_event(&self, event: &libp2p::autonat::Event) {
+        self.nat_discovery.on_autonat_event(event);
+    }
+
+    /// Notify the reachability tracker that the stabilization detector
+    /// (Unit 9) has classified `peer` with `stable`. See
+    /// [`crate::ReachabilityTracker::update_from_stabilization`].
+    pub fn on_stabilization(&self, peer: PeerId, stable: bool) {
+        self.nat_discovery
+            .reachability()
+            .update_from_stabilization(peer, stable);
+    }
+
+    /// Shared per-peer reachability tracker.
+    ///
+    /// Cheap to clone (Arc inside). Exposed so the routing layer (Unit 8) can
+    /// consult per-peer reachability without taking a hard dependency on the
+    /// topology behaviour internals.
+    pub fn reachability(&self) -> crate::ReachabilityTracker {
+        self.nat_discovery.reachability()
+    }
+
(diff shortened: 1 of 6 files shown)


mfw78 (commented):
test inline

mfw78 (commented):
code-review findings (high effort)

mfw78 (commented):
## code-review findings (high effort)

Inline-comment API was sandbox-blocked, posting consolidated review. Ranked by severity.

1. Stabilization-loss demotes any tracked peer (reachability.rs:262-285). The else-if at line 274 fires for every existing record, not just stable=true peers. Spurious update_from_stabilization(P, false) on a peer made Public via handshake/AutoNAT flips it to Private, violating the file-level invariant that a spurious negative cannot blacklist a peer. Add an entry.stable guard.

2. on_handshake_failed counts ConnectionClosed/Timeout as peer-fault (protocol_handlers.rs:266-268). HandshakeError::ConnectionClosed fires on our own CloseConnection (duplicate-conn replacement, banned-peer reject from remote side, shutdown). Three churn races flip an innocent peer to Private. Filter to protocol-fault variants (NetworkIdMismatch, InvalidPeer, InvalidOverlay, InvalidSignature).

3. Unbounded growth — forget() is wired to nothing (reachability.rs + protocol_handlers.rs). Records are created on every handshake event and never pruned. Long-running node accumulates entries for transient/scanner peers indefinitely. Wire forget() into disconnect handler.

4. Failure counter has no time decay (reachability.rs:230-247). Saturating_add reset only by success/AutoNAT/stabilization. Three failures spaced over weeks still flip a peer to Private with no recent evidence. Add a first-failure timestamp + decay window.

Notes (not findings): prefer_survivor tie order = first argument (documented). update_from_autonat correctly ignores Error variants. No test covers finding #1.

Generated by /code-review --comment (high effort).

PR #107, opened 2026-06-01
refactor(topology): typed proximity, bin and neighborhood depth

mfw78:
Unit 13. Replaces bare-u8 PO/bin math with typed newtypes; `NeighborhoodDepth::recompute` mirrors bee `kademlia.go:896-920`. 215 tests pass.


mfw78:
## Code review (default effort, recall-biased)

Sandbox blocked the inline-comment API (`POST /repos/.../pulls/{n}/comments` and `POST /repos/.../pulls/{n}/reviews` with comments). Posting findings here.

### 1. `crates/swarm/peers/peer-manager/src/neighborhood_depth.rs:120-126` (test `single_shallow_bin_above_watermark`)

With `counts[0]=5, saturation=4, watermark=4`, the nectar algorithm picks `candidate=1` (bin 1 is the shallowest unsaturated). The deep-to-shallow watermark walk then short-circuits at `if idx < candidate { break; }`, so bin 0 is never summed and `depth` stays at `candidate=1`. The comment ("Watermark walk from deepest backs up to bin 0... Depth = 0") and the assertion `d.get() == 0` contradict the documented `recompute_neighborhood_depth` body shipped at the pinned `nectar` rev `3c99edc4`. Either the assertion is wrong (should be `1`) or the algorithm at that rev differs from what the local nectar source shows.

### 2. `crates/swarm/topology/src/kademlia/routing.rs:156-170` (`recalc_neighborhood_depth` vs `recalc_depth`)

The two functions are not interchangeable. `recalc_depth` (line 147) returns the deepest bin meeting `nominal()`; `recalc_neighborhood_depth` uses bee's `recalcDepth`: shallowest unsaturated bin, anchored by a deep-to-shallow watermark tail sum. They return different values for almost any non-trivial routing table. The doc claim "consistent with the existing per-bin saturation check" is misleading and a future migration risk. Either drop the new helper until callers move, or document the semantic difference loudly.

### 3. `crates/swarm/topology/src/kademlia/routing.rs:1262-1267` (silent length-mismatch assumption)

```rust
let mut counts = [0usize; NUM_BINS];
for (slot, size) in counts.iter_mut().zip(self.connected_peers.bin_sizes().into_iter()) {
    *slot = size;
}
```

`bin_sizes()` returns a `Vec` of length `max_po + 1`. If `max_po < 31` (any non-default `SwarmSpec`), only a prefix of `counts` is populated; if it were ever `> 31`, deeper bins would be silently truncated. Works today (Swarm default `max_po == MAX_PO == 31`) but the assumption is undocumented. Suggest a `debug_assert_eq!(self.connected_peers.bin_sizes().len(), NUM_BINS)` or an explicit comment pinning the precondition.

### 4. `crates/swarm/peers/peer-manager/src/proximity_index.rs:253-263` (`bin_sizes_typed` silent collapse)

`let raw = idx.min(MAX_BIN as usize) as u8;` followed by `out.insert(bin, size)` will silently overwrite the `Bin::MAX` entry if multiple `idx > MAX_BIN` values ever appear (impossible today because `ProximityIndex::new` constrains `bin_counts.len() == max_po + 1 <= 32`, but this is a footgun if max_po construction ever changes). Same applies to `iter_by_bin`/`iter_by_bin_desc`'s `Bin::new(po).unwrap_or(Bin::MAX)` (lines 1047, 1056) - any rogue `po > 31` is silently coerced to `Bin::MAX` instead of being surfaced or skipped.

---

PR URL: https://github.com/nxm-rs/vertex/pull/107

mfw78 (commented):
Code review (default effort, recall-biased) - see top-level comment with findings.

mfw78:
Closing as superseded by what already landed.

The nectar pin (3c99edc) now exposes:
- `nectar_primitives::Bin` (typed kademlia bin index)
- `nectar_primitives::ProximityOrder` (typed proximity metric)
- `nectar_primitives::recompute_neighborhood_depth(&[u8; 32], saturation, low_watermark) -> Bin` (port of bee's recalcDepth)

The doc on nectar's `neighborhood_depth.rs` is explicit: 'The wrapper type (`NeighborhoodDepth`) stays in the routing layer of each downstream impl; nectar only owns the math.'

#107 was written before those types existed and duplicates them rather than consuming them. Rebasing it onto main creates 8 conflict files (~3000 lines) for what amounts to redoing work that already landed. Filing a focused follow-up issue for the work that genuinely remains: wire nectar's typed Bin / NeighborhoodDepth math into vertex's ProximityIndex and routing.

PR #108, opened 2026-06-01
feat(node-bootnode): wire pricing stub for bee-mainnet interop

mfw78:
Unit 15 — the **critical** bee-mainnet interop requirement. Without it bee disconnects every vertex bootnode after handshake (libp2p.go:697-702/1191-1198: any ConnectIn/Out failure triggers Disconnect; pricing init opens a stream). Stub responds to AnnouncePaymentThreshold + announces ours outbound. 11 tests pass.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8351,8 +8351,10 @@ version = "0.1.0"
 dependencies = [
  "alloy-primitives",
  "asynchronous-codec",
+ "auto_impl",
  "bytes",
  "futures",
+ "futures-bounded",
  "libp2p",
  "quick-protobuf",
  "quick-protobuf-codec",
--- a/crates/swarm/net/pricing/Cargo.toml
+++ b/crates/swarm/net/pricing/Cargo.toml
@@ -18,6 +18,7 @@ vertex-swarm-net-proto.workspace = true
 
 ## async
 futures.workspace = true
+futures-bounded = "0.2"
 
 ## p2p
 asynchronous-codec.workspace = true
@@ -29,6 +30,7 @@ libp2p.workspace = true
 alloy-primitives.workspace = true
 
 ## misc
+auto_impl.workspace = true
 bytes.workspace = true
 strum.workspace = true
 thiserror.workspace = true
--- a/crates/swarm/net/pricing/src/behaviour.rs
+++ b/crates/swarm/net/pricing/src/behaviour.rs
@@ -1,0 +1,281 @@
+//! `NetworkBehaviour` for the pricing protocol.
+//!
+//! Two orthogonal axes of configuration:
+//!
+//! * **Listen-only vs. announce-on-connect** ([`PricingRole`]). Bootnodes
+//!   advertise the protocol but never dial a stream of their own; clients
+//!   and full nodes both advertise AND queue an announce on every new
+//!   connection. The interop requirement is satisfied as soon as we accept
+//!   inbound streams, so the announce side is only needed when WE want our
+//!   threshold tracked by the remote (i.e. when we participate in chunk
+//!   accounting).
+//! * **Observer for inbound thresholds** ([`PricingMode`]). Stub-mode
+//!   discards them; full mode forwards them to a
+//!   [`PaymentThresholdObserver`] that hooks into the accounting subsystem.
+
+use std::{
+    collections::VecDeque,
+    sync::Arc,
+    task::{Context, Poll},
+};
+
+use libp2p::{
+    Multiaddr, PeerId,
+    core::Endpoint,
+    swarm::{
+        ConnectionClosed, ConnectionDenied, ConnectionId, FromSwarm, NetworkBehaviour,
+        NotifyHandler, THandler, THandlerInEvent, THandlerOutEvent, ToSwarm,
+    },
+};
+use tracing::{debug, warn};
+
+use crate::{
+    AnnouncePaymentThreshold,
+    handler::{PricingHandler, PricingHandlerCommand, PricingHandlerEvent},
+    stub::{PaymentThresholdObserver, StubObserver, stub_announcement},
+};
+
+/// Maximum pending swarm events before we start dropping with a warning.
+const MAX_PENDING_EVENTS: usize = 1024;
+
+/// Behaviour role: whether we open an outbound stream to announce our own
+/// threshold on every new connection. Bootnodes use [`Self::ListenOnly`];
+/// nodes participating in chunk accounting use [`Self::Announcer`].
+#[derive(Debug, Clone, Default)]
+#[non_exhaustive]
+pub enum PricingRole {
+    /// Listen-only: advertise the protocol so the remote's `ConnectIn`/`ConnectOut`
+    /// succeeds, accept inbound threshold announcements, never dial an
+    /// outbound stream of our own. Used by bootnodes.
+    #[default]
+    ListenOnly,
+    /// Announce-on-connect: queue an outbound `AnnouncePaymentThreshold` to
+    /// the remote on every new connection. Used by clients and full nodes
+    /// that participate in chunk accounting.
+    Announcer(AnnouncePaymentThreshold),
+}
+
+/// How received thresholds are routed.
+///
+/// `Stub` is the default for bootnodes — incoming thresholds are observed
+/// (so we still emit a `PricingEvent::ThresholdReceived` for visibility) but
+/// not forwarded to any accounting subsystem. `Full` accepts an
+/// `Arc<dyn PaymentThresholdObserver>` that forwards thresholds into
+/// accounting.
+#[derive(Default)]
+#[non_exhaustive]
+pub enum PricingMode {
+    /// Stub mode: observe inbound thresholds but do not feed an accounting
+    /// subsystem.
+    #[default]
+    Stub,
+    /// Full mode: forward inbound thresholds to the supplied observer.
+    Full(Arc<dyn PaymentThresholdObserver>),
+}
+
+impl std::fmt::Debug for PricingMode {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            Self::Stub => f.write_str("Stub"),
+            Self::Full(_) => f.write_str("Full(<observer>)"),
+        }
+    }
+}
+
+/// Events emitted by [`PricingBehaviour`].
+#[derive(Debug)]
+#[non_exhaustive]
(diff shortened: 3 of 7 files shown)


mfw78 (commented):
test post permission check

mfw78 (commented):
## Automated code review (default effort, recall-biased)

Inline comments could not be posted (POST to the GitHub reviews/comments API is sandboxed in this environment); findings are consolidated below with `file:line` anchors so they can be addressed in-place. All findings focus on real correctness / bee-mainnet interop risks for the new pricing stub.

---

### 1. `crates/swarm/node/src/node/bootnode.rs:34` — Hardcoded threshold is a latent interop trap

`BOOTNODE_DEFAULT_PAYMENT_THRESHOLD: u64 = 13_500_000` is a literal that is not tied to any constant in bee. Bee's pricing handler rejects (and disconnects via `p2p.NewDisconnectError(ErrThresholdTooLow)`) any announced threshold strictly less than its `minPaymentThreshold = 2 * refreshRate = 9_000_000` (`bee/pkg/node/node.go:201, 209, 994`). Today's value clears that floor with ~50% margin, but:

- The dependency on the bee constant is not pinned by code, comment, or test.
- If anyone tunes vertex's value *down* without re-reading bee, every bootnode silently becomes undeployable on mainnet.
- The doc comment claims it "matches the bee `lightNodePaymentThreshold` order of magnitude" — but `lightPaymentThreshold = paymentThreshold / 10` is operator-configured on the bee side, not a constant. The comment is misleading.

Recommendation: make the value either configurable (so operators can match local bee settings) or — minimum — add a `#[test]` that asserts `>= BEE_MIN_PAYMENT_THRESHOLD` against a constant that explicitly documents the bee source location.

---

### 2. `crates/swarm/net/pricing/src/behaviour.rs:119` — Silent-drop of the critical announce event

`push_event` discards events with only a `warn!` once the queue reaches `MAX_PENDING_EVENTS = 1024`. The events that flow through here include the per-connection `NotifyHandler::Announce` produced by `announce_to` — the single behaviour-level action this PR exists to perform.

Failure scenario: a connection burst (or a slow swarm `poll` cycle) fills the deque; new `handle_established_*_connection` calls invoke `announce_to`, which hits the cap and silently drops the `NotifyHandler` for the freshly-established connection. The peer's handler never gets the announce command, our outbound stream is never opened. On the bee side `ConnectOut` to *us* still succeeds, but their accounting never sees our threshold. We have a `warn!`, no metric, and no `PricingEvent` — operators have no signal that the interop guarantee silently broke for that peer.

Recommendation: emit a `PricingEvent::AnnouncementDropped { peer }` and/or bump a counter when the queue overflows. Consider treating an announce-drop as a hard error — this queue should effectively never fill.

---

### 3. `crates/swarm/net/pricing/src/handler.rs:116` — Potential lost wakeup: `inbound.poll_unpin(cx)` skipped while `pending_events` drains

`poll` ordering is `pending_events` → `inbound.poll_unpin(cx)` → outbound substream request. When `pending_events.pop_front()` returns `Some`, we return `Ready` *without* touching `self.inbound`, so the inbound `FuturesSet`'s waker is not re-registered with the current `cx` for that poll cycle.

Failure scenario: the handler emits a couple of `AnnouncementSent`/`OutboundError` events that pile into `pending_events`. The swarm drains them one-by-one (each call returns `Ready` after popping one event), and during that span an in-flight `Inbound::new(PricingInner).upgrade_inbound(...)` future completes. Because `inbound.poll_unpin(cx)` was never called with the current waker, completion does not schedule a wakeup. The handler stays `Pending` until the next external event poke; on a connection where pricing is the only active protocol, no other event arrives and the threshold is delivered late or not at all.

Recommendation: always call `self.inbound.poll_unpin(cx)` (and the outbound branch) on every `poll`, push their results into `pending_events`, then return whatever sits at the head.

---

### 4. `crates/swarm/net/pricing/src/handler.rs:99` — `connection_keep_alive` returns `true` unconditionally

Bee's pricing protocol is one-shot per connection: single `AnnouncePaymentThreshold`, then stream close (`bee/pkg/pricing/pricing.go:85-95`). Once we've received their announcement and sent ours, this handler has no further work — yet it perpetually votes "keep this connection alive".

Failure scenario: a peer that has finished pricing, pingpong, hive, and has no open streams is still held open by the pricing handler alone. The connection cannot age out via `connection_keep_alive == false` consensus across all behaviours; only an explicit dial-out replacement or transport-level error will close it. The current code comment ("bee disconnects peers whose pricing exchange never completes") is about the *initial* exchange — once complete, indefinite keep-alive serves no protocol purpose.

Recommendation: return `true` only while `outbound_in_flight || !pending_outbound.is_empty() || !inbound.is_empty()`. Let topology own the long-term keep-alive decision.

---

### 5. `crates/swarm/net/pricing/src/handler.rs:16` — `MAX_PENDING_OUTBOUND = 2` is tight enough to silently drop the announce

With `outbound_in_flight = true` (the first DialUpgrade is in flight) and additional `Announce` commands arriving (e.g. re-negotiation, a future protocol-config update path, or two simultaneous connections to the same peer multiplexed onto one handler), the third `Announce` is dropped with only `warn!` and that peer never gets our threshold.

A cap of 2 is tight for the queue that carries the one event justifying this PR. Recommendation: raise to e.g. 16, *or* surface drops as a `PricingHandlerEvent::OutboundDropped` so the behaviour can re-queue / a metric can fire.

---

### 6. `crates/swarm/net/pricing/src/handler.rs:213` — `ListenUpgradeError` silently swallowed

The `on_connection_event` match has `_ => {}` as its catch-all, which absorbs `ConnectionEvent::ListenUpgradeError` (inbound upgrade failure). When bee opens an inbound pricing stream to us and our `Inbound::new(PricingInner).upgrade_inbound(...)` fails on the libp2p-side negotiation/timeout *before* `try_push`, neither a metric nor a `PricingEvent::InboundError` is emitted.

For an interop stub this is exactly the case operators need visibility into. Recommendation: add an explicit `ConnectionEvent::ListenUpgradeError(err)` arm that pushes an `InboundError` event; mirror the `pingpong` handler's pattern via `UpgradeError::record_and_convert`.

---

### 7. `crates/swarm/node/src/node/bootnode.rs:340` — Test asserts a hardcoded string literal, not the constant

```rust
assert!(
    protocols.contains(&"/swarm/pricing/1.0.0/pricing"),
    "bootnode must advertise pricing protocol; got {protocols:?}"
);
```

If `vertex_swarm_net_pricing::PROTOCOL_NAME` is ever renamed/bumped (e.g. `1.0.0` → `1.1.0`), this test still passes against the old literal even though the bootnode no longer advertises that protocol. Recommendation: assert against `vertex_swarm_net_pricing::PROTOCOL_NAME`. The same applies to `supported_protocols()` itself — return `&[PROTOCOL_NAME]` rather than duplicating the literal.

---

*Findings ranked most-severe first. 7 total: 1 hardcoded-constant interop trap, 4 silent-drop / lost-wakeup concurrency bugs, 1 keep-alive semantics issue, 1 test brittleness.*

mfw78:
**F1 (hardcoded payment threshold): moot for bootnodes.**

Bootnodes are now `PricingRole::ListenOnly` and never announce an outbound threshold, so the constant is gone entirely. Clients and full nodes will pass their threshold through the new `PricingBehaviour::new_announcer(threshold, mode)` constructor when accounting work lands; the value will be sourced from the accounting subsystem rather than a hardcoded literal. See c8062f1.

mfw78:
**F2 (silent drop of the critical announce event): addressed.**

The handler now emits an explicit `PricingHandlerEvent::OutboundDropped` when the per-handler `pending_outbound` queue overflows; the behaviour surfaces this as `PricingEvent::AnnouncementDropped { peer }`. Operators see when the announce path silently failed to deliver our threshold to a peer. See c8062f1.

mfw78:
**F3 (lost wakeup in handler::poll): addressed.**

`PricingHandler::poll` now drains the inbound `FuturesSet` unconditionally on every call (pushing results into `pending_events`), so its waker is always re-registered with the current context. A Ready return from an earlier branch no longer leaves a finished inbound stream un-noticed until some other event pokes the handler. See c8062f1.

mfw78:
**F4 (unconditional connection_keep_alive): addressed.**

`PricingHandler::connection_keep_alive` now returns `true` only while pricing has actual work: `outbound_in_flight`, queued outbound, queued events, or live inbound. Once pricing has nothing pending, topology owns the long-term keep-alive decision. See c8062f1.

mfw78:
**F5 (tight MAX_PENDING_OUTBOUND): addressed.**

`MAX_PENDING_OUTBOUND` bumped from 2 to 16 with a documentation comment explaining the sizing relative to steady-state one-announce-per-connection. The new `OutboundDropped` event (see F2) closes the visibility gap when the cap IS hit. See c8062f1.

mfw78:
**F6 (swallowed ListenUpgradeError): addressed.**

`on_connection_event` now has an explicit `ConnectionEvent::ListenUpgradeError(...)` arm that pushes a `PricingHandlerEvent::InboundError` and logs at warn, mirroring the pingpong handler's pattern. See c8062f1.

mfw78:
**F7 (test asserts a hardcoded protocol string): addressed.**

`supported_protocols_includes_pricing` now asserts against `vertex_swarm_net_pricing::PROTOCOL_NAME`, so a future bump to `1.1.0` (or any rename) breaks the test deterministically rather than silently passing against an old literal. See c8062f1.

mfw78:
Closing as the wrong shape.

The existing `ClientBehaviour` (`crates/swarm/node/src/protocol/behaviour.rs`) already manages pricing, retrieval, pushsync, and pseudosettle through a single unified handler that uses the pricing crate's wire-level types (`vertex_swarm_net_pricing::{inbound, outbound, AnnouncePaymentThreshold}` referenced from `protocol/upgrade.rs` and `protocol/handler.rs`).

The new `PricingBehaviour` and `PricingHandler` this PR added to `crates/swarm/net/pricing/` would have been redundant with what `ClientBehaviour` already does — we would have ended up with two parallel pricing implementations side-by-side.

The right work is a small `ClientBehaviour` change to make it node-type aware (Bootnode role activates only the pricing path; other roles activate retrieval / pushsync / pseudosettle as today) plus a `BootnodeBehaviour` rewire to compose `identify + topology + client`. Filing a focused follow-up issue.

Separately filed: ethersphere/bee#5496 to ask upstream to make pricing optional for connections involving bootnode-mode peers, removing the requirement to carry any pricing wiring on a bootnode at all if that lands.

PR #109, opened 2026-06-01
feat(net-handshake)!: bee-mainnet 15.0.0 handshake + typestate phases

mfw78:
Unit 4. Bumps handshake protocol to /swarm/handshake/15.0.0/handshake; signs full bee BzzAddress (underlay+overlay+nonce+timestamp+chequebook) per bee/pkg/bzz/address.go:138-160. Typestate Handshake<R, S> with consuming-self phase transitions. WelcomeMessage newtype (140-char cap). 27 tests pass including initiator/responder progression + recovery against a known bee keypair.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8151,6 +8151,7 @@ dependencies = [
  "vertex-swarm-identity",
  "vertex-swarm-net-proto",
  "vertex-swarm-peer",
+ "vertex-swarm-primitives",
  "vertex-swarm-spec",
  "vertex-swarm-test-utils",
 ]
@@ -8390,6 +8391,7 @@ version = "0.1.0"
 dependencies = [
  "alloy-primitives",
  "alloy-signer",
+ "alloy-signer-local",
  "arbitrary",
  "auto_impl",
  "bytes",
--- a/crates/swarm/net/handshake/Cargo.toml
+++ b/crates/swarm/net/handshake/Cargo.toml
@@ -64,4 +64,5 @@ arbitrary.workspace = true
 proptest.workspace = true
 proptest-arbitrary-interop.workspace = true
 rand.workspace = true
+vertex-swarm-primitives.workspace = true
 vertex-swarm-test-utils = { workspace = true }
--- a/crates/swarm/net/handshake/src/codec/ack.rs
+++ b/crates/swarm/net/handshake/src/codec/ack.rs
@@ -1,240 +1,372 @@
-//! Ack message encoding/decoding for handshake protocol.
+//! Ack message encoding/decoding for handshake 15.0.0.
+//!
+//! Wire layout (matches bee `pb/handshake.proto::Ack`):
+//!
+//! ```text
+//! Ack {
+//!   BzzAddress address     = 1;
+//!   uint64     network_id  = 2;
+//!   bool       full_node   = 3;
+//!   string     welcome_msg = 99;
+//! }
+//! BzzAddress {
+//!   bytes underlay           = 1;
+//!   bytes signature          = 2;
+//!   bytes overlay            = 3;
+//!   bytes nonce              = 4;
+//!   int64 timestamp          = 5;
+//!   bytes chequebook_address = 6;
+//! }
+//! ```
 
-use nectar_primitives::SwarmAddress;
+use alloy_primitives::Signature;
 use tracing::debug;
-use vertex_swarm_peer::{SwarmNodeType, SwarmPeer};
+use vertex_swarm_peer::{BzzAddress, Nonce, SwarmAddress, SwarmNodeType, Timestamp};
 
 use crate::HandshakeError;
-use crate::MAX_WELCOME_MESSAGE_CHARS;
+use crate::welcome::WelcomeMessage;
 
-/// Decode an Ack proto message, validating network_id and returning components.
+/// Validated contents of a decoded [`Ack`](vertex_swarm_net_proto::handshake::Ack).
+///
+/// Exposed publicly so the [`crate::session`] typestate API can consume it.
+/// `ethereum_address` is recovered from the signature; consumers building a
+/// legacy [`SwarmPeer`] should pass it to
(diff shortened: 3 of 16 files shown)


mfw78:
test

mfw78 (commented):
## Code review findings (high effort)

Inline `gh api` POSTs are not permitted in this session; posting findings as a single review comment with file:line references instead.

### 1. `crates/swarm/net/handshake/src/welcome.rs:31` (CONFIRMED — interop)
**Welcome message length uses Unicode char count; bee uses byte length.**
`WelcomeMessage::new` checks `s.chars().count() > MAX_WELCOME_MESSAGE_CHARS`, but bee's `handshake.go` (both `New` and `SetWelcomeMessage`) checks `len(welcomeMessage) > MaxWelcomeMessageLength` and `len()` on a Go string is bytes. A locally-configured 140-char emoji message (up to 4 bytes per char, ~560 bytes) passes our validation but bee peers reject the Ack with `ErrWelcomeMessageLength` and abort the handshake. Either match bee (use `s.len()`) or clamp the wire-bound length at 140 bytes at the codec layer.

### 2. `crates/swarm/peers/peer/src/bzz.rs` `parse_chequebook` (PLAUSIBLE — interop)
**Zero-byte vs 20-zero-byte chequebook diverges from bee's semantics.**
Bee's `ParseAddress` does `cb := common.BytesToAddress(chequebookAddress)` for both empty and 20-byte inputs; in-memory it's the same `common.Address{}`. Our parser distinguishes empty (`None`) from 20 zeros (`Some(Address::ZERO)`). A bee peer sending the 20-zero form to indicate "no chequebook" results in a Rust `Some(Address::ZERO)`. Any consumer that branches on `chequebook.is_some()` (e.g. chequebook gating) will misclassify those peers. Either treat `Address::ZERO` as `None`, or document the semantic equivalence and audit consumers.

### 3. `crates/swarm/net/handshake/src/protocol.rs` `now_timestamp()` capture timing (PLAUSIBLE)
**`now` is sampled once at handler entry and reused for both signing and remote skew validation.**
Bee's `bzz.CheckTimestamp` uses `MaxClockSkew = 60s`. Under load, the gap between sampling `now_timestamp()` and writing the Ack on the wire can be seconds. We sign with our captured `now`; by the time bee receives and evaluates, the gap eats into the 60s budget. Sample `now` immediately before each `BzzAddress::sign` call (and pass a fresh `now` to `decode_*` for skew validation) rather than once at the top of `do_inbound_exchange` / `do_outbound_exchange`.

### 4. `crates/swarm/peers/peer/src/bzz.rs` `BzzAddress::parse` rejects past timestamps (PLAUSIBLE)
**Stricter than bee's `ParseAddress`, which only rejects `<= 0` at parse time.**
Bee splits parse-time positivity from skew validation: `ParseAddress` only checks `timestamp <= 0`; `bzz.CheckTimestamp` only checks future bound (`newTimestamp > now + MaxClockSkew`). We reject both past and future > 6h. A peer reconnecting after a long partition (with a stored handshake record signed >6h ago) is parseable by bee but rejected here. If aligning with bee, only enforce the future-bound at parse.

### 5. `crates/swarm/net/handshake/src/session.rs` initiator `complete` defaulting (PLAUSIBLE — defense in depth)
```rust
let ethereum = self.state.remote_ethereum.unwrap_or_default();
... self.state.remote_node_type.unwrap_or(SwarmNodeType::Client),
    self.state.remote_welcome.clone().unwrap_or_default(),
```
In the defensive path that already checks `remote_bzz` for `Some`, these silently substitute `Address::ZERO` / `Client` / empty welcome rather than erroring. The typestate guarantees `Some` today, but if the invariant is ever violated the resulting `SwarmPeer` advertises `ethereum_address = 0x000...000`, which then flows into the peer manager and any chequebook verification. Bail with `MissingField(...)` symmetrically with `remote_bzz` instead.

### 6. `crates/swarm/net/handshake/src/welcome.rs` `WelcomeMessage::truncated` (PLAUSIBLE — UX)
`.chars().take(MAX_WELCOME_MESSAGE_CHARS)` slices on scalar boundaries, which can cut a flag emoji or ZWJ sequence mid-cluster. The local-side advertised message can render as a mojibake fragment to bee peers. Prefer `unicode-segmentation::Graphemes` to slice on grapheme boundaries.

### 7. `crates/swarm/net/handshake/src/session.rs` `SessionInputs.welcome_message` dead on the wire (PLAUSIBLE — API correctness)
`protocol.rs` populates `SessionInputs.welcome_message` for both initiator and responder but then ignores it and re-derives the welcome via a second `local_welcome(&self.identity)` call when encoding. The public field on `SessionInputs` is misleading: callers may reasonably assume that setting it changes the wire payload, but it does not. Either drop the field, or thread it through `encode_ack`/`encode_synack` so the session is the single source of truth.

### 8. `crates/swarm/net/handshake/src/codec/ack.rs` `map_parse_error` `drift_seconds` (PLAUSIBLE — observability)
The closure runs `now.map(|n| timestamp.as_secs().saturating_sub(n.as_secs())).unwrap_or(0)` only on the `TimestampOutsideSkewWindow` branch; for any other parse failure the original `drift_seconds` value is unused but the closure is still evaluated. Mainly cosmetic, but: when `now` is `None` and `BzzAddress::parse` returns skew error (it can't today, but the API allows it), the reported drift is `0` rather than something diagnostic. Either drop the `now: Option<...>` branch in `parse` (always require `now` when validating skew) or surface `drift_seconds = None` to operators rather than `0` to avoid misleading "drift = 0s" logs.

---

Effort: high (3 angles × 6 candidates → 1-vote verify, recall-biased). Most-severe first. Item 1 is the highest-impact for bee-mainnet interop; items 2–4 are most likely to bite once we connect to real peers.

mfw78:
Superseded by #102 which absorbs the handshake 15.0.0 wire + protocol-version bump. The proto schema changes (nested BzzAddress with timestamp + chequebook), codec rewrites, and version constant bump now ship together with the BzzAddress refactor in #102 (merged commit b6ea88c).

PR #110, opened 2026-06-01
feat(net-handshake): admission control trait + kademlia gate

mfw78:
Unit 5. HandshakeAdmissionControl trait so kademlia can veto a peer pre-Ack (mirrors bee Picker). AdmissionDecision { Accept, Reject(reason), AcceptEvict } (#[non_exhaustive]). KademliaAdmissionControl impl using new admission_within_capacity(overlay) — neighborhood bins always accept; eviction is Unit 8. 117 tests pass.

### DIFF
--- a/crates/swarm/net/handshake/src/admission.rs
+++ b/crates/swarm/net/handshake/src/admission.rs
@@ -1,0 +1,144 @@
+//! Handshake admission control.
+//!
+//! Lets routing (or any other gate) veto a handshake after the peer's
+//! identity is known, but before the final Ack is sent. Mirrors bee's
+//! `p2p.Picker` (`bee/pkg/p2p/libp2p/internal/handshake/handshake.go:167`).
+//!
+//! The handshake crate stays free of any routing/topology dependency: it
+//! defines the trait and accepts a dyn-erased implementation via
+//! [`HandshakeBehaviour::with_admission_control`](crate::HandshakeBehaviour::with_admission_control).
+
+use std::sync::Arc;
+
+use libp2p::PeerId;
+use vertex_swarm_peer::{SwarmAddress, SwarmNodeType};
+
+pub use vertex_net_peer_registry::ConnectionDirection;
+
+/// Reason a peer was rejected from completing the handshake.
+///
+/// `#[non_exhaustive]` so adding new reasons is non-breaking.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, strum::Display, strum::IntoStaticStr)]
+#[strum(serialize_all = "snake_case")]
+#[non_exhaustive]
+pub enum AdmissionRejection {
+    /// Routing capacity is full for this peer's proximity bin.
+    Saturated,
+    /// Peer is on the blocklist (banned).
+    Blocklisted,
+    /// Neighborhood is oversaturated and this peer falls outside the
+    /// configured headroom.
+    OversaturatedNeighborhood,
+}
+
+/// Why an existing peer is being evicted to make room for an incoming one.
+///
+/// `#[non_exhaustive]` so new eviction reasons can be added without breaking
+/// downstream consumers.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, strum::Display, strum::IntoStaticStr)]
+#[strum(serialize_all = "snake_case")]
+#[non_exhaustive]
+pub enum EvictReason {
+    /// Incoming peer is closer (higher proximity order) than the evictee.
+    Oversaturated,
+}
+
+/// Result of evaluating a peer for handshake admission.
+///
+/// `#[non_exhaustive]` so adding new outcomes (e.g. defer/retry) is
+/// non-breaking.
+#[derive(Debug, Clone)]
+#[non_exhaustive]
+pub enum AdmissionDecision {
+    /// Allow the handshake to complete.
+    Accept,
+    /// Reject the handshake. The handshake will abort before the final Ack
+    /// and the handler reports [`HandshakeError::AdmissionRejected`](crate::HandshakeError::AdmissionRejected).
+    Reject(AdmissionRejection),
+    /// Accept the new peer but evict an existing one first.
+    ///
+    /// The handshake crate does not perform eviction; it just surfaces the
+    /// decision. Unit 8 wires the eviction path through the topology
+    /// behaviour.
+    AcceptEvict {
+        evict: PeerId,
+        reason: EvictReason,
+    },
+}
+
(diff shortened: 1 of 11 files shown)


mfw78 (commented):
test

mfw78 (commented):
Code review (effort: high, recall-biased). Inline-comment posting is not permitted in this sandbox, so findings are reproduced here with explicit file:line references.

---

### 1. `crates/swarm/topology/src/kademlia/admission.rs:44-54` — Inbound peers are not counted at gate time; the gate is effectively a no-op at saturation

The comment on lines 44-49 asserts that the in-flight peer is already in `effective_count` via `reserve_inbound` "after handshake completes." But the admission gate runs **inside** `HandshakeProtocol::do_inbound_exchange` *before* the handshake completes. `reserve_inbound` is only invoked later in `protocol_handlers.rs::on_handshake_completed` (after the `HandshakeEvent::Completed` event is processed by topology).

Concrete failure: bin has `effective_count == ceiling` (target + headroom). An inbound peer's identity is verified, gate calls `admission_within_capacity` → `effective_count(po) <= ceiling` is `true` → Accept. Handshake completes successfully. Then `on_handshake_completed` runs `should_accept_inbound` (strict `<`) → rejects, silently closes the connection. The remote never sees `HandshakeError::AdmissionRejected` — exactly the structured signal this PR set out to provide. The gate is also informationally redundant with the later check for inbound near saturation.

Suggested fix: either (a) reserve an inbound slot earlier so the gate sees a consistent count, or (b) have `KademliaAdmissionControl::evaluate` add the +1 for `direction == Inbound` before consulting routing (i.e. `effective_count(po) + 1 <= ceiling`).

---

### 2. `crates/swarm/topology/src/kademlia/admission.rs:101-117` — Test `accepts_when_bin_at_ceiling` models a state production never reaches

The test pre-reserves the inbound peer (`RoutingCapacity::reserve_inbound(&*routing, &peer)`) *before* evaluating. As noted in finding (1), production does not reserve before evaluating an inbound peer at the gate — `reserve_inbound` is called only after `HandshakeEvent::Completed` is dispatched. The test gives false confidence that the gate behaves correctly at ceiling.

Suggested fix: either drop the pre-reserve in this test (to model the actual inbound flow and surface the off-by-one), or add a sibling test that explicitly reproduces the production order (evaluate-then-reserve) and assert the desired admission behavior at ceiling.

---

### 3. `crates/swarm/net/handshake/src/protocol.rs:202-211` (inbound) vs `:279-289` (outbound) — Asymmetric Ack semantics let outbound finish "successfully" while inbound rejects

`do_outbound_exchange` evaluates admission **before** sending the final Ack (good — abort is clean), but `do_inbound_exchange` evaluates admission **after** `decode_ack` has consumed the remote's Ack. By that point, the outbound peer has already completed `Framed::send::<Ack, _>` and `close()`; its `handle_outbound` returns `Ok(HandshakeInfo)` and topology marks the peer as connected. If the inbound side then `Reject`s, only the inbound peer fails with `AdmissionRejected`; the outbound side observes a successful handshake followed by a peer disconnect, recording it as a `HandshakeFailed`-style `DialFailed` only via the disconnect path. Two peers with independent local routing state (e.g. different depths, different `effective_count`) can therefore disagree on admission — the outbound side will dial the next time the peer reaches deficit, churning the connection.

Suggested fix: move the inbound `evaluate_admission` call to **before** `Framed::send::<SynAck>` (and have the gate work off `Syn`-time identity if available), or move the outbound check to *after* receiving the final response from the inbound side so both decisions occur post-symmetric-exchange. At minimum, document the asymmetry and confirm topology treats post-handshake disconnects-from-remote as `AdmissionRejected` for backoff purposes.

---

### 4. `crates/swarm/net/handshake/src/protocol.rs:119` — `AcceptEvict.evict` PeerId is silently discarded by the gate

```rust
AdmissionDecision::Accept | AdmissionDecision::AcceptEvict { .. } => Ok(()),
```

The `AcceptEvict { evict, reason }` arm collapses to `Ok(())` and drops the `evict` peer. While this is fine in this PR (no impl currently returns `AcceptEvict`), the trait already exposes `AcceptEvict` to downstream implementors, and the handshake crate doc says it "surfaces the decision" — but here the decision is consumed in a way that makes it unobservable beyond Accept. If a future `HandshakeAdmissionControl` impl (or even a downstream consumer of the crate) returns `AcceptEvict` thinking it triggers an eviction, the named peer will never be evicted and the bin will grow past ceiling.

Suggested fix (low cost, now): either forward the `AcceptEvict { evict, reason }` decision out as part of `HandshakeInfo` / a new `HandshakeHandlerEvent` variant, or (less invasive) explicitly `unreachable!()` / `debug_assert!` in this arm with a comment that wiring is gated on Unit 8 — so a stray `AcceptEvict` doesn't fail-silent.

---

*Effort: high. Process: 3 finder angles (line-by-line, removed-behavior, cross-file) → dedup → recall-biased verification against `protocol.rs`, `protocol_handlers.rs`, `routing.rs`, `limits.rs`. No tests were run.*

mfw78:
Superseded by #143.

#143 addresses every finding from the review:

* **F1 (off-by-one inbound)**: `KademliaRouting::admission_within_capacity` now takes `extra: usize`. `KademliaAdmissionControl::evaluate` passes `1` for inbound (peer not yet reserved at gate time, topology calls `reserve_inbound` only after processing `HandshakeEvent::Completed`) and `0` for outbound (slot already reserved at dial planning). The inbound gate no longer accepts at ceiling and silently disconnects post-completion.
* **F2 (test models wrong state)**: rewritten as separate `inbound_accepts_at_ceiling_minus_one` / `inbound_rejects_when_bin_already_full` / `outbound_accepts_at_ceiling_after_dial_reserve` / `outbound_rejects_when_bin_oversaturated` cases that model the actual production orderings.
* **F3 (Ack-asymmetry)**: documented at the module level in `admission.rs` (outbound learns from SYNACK and aborts before sending ACK; inbound learns from ACK and necessarily gates after the outbound side has finished). The asymmetry is inherent to the 3-message exchange and would require a fourth message to remove. The rejecting side surfaces `AdmissionRejected`; the remote sees a transport-level disconnect, which topology already handles.
* **F4 (`AcceptEvict` silently dropped)**: `AcceptEvict` removed from #143. `#[non_exhaustive]` on `AdmissionDecision` means a future PR can add it once the eviction wiring is end-to-end, which is the right place to introduce a variant whose action is non-trivial.

Plus the project rules the original PR violated: no "Unit N" rustdoc refs, no em-dashes, no inline bee line references in production code, fmt clean, clippy clean.

PR #111, opened 2026-06-01
feat(net-hive)!: hive 2.0.0 wire + verifier trait

mfw78:
Unit 6. Bumps protocol to /swarm/hive/2.0.0/peers; Peer proto extended to bee's BzzAddress (adds timestamp int64 + chequebook_address bytes). HiveVerifier trait with HiveRejection { InvalidSignature, TimestampOutsideSkewWindow, NetworkIdMismatch, Blocklisted, PrivateMultiaddr } (#[non_exhaustive]).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8183,6 +8183,8 @@ name = "vertex-swarm-net-hive"
 version = "0.1.0"
 dependencies = [
  "alloy-primitives",
+ "alloy-signer",
+ "bytes",
  "futures",
  "hashlink",
  "libp2p",
@@ -8198,6 +8200,7 @@ dependencies = [
  "vertex-net-utils",
  "vertex-observability",
  "vertex-swarm-api",
+ "vertex-swarm-identity",
  "vertex-swarm-net-handler-core",
  "vertex-swarm-net-headers",
  "vertex-swarm-net-proto",
--- a/crates/swarm/net/hive/Cargo.toml
+++ b/crates/swarm/net/hive/Cargo.toml
@@ -35,6 +35,9 @@ libp2p.workspace = true
 ## crypto
 alloy-primitives.workspace = true
 
+## bytes
+bytes.workspace = true
+
 ## cache
 hashlink.workspace = true
 parking_lot.workspace = true
@@ -49,3 +52,5 @@ strum.workspace = true
 thiserror.workspace = true
 
 [dev-dependencies]
+vertex-swarm-identity.workspace = true
+alloy-signer.workspace = true
--- a/crates/swarm/net/hive/src/bzz.rs
+++ b/crates/swarm/net/hive/src/bzz.rs
@@ -1,0 +1,286 @@
+//! Local minimal `BzzAddress` for hive 2.0.0.
+//!
+//! This type mirrors bee's `pkg/bzz.Address` on the wire and is sufficient for
+//! hive 2.0.0 peer-record verification. It will be replaced by the canonical
+//! `BzzAddress` introduced by Unit 2 of the indexed-snacking-crown plan once
+//! that lands; the field layout is intentionally identical.
+//!
+//! ### Sign data layout (bee 15.0.0+)
+//!
+//! ```text
+//! "bee-handshake-"        // 14 bytes
+//! || underlay_bytes       // wire-encoded multiaddrs (length-prefixed entries)
+//! || overlay              // 32 bytes
+//! || network_id           // 8 bytes, big-endian
+//! || nonce                // 32 bytes
+//! || timestamp            // 8 bytes, big-endian (unix seconds, as u64 cast)
+//! || chequebook_address   // 0 or 20 bytes (zeros == "no chequebook")
+//! ```
+//!
+//! Reference: `bee/pkg/bzz/address.go` `generateSignData` (lines 138-160).
+//! The EIP-191 personal-sign prefix is applied by
+//! [`alloy_primitives::Signature::recover_address_from_msg`].
+//!
+//! ### Timestamp skew window
+//!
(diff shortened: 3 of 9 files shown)


mfw78:
(scratch comment from review tooling — superseded; please ignore)

mfw78 (commented):
See inline comments

mfw78 (commented):
## Automated code review — hive 2.0.0 + verifier trait

Recall-biased pass at high effort. Inline-comment APIs were not available in this session, so findings are consolidated here with file:line references. Ranked roughly by impact.

---

### 1. `crates/swarm/net/hive/src/codec.rs:27` — **Vertex-to-vertex hive gossip is dead-on-arrival**

Every record we emit hard-codes `timestamp: 0`, but the new inbound `BzzAddress::from_wire` rejects any record with `timestamp <= 0` (returns `TimestampInvalid`, which then maps via `From<BzzAddressError>` to `HiveRejection::TimestampOutsideSkewWindow`). The PR comment notes bee skips these — that also means a peer *vertex* receiving our broadcast rejects every entry. Until Unit 4 lands, hive gossip between two vertex nodes is silently non-functional. Consider either (a) not sending the broadcast at all in this interim state (and bumping a metric), or (b) gating the emit behind a `cfg`/feature so it's explicit that no usable peers leave this node.

### 2. `crates/swarm/net/hive/src/verifier.rs:74-80` — **Label collapse regresses observability**

The pre-PR validator emitted distinct metric labels for `overlay_length`, `nonce_length`, `chequebook_length`, `no_underlays`, and `underlay_decode`. The `From<BzzAddressError>` impl folds all of these into the single `Malformed` bucket, so `hive_peer_validation_failures_total{reason="malformed"}` is now load-bearing for five distinct failure modes. Either keep the fine-grained variants as siblings under `HiveRejection`, or expose a sub-label so operators can still distinguish underlay-decode from chequebook-length failures when triaging gossip storms.

### 3. `crates/swarm/net/hive/src/verifier.rs:77` — **Mislabel: structural timestamp invariant reported as skew failure**

`BzzAddressError::TimestampInvalid` is emitted by `BzzAddress::from_wire` when `timestamp <= 0` — a structural/malformed condition, not a clock-skew condition. Mapping it to `HiveRejection::TimestampOutsideSkewWindow` makes the skew counter spike on every malformed-timestamp record (including every record we currently emit ourselves; see finding 1). Suggest mapping `TimestampInvalid` to `Malformed`, and reserving `TimestampOutsideSkewWindow` for the actual `check_timestamp_skew` failure inside `verify`.

### 4. `crates/swarm/net/hive/src/verifier.rs:184` — **Self-overlay defence reordered behind timestamp/skew, weakening the diagnostic**

Pre-PR, self-overlay was the *first* check in `validate_proto_peer`, so a misbehaving peer (or relay loop) that echoed our overlay back to us produced a clean `self_overlay` metric. Now, because `BzzAddress::from_wire` runs first and our own outbound broadcasts carry `timestamp: 0`, a self-echo lands as `Malformed`/`TimestampOutsideSkewWindow` instead. The defence still works (we reject), but the operator signal is gone. Consider moving the self-overlay check ahead of `from_wire` (operating on the raw overlay bytes), or at least ahead of `check_timestamp_skew` inside `verify`.

### 5. `crates/swarm/net/hive/src/bzz.rs:270` — **Public `generate_sign_data` casts `i64 -> u64` with no range check**

Inbound callers go through `from_wire` which rejects `timestamp <= 0`, so this is safe today — but `generate_sign_data` is `pub`, so a future caller (e.g. an outbound signer wired up in Unit 4) that hands in a negative `i64` will silently produce big-endian bytes for a huge `u64` and emit an unverifiable signature. Either accept `u64` here, or `assert!(timestamp >= 0)` / return a `Result`.

### 6. `crates/swarm/net/hive/src/verifier.rs:201-205` — **Blocklist short-circuits ECDSA on an unverified claimed overlay**

The blocklist lookup keys on `addr.overlay`, but at this point we have not verified that the sender actually owns that overlay. An attacker can craft a record bearing any blocklisted overlay (with arbitrary multiaddrs/signature/nonce/chequebook); we will report `Blocklisted` against a peer who never sent the record. This is more a metric-poisoning vector than an exploit, but it also hides that the cheapness comes at the cost of trusting the claimed overlay. Consider running signature recovery + overlay derivation first, then doing the blocklist lookup against the *derived* overlay.

### 7. `crates/swarm/net/hive/src/verifier.rs` — **Dead `HiveRejection` variants: `NetworkIdMismatch` and `PrivateMultiaddr` declared but never emitted**

`DefaultHiveVerifier::verify` never returns either of these. Docs imply `hive_peer_validation_failures_total{reason="network_id_mismatch"}` is a useful operator signal, but it will never increment. Either implement the corresponding checks, or drop the variants until the policy that emits them lands, so dashboards/alerts don't go stale.

### 8. `crates/swarm/net/hive/src/protocol.rs:89` — **Minor: per-stream `Arc<DefaultHiveVerifier>` allocation on the hot path**

A fresh `Arc<DefaultHiveVerifier>` is allocated per inbound stream. Not a correctness bug, but verifier construction is now on the hot path and `HiveInner` already owns `identity` (which carries network_id + overlay). Consider lifting verifier construction into `HiveInner::new` (or the behaviour) and reusing the `Arc` across streams.

---

*Generated by code-review skill (high effort, recall-biased). Comments are advisory.*

mfw78:
Superseded by #102 which absorbs the hive 2.0.0 wire + per-record verification. The proto schema changes (nested BzzAddress), codec rewrites, and version constant bump now ship together with the BzzAddress refactor in #102 (commit b6ea88c). The verifier trait refactor is deferred and can land as a follow-up.

PR #112, opened 2026-06-01
refactor(topology): typed SaturationDecision + bootnode evict policy

mfw78:
Unit 8. Replaces bool returns with #[non_exhaustive] SaturationDecision { Accept, AcceptEvict(EvictionCandidate), Reject(RejectionReason) }. BinPolicy trait — StandardPolicy rejects on oversat, BootnodePolicy evicts lowest-reachability peer (mirrors bee kademlia.go:1193-1199). Deterministic eviction order: (reachability, score, last_seen).

### DIFF
--- a/crates/swarm/topology/src/kademlia/config.rs
+++ b/crates/swarm/topology/src/kademlia/config.rs
@@ -44,6 +44,14 @@ impl KademliaConfig {
         self.limits = self.limits.with_inbound_headroom(headroom);
         self
     }
+
+    /// Borrow the depth-aware bin limits. Used by callers (e.g. the
+    /// saturation policy bridge and integration tests) that need to
+    /// build a [`crate::saturation::SaturationContext`] without
+    /// constructing limits separately.
+    pub fn depth_aware_limits(&self) -> &DepthAwareLimits {
+        &self.limits
+    }
 }
 
 #[cfg(test)]
--- a/crates/swarm/topology/src/kademlia/limits.rs
+++ b/crates/swarm/topology/src/kademlia/limits.rs
@@ -16,7 +16,7 @@ pub(crate) const DEFAULT_INBOUND_HEADROOM: usize = 4;
 ///
 /// Stateless: callers provide depth explicitly to avoid dual-source-of-truth bugs.
 #[derive(Debug, Clone)]
-pub(crate) struct DepthAwareLimits {
+pub struct DepthAwareLimits {
     total_target: usize,
     /// Minimum peers per bin.
     nominal: usize,
@@ -31,7 +31,7 @@ impl Default for DepthAwareLimits {
 
 impl DepthAwareLimits {
     /// Create with total target and nominal minimum per bin.
-    pub(crate) fn new(total_target: usize, nominal: usize) -> Self {
+    pub fn new(total_target: usize, nominal: usize) -> Self {
         Self {
             total_target,
             nominal,
--- a/crates/swarm/topology/src/kademlia/mod.rs
+++ b/crates/swarm/topology/src/kademlia/mod.rs
@@ -7,6 +7,7 @@ mod config;
 mod limits;
 pub(crate) mod peer_selection;
 mod routing;
+pub mod saturation;
 mod task;
 
 pub use args::RoutingArgs;
--- a/crates/swarm/topology/src/kademlia/saturation.rs
+++ b/crates/swarm/topology/src/kademlia/saturation.rs
@@ -1,0 +1,478 @@
+//! Typed saturation decisions and bin-admission policies.
+//!
+//! Replaces bool-returning saturation checks with explicit
+//! [`SaturationDecision`] variants so the routing layer (and future
+//! handshake admission control in Unit 5) can express the difference
+//! between an outright rejection, a clean accept, and a slot-reuse
+//! ("accept-evict") path.
+//!
+//! Two [`BinPolicy`] implementations are provided:
+//!
+//! - [`StandardPolicy`] — the default; mirrors today's `Reject` behaviour
+//!   when a bin is over-saturated.
+//! - [`BootnodePolicy`] — bootnode mode; on over-saturation it evicts the
+//!   lowest-quality peer in the target bin and admits the newcomer. This
+//!   mirrors bee's
+//!   [`Kad.Connected`](https://github.com/ethersphere/bee/blob/master/pkg/topology/kademlia/kademlia.go#L1193-L1199)
+//!   bootnode branch.
+//!
+//! Eviction selection orders peers ascending by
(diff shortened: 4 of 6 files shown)


mfw78:
(comment redacted by reviewer)

mfw78:
(redacted)

mfw78:
(redacted)

mfw78:
(redacted)

mfw78:
Finding 1: BootnodePolicy does not cap bin size under burst. AcceptEvict removes only one peer per decide call but is_oversaturated is keyed off effective_count (dialing+handshaking+active). When effective_count exceeds ceiling by more than one, the policy admits-and-evicts once per inbound, leaving the bin still over ceiling. File: crates/swarm/topology/src/kademlia/saturation.rs lines 305-323. Suggest iterating eviction until effective_count - evictions is at or below ceiling, or surface the over-by-N to the caller.

mfw78:
Finding 2: BootnodePolicy is a churn engine for newly admitted peers. PeerReachability::Unknown has evict_rank 0 (worst) and is the dominant sort key. A peer freshly admitted by a bootnode starts as Unknown (AutoNAT has not classified it yet). The next over-saturation event in the same bin picks that fresh peer as the eviction victim if any other peer has progressed to Public. Under steady-state inbound load this becomes admit-then-immediately-evict, wasting handshake CPU and preventing new peers from ever reaching Public. File: crates/swarm/topology/src/kademlia/saturation.rs lines 42-58 and 305-323. Suggest treating Unknown neutrally vs Public for a recently-admitted window, or gating eviction by a minimum age on last_seen.

mfw78:
Finding 3: Empty bin_peers silently downgrades BootnodePolicy to StandardPolicy. When bin_peers is empty but is_oversaturated() is true, BootnodePolicy falls back to Reject(BinFull), the same behaviour as StandardPolicy, with no warning, log, or error. The doc comment acknowledges this is legitimate when over-saturation comes from in-progress slots, but the same path is reached if the routing layer simply forgets to enumerate the bin (or passes an empty slice). Bootnode admission becomes standard admission silently. File: crates/swarm/topology/src/kademlia/saturation.rs lines 310-321. Suggest a debug_assert or tracing::warn so this misuse surfaces.

mfw78:
Finding 4: New SaturationDecision API is not wired in; PR description overstates replaces. PR description says Replaces bool returns with SaturationDecision. The actual admission path in KademliaRouting::should_accept_inbound still returns bool, and protocol_handlers.rs still consumes that bool. The new module is consumed only by tests. Two parallel admission surfaces now exist: the typed one nothing uses, and the bool one everything uses. Risk: drift between them, and readers infer a behaviour change when only scaffolding shipped. Files: crates/swarm/topology/src/kademlia/routing.rs lines 489-502 and crates/swarm/topology/src/protocol_handlers.rs line 113. Suggest clarifying the wording (introduces typed SaturationDecision for upcoming Unit-5 integration) or wiring should_accept_inbound through StandardPolicy.

mfw78:
Finding 5: PeerStats derives PartialEq while holding f64. PeerStats with score: f64 derives PartialEq. NaN score makes equality non-reflexive (p == p false). Today only tests and compare_for_eviction (which uses partial_cmp with unwrap_or(Equal), safe) consume it, but the public PartialEq impl is a trap for future callers. File: crates/swarm/topology/src/kademlia/saturation.rs lines 162-173. Drop the derive, exclude score via a manual impl, or document the no-NaN invariant.

mfw78:
Finding 6: compare_for_eviction allocates per-comparison via PeerId::to_bytes(). The expression a.peer_id.to_bytes().cmp(b.peer_id.to_bytes()) allocates two Vec<u8> per comparison; min_by performs N-1 comparisons per worst_peer call. Cheap today (eviction is rare), but it makes worst_peer non-trivial in tight loops, and PeerId exposes Multihash digest bytes that compare without allocation. File: crates/swarm/topology/src/kademlia/saturation.rs lines 267-271. Suggest caching (&[u8], &PeerStats) pairs before min_by, or comparing via canonical multihash bytes directly.

mfw78:
Finding 7: PeerReachability cannot express confirmed private or NAT-bound peers. Only Unknown and Public exist. Once Unit 10 wires real reachability, a peer confirmed unreachable (Private in AutoNAT terms) will still map to Unknown here, ranked identically to a peer never probed. That collapses two semantically different states into one eviction bucket and makes the comment Unknown < Public so the policy keeps confirmed-reachable peers misleading: there is no representation of confirmed-unreachable today. File: crates/swarm/topology/src/kademlia/saturation.rs lines 42-58. Add a Private variant now with evict_rank below Unknown so the bucket survives Unit 10s wiring, or document the planned addition.

mfw78:
Finding 8: EvictReason::LowestReachability is misleading when score or last_seen actually decided. EvictReason has a single variant, LowestReachability, and worst_peer always stamps that reason. But the ordering chain falls through to score, last_seen, and peer_id whenever reachability rank is tied (which it always is today since UnknownReachability is the default and only one non-default variant exists). The reason reported in tests and logs reads LowestReachability even when the deciding axis was the score tie-break, which is misleading for diagnostics. File: crates/swarm/topology/src/kademlia/saturation.rs lines 122-126 and 245-250. Add LowestScore / Oldest / Tiebreak variants computed from the actual deciding axis, or rename the variant to LowestRank.

mfw78:
Superseded by #144.

#144 keeps the same architectural shape (typed `SaturationDecision` + `BinPolicy` trait + `StandardPolicy`/`BootnodePolicy`) and addresses the substantive issues from the original:

* **Parallel `PeerReachability` enum**: the PR defined its own 2-variant copy (`Unknown`/`Public`) inside `saturation.rs` with a comment that "Unit 10 will replace the in-module definition". The full 3-variant `PeerReachability` from `vertex_swarm_topology::reachability` has been on main since #106 merged. #144 uses the existing enum, so Private peers are correctly evicted before Unknown ones (the PR's 2-variant version would have grouped them under `evict_rank=0` and broken the dominant-reachability contract its own tests claim to enforce).
* **`DepthAwareLimits` leaked through public surface**: the PR made `DepthAwareLimits` (and `DepthAwareLimits::new`) `pub` so the integration test in `tests/oversaturation.rs` could construct contexts via `KademliaConfig::depth_aware_limits()`. #144 takes `oversaturated: bool` directly in `SaturationContext`: the routing layer computes the verdict once against its own limits and hands it in. `DepthAwareLimits` stays `pub(crate)` and the integration test is unnecessary because the policy is testable in unit form.

Plus the project rules: no "Unit 5" / "Unit 8" / "Unit 10" rustdoc refs, no em-dashes, no inline bee line references (`kademlia.go:1193-1199` had two inline mentions).

PR #113, opened 2026-06-01
feat(net-pingpong): typed messages + stabilization detector

mfw78:
Unit 9. Typed PingpongMessage { Ping(Greeting), Pong(GreetingEcho) } with Greeting max-len. RTT observed and fed to peer-score (EMA). StabilizationDetector trait (N consecutive ok within T window); stable status surfaced via TopologyHandle for Unit 10's reachability promotion.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8447,6 +8447,7 @@ dependencies = [
 name = "vertex-swarm-peer-score"
 version = "0.1.0"
 dependencies = [
+ "libp2p",
  "paste",
  "postcard",
  "serde",
--- a/crates/net/peer/score/src/lib.rs
+++ b/crates/net/peer/score/src/lib.rs
@@ -9,12 +9,31 @@ use serde::{Deserialize, Serialize};
 const MIN_SCORE: f64 = -100.0;
 const MAX_SCORE: f64 = 100.0;
 
+/// Smoothing factor for the exponential moving average of RTT.
+///
+/// `alpha = 0.2` weights the most recent sample at 20% and historical at 80%,
+/// matching bee's pingpong RTT smoothing. Higher alpha responds faster to
+/// changes; lower alpha is more stable. Chosen empirically for liveness probes
+/// that arrive every few seconds.
+const RTT_EMA_ALPHA: f64 = 0.2;
+
 /// Lock-free peer scoring using atomics for concurrent access.
+///
+/// In addition to the score and rolling average latency, an exponentially
+/// weighted RTT estimate is maintained via [`PeerScore::update_rtt`]. The EMA
+/// is exposed separately from the rolling average to keep the historical mean
+/// for legacy serialisation while giving stabilization/connection-quality
+/// consumers a smoother, more responsive estimate.
 #[derive(Debug, Serialize, Deserialize)]
 pub struct PeerScore {
     score: AtomicF64,
     latency_sum_nanos: AtomicU64,
     latency_samples: AtomicU32,
+    /// Exponential moving average of recent ping RTT, in nanoseconds.
+    ///
+    /// `0` is the sentinel for "no samples yet"; once at least one
+    /// `update_rtt` call lands the value is strictly positive.
+    ema_rtt_nanos: AtomicF64,
 }
 
 impl Clone for PeerScore {
@@ -24,6 +43,7 @@ impl Clone for PeerScore {
             score: AtomicF64::new(self.score.load(Ordering::Relaxed)),
             latency_sum_nanos: AtomicU64::new(self.latency_sum_nanos.load(Ordering::Relaxed)),
             latency_samples: AtomicU32::new(self.latency_samples.load(Ordering::Relaxed)),
+            ema_rtt_nanos: AtomicF64::new(self.ema_rtt_nanos.load(Ordering::Relaxed)),
         }
     }
 }
@@ -40,6 +60,7 @@ impl PeerScore {
             score: AtomicF64::new(0.0),
             latency_sum_nanos: AtomicU64::new(0),
             latency_samples: AtomicU32::new(0),
+            ema_rtt_nanos: AtomicF64::new(0.0),
         }
     }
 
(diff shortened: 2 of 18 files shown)


mfw78:
test

mfw78 (commented):
test review comment

mfw78:
(skill harness test artifact — please ignore)

mfw78 (commented):
## Code review findings - Unit 9 (typed pingpong + stabilization)

Reviewed at high effort with three independent finder angles (line-by-line, removed-behavior, cross-file). Posting as a single PR review because the harness denied `gh api POST repos/.../pulls/.../comments` (inline review-comment API) and no inline-comment MCP tool is available in this session. File:line refs are inlined per finding.

Severity: `(H)` high, `(M)` medium, `(L)` low.

---

### 1. (H) PeerScore postcard layout change breaks on-disk backward compatibility

**File:** `crates/net/peer/score/src/lib.rs` (struct definition); test rename in same file (`test_backward_compat` -> `test_wire_format_layout`).

Adding `ema_rtt_nanos: AtomicF64` as a new field on `PeerScore` changes the postcard layout from `(f64, u64, u32)` to `(f64, u64, u32, f64)`. `PeerScore` is persisted via `ScoreTable` in `crates/swarm/peers/peer-manager/src/db_store.rs:25` using postcard, a sequential, non-self-describing format. After upgrade, `get_score` will hit `DeserializeUnexpectedEnd` for every pre-existing row - the bytes on disk are 4 bytes short of the new struct.

The old `test_backward_compat` was renamed and now only verifies the new 4-field layout round-trips. There is no test for legacy 3-field decoding (because it cannot), and no migration logic anywhere in `peer-manager`.

**Fix:** restore legacy-shape decoding (custom `Deserialize` that tolerates the missing trailing field, defaulting `ema_rtt_nanos` to 0), version the on-disk record, or schedule a migration / table drop.

**Failure scenario:** upgrade a running node -> restart -> `DbPeerStore::get_score` returns `Err` for every persisted peer -> score subsystem effectively forgets all peer history.

---

### 2. (M) ConsecutiveOkDetector HashMap grows unbounded - forget is never called

**File:** `crates/swarm/peers/peer-score/src/stabilization.rs` (`forget` doc says "Call on disconnect to avoid unbounded growth.") and `crates/swarm/topology/src/protocol_handlers.rs` (newly added handlers never invoke it).

`ConsecutiveOkDetector::forget(peer)` exists with an explicit docstring, but grep across `crates/swarm/topology/src/` finds no caller against the stabilization detector. Disconnect / ban / removal paths never call it. With peer churn, `peers: Mutex<HashMap<PeerId, PeerState>>` accumulates entries for every peer ever observed. A failed observation does `peers.remove(&peer)`, but only on explicit `ok = false`; a peer that disconnects silently after a successful streak stays forever.

**Fix:** call `self.stabilization.lock().forget(&peer_id)` (or expose a helper on the shared handle) from the connection-closed / peer-removed handler.

---

### 3. (M) Stabilization "window" check resets the streak instead of sliding

**File:** `crates/swarm/peers/peer-score/src/stabilization.rs` (`observe`).

Detector advertises "N consecutive ok within a sliding T window", but the second guard `if now.duration_since(entry.streak_started_at) > self.config.window { reset; return; }` resets the streak (not slides it) whenever total elapsed exceeds the window. With `DEFAULT_REQUIRED_OK = 3` and `DEFAULT_WINDOW = 30s`, a 16s ping cadence never stabilizes: at t=32s, 32 > 30 -> reset to consecutive_ok=1, repeat forever.

**Fix:** if intent is "3 consecutive observations where each pair is within the window", drop the `streak_started_at` check and keep only `now.duration_since(entry.last_ok_at) > window`. If the stricter "3 within window from first" is intentional, document it and size the window comfortably above N x ping cadence.

---

### 4. (M) unreachable_pingpong_event() returns a random PeerId - misroutes future PingpongEvent variants

**File:** `crates/swarm/topology/src/composed.rs` (`unreachable_pingpong_event`) and `crates/swarm/topology/src/protocol_handlers.rs` (wildcard `ProtocolEvent::Pingpong(other)` arm).

`ProtocolEvent::peer_connection` now routes any unrecognised `PingpongEvent` (the enum is `#[non_exhaustive]`) through `unreachable_pingpong_event()`, which returns `(PeerId::random(), ConnectionId::new_unchecked(0))`. That tuple is handed to `process_protocol_event(peer_id, ...)` and ends up in `trace!(%peer_id, event = ?other, "Unhandled pingpong event")`. The trace tags every unknown variant with a freshly-fabricated peer id, masking which peer actually sent it. Also a footgun for any future caller that promotes that wildcard from `trace!` to "record metric / score against peer".

**Fix:** make `peer_connection` return `Option<(PeerId, ConnectionId)>`, or move the routing decision into `process_protocol_event` where the variant is still in scope and the real peer_id can be extracted (every current `PingpongEvent` variant carries `peer_id`).

---

### 5. (L) Inbound Greeting::try_from(ping.greeting) may reject legitimate Bee peers

**File:** `crates/swarm/net/pingpong/src/protocol.rs:46-50`.

Bee accepts arbitrary-length greetings. Vertex now hard-rejects anything > MAX_GREETING_CHARS (64) with `ProtocolStreamError::Upgrade`, which closes the stream. 64 chars is plenty for "ping", but any Bee variant that probes with a longer payload would look like a misbehaving peer and get scored down. Since `MAX_MESSAGE_SIZE = 4096` already bounds memory, consider widening the per-greeting cap, or downgrading the rejection to a debug-truncate-and-continue.

---

### 6. (L) update_rtt EMA can collapse to Some(0) for sub-nanosecond samples

**File:** `crates/net/peer/score/src/lib.rs` (`update_rtt` / `ema_rtt_nanos`).

Once seeded, an EMA result `0 < v < 1.0` casts to `Some(0u64)` via `v as u64`, contradicting the documented sentinel ("0 is the sentinel for no samples yet"). Theoretical for real RTTs but the consumer cannot distinguish "never observed" from "decayed to ~0" from the public API. Round up `v as u64` to at least 1 when `v > 0.0`, or wrap in an explicit `has_sample` flag.

---

### 7. (L) Pong and RttObserved both fire per exchange, doubling per-event dispatch

**File:** `crates/swarm/net/pingpong/src/behaviour.rs:128-145`.

The behaviour push_backs two `ToSwarm::GenerateEvent`s for each successful exchange (`RttObserved` then `Pong`). Both are routed through `ProtocolEvent::peer_connection` and `process_protocol_event`. `on_pingpong_pong` updates `record_latency` (rolling avg) and `on_pingpong_rtt` updates `update_rtt` (EMA) - two different sinks, intentional, but every RTT-only consumer pays double dispatch. Not a correctness bug. Pick one: (a) consume `Pong` for both updates and drop `RttObserved`; (b) keep `RttObserved` and remove `record_latency` from the `Pong` handler.

---

### Out-of-scope nit

`SharedStabilizationDetector = Arc<StdMutex<Box<dyn StabilizationDetector>>>` double-locks (outer `StdMutex` + inner `Mutex<HashMap>` in `ConsecutiveOkDetector`) - collapse one. Performance only.

mfw78:
Closing based on investigation findings (full reports in conversation context, summarised here).

**Bee's stabilization is not what vertex thought it was.**

Agent A read bee mainnet at `21de6da9db529745017cb020989f5c42e672074d`: bee's `stabilization.Detector` is a network-LEVEL warmup gate, not a per-peer mechanism. It watches the global rate at which new peers are added to the kademlia table; once the standard deviation across N time-buckets falls below a threshold it flips one-shot from `Monitoring` to `Stabilized` and stays there. Six consumers in bee use it (pushsync, pusher, salud, reserve, storage-incentives, API warmup-flag), all asking the same question: 'is the node finished bootstrapping?' The detector has zero peer identity in its API or state, and never adds/evicts/promotes/demotes any peer.

This PR's `ConsecutiveOkDetector` is a per-peer pingpong success counter with a sliding window. The two share a name and nothing else. They differ on signal, scope, algorithm, lifecycle, and consumer surface.

**Upstream survey (rust-libp2p `3e72d4c071d5`, Lighthouse `8e4df4ababf6`):**

- `libp2p-ping::Event { peer, connection, result: Result<Duration, Failure> }` is exactly the per-peer health signal vertex wants. Its docstring is explicit: 'It is up to the user to implement a health-check / connection management policy based on the ping protocol.' Adopting libp2p-ping would let the existing `ReachabilityTracker` subscribe directly; the entire `vertex-swarm-net-pingpong` crate and the stabilization detector reduce to one event handler.
- `libp2p-autonat` is already integrated into `ReachabilityTracker` and already does multi-probe confidence building (`confidence_max: 3`). For 'promote to Public' the probe round-trip is strictly more authoritative than a pingpong reply.
- Lighthouse's `Score { Max, Real(RealScore { lighthouse_score, gossipsub_score, score, last_updated }) }` with half-life decay matches the shape vertex's existing `PeerScore` already follows (latency rolling avg + EMA RTT + score field). Lighthouse has no separate 'stabilization' concept; healthiness is `score >= threshold`.

**Filed instead:** two focused follow-ups for the actual underlying needs:

- Network-level warmup gate (the real bee mechanism), tracked separately.
- Reachability tracker simplification (consume libp2p-ping events or rely on AutoNAT alone), tracked separately.

The PR's typed `PingpongMessage` and RTT measurement might be salvageable into a smaller PR, but only if vertex chooses to keep `vertex-swarm-net-pingpong` instead of migrating to `libp2p-ping` (a separate decision in the follow-ups).

PR #115, opened 2026-06-01
test(integration): multi-node vertex cluster harness

mfw78:
Unit 14. New cluster harness + bootnode_cluster integration test (1 vertex bootnode + 2 vertex clients). Asserts handshake completion, hive gossip convergence, kademlia depth, and discard-metric assertions where wired.

Known issue (follow-up): harness shutdown path via initiate_graceful_shutdown only fires when an outer TaskManager is actively polled; in #[tokio::test] reuse outside the bundled test the GracefulShutdown event queues but the Signal never fires, so 5s join timeouts mask as 'did not terminate' failures. Bundled test passes.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8452,6 +8452,7 @@ dependencies = [
  "thiserror 2.0.18",
  "tokio",
  "tracing",
+ "tracing-subscriber",
  "vertex-net-local",
  "vertex-net-peer-store",
  "vertex-node-api",
@@ -8633,15 +8634,20 @@ dependencies = [
  "alloy-primitives",
  "alloy-signer",
  "alloy-signer-local",
+ "eyre",
  "libp2p",
  "nectar-primitives",
  "parking_lot",
  "rand 0.9.4",
+ "tokio",
  "vertex-swarm-api",
  "vertex-swarm-identity",
+ "vertex-swarm-node",
  "vertex-swarm-peer",
  "vertex-swarm-primitives",
  "vertex-swarm-spec",
+ "vertex-swarm-topology",
+ "vertex-tasks",
 ]
 
 [[package]]
--- a/crates/swarm/identity/src/lib.rs
+++ b/crates/swarm/identity/src/lib.rs
@@ -25,6 +25,13 @@ pub use vertex_swarm_api::SwarmIdentity as IdentityTrait;
 ///
 /// Holds an `Arc<Spec>` for shared access to the network specification.
 /// Wrap in `Arc<Identity>` for sharing across components.
+///
+/// `Clone` is cheap: all heap state lives behind `Arc`, so cloning bumps
+/// reference counts only. This enables the integration harness (see
+/// `vertex-swarm-test-utils::cluster`) to construct a single persistent
+/// `Identity` and hand owned copies to each node builder, mirroring how a
+/// real bootnode reuses a stable keystore across restarts.
+#[derive(Clone)]
 pub struct Identity {
     spec: Arc<Spec>,
     signer: Arc<LocalSigner<SigningKey>>,
--- a/crates/swarm/net/handshake/src/protocol.rs
+++ b/crates/swarm/net/handshake/src/protocol.rs
@@ -153,9 +153,20 @@ impl<I: SwarmIdentity> HandshakeProtocol<I> {
         let local_peer =
             prepare_local_peer(&self.identity, &self.additional_addrs, &observed_multiaddr)?;
 
-        // Send SYNACK: echo their observed addr back + our identity.
+        // Echo what we observed about the dialer (their `remote_addr` as
+        // libp2p reported it) back in the SYNACK; the dialer validates the
+        // peer id of this address against its own. An earlier version sent
+        // the responder's own observed address instead, which the dialer
+        // always rejected because the peer ids could never match.
+        let mut dialer_observed = self.remote_addr.clone();
+        if extract_peer_id(&dialer_observed).is_some() {
+            dialer_observed.pop();
+        }
+        let dialer_observed = dialer_observed.with(Protocol::P2p(self.peer_id));
+
+        // Send SYNACK: our identity + the dialer's address as we observe it.
         let synack = encode_synack(
-            &observed_multiaddr,
+            &dialer_observed,
             &local_peer,
             self.identity.node_type(),
             self.identity.welcome_message().unwrap_or_default(),
--- a/crates/swarm/node/Cargo.toml
+++ b/crates/swarm/node/Cargo.toml
@@ -79,8 +79,10 @@ vertex-swarm-localstore = { workspace = true, optional = true }
 vertex-swarm-redistribution = { workspace = true, optional = true }
 
 [dev-dependencies]
-tokio = { workspace = true, features = ["rt", "macros"] }
-vertex-swarm-test-utils = { workspace = true }
+tokio = { workspace = true, features = ["rt", "rt-multi-thread", "macros", "time", "sync"] }
+vertex-swarm-test-utils = { workspace = true, features = ["cluster"] }
+eyre.workspace = true
+tracing-subscriber = { workspace = true, features = ["env-filter"] }
 
 [features]
 default = ["std"]
--- a/crates/swarm/node/tests/bootnode_cluster.rs
+++ b/crates/swarm/node/tests/bootnode_cluster.rs
@@ -1,0 +1,211 @@
+//! Integration test: 1 vertex bootnode + 2 vertex clients in-process.
+//!
+//! Asserts convergence:
+//!  (a) Both clients complete a handshake with the bootnode.
+//!  (b) Clients learn each other's overlay via hive gossip from the bootnode.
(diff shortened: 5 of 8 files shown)


mfw78:
test

mfw78:
**GLOBAL_EXECUTOR holds a stale tokio Handle across `#[tokio::test]` runs**

`TaskManager::new` writes the executor into a `OnceLock` (`GLOBAL_EXECUTOR`) on the first call and never replaces it. `cluster::build` here only creates a new `TaskManager` when `try_current()` errors, i.e. when nothing has populated `GLOBAL_EXECUTOR` yet.

`cargo test` compiles all `#[tokio::test]`s in a crate into a single binary. The first test that runs `cluster::build` installs a `TaskManager` keyed on *that test's* runtime `Handle`. When that runtime is dropped at end-of-test and a second `#[tokio::test]` calls `cluster::build`, `try_current()` returns `Ok` (the `OnceLock` is still set), so no fresh `TaskManager` is created. `spawn_node_task` then calls `TaskExecutor::current()` and uses a `Handle` for a runtime that no longer exists ⇒ `Handle::spawn` panics with `tried to spawn on a runtime that has shut down`.

Today there is only one test in the crate, but the harness is advertised as reusable and the next test added to the same binary will trip this. Either:

- expose a `cluster::build`-side `graceful_shutdown_with_timeout` that runs on `Drop` of `Cluster` so the per-test `TaskManager` actually fires (and a future PR can teach `vertex-tasks` to clear `GLOBAL_EXECUTOR`), or
- have `cluster::build` refuse to reuse an existing `GLOBAL_EXECUTOR` whose `Handle` belongs to a different runtime (compare `Handle::current()` to the cached one and create a new manager regardless), or
- document a single-test-per-binary requirement (one `tests/<name>.rs` integration target per cluster test).

mfw78:
`crates/swarm/test-utils/src/cluster.rs` L464-472 — **`spawn_node_task` swallows panics from the node `run` loop**

`spawn_with_graceful_shutdown_signal` spawns a *regular* (non-critical) task — the inner future is not `catch_unwind`'d, so if `bootnode.run(graceful)` or `client.run(graceful)` panics the spawned task aborts and `spawn` resolves to `Err(JoinError::is_panic)`. This bridge then does `let _ = spawn.await;` and immediately falls through to `result_rx.await`, which errors because `result_tx` was dropped without sending.

The cluster test surfaces the panic as

> "cluster-bootnode task dropped without producing a result"

rather than the actual panic message and backtrace. A test harness whose entire point is debugging node behaviour should *propagate* panics, not erase them. Two reasonable fixes:

- inspect the `JoinError`: `match spawn.await { Err(je) if je.is_panic() => Err(eyre::eyre!("{name} panicked: {:?}", je.into_panic())), _ => result_rx.await.unwrap_or_else(...) }`, or
- replace the manual bridge with `spawn_critical_with_graceful_shutdown_signal`, which already wraps the future in `catch_unwind` and reports via `TaskEvent::Panic`.

Bonus: today a panic in the node loop also leaves `Cluster::shutdown` waiting up to 5 s on the bridge timeout for each node, slowing down failing tests.

mfw78:
`crates/swarm/node/tests/bootnode_cluster.rs` L179 — **`bootnode_topo.subscribe()` is racy against `PeerReady`**

The cluster builder calls `kick_bootnode_dial`, which queues `connect_bootnodes()` on each client and returns. By the time `cluster.build().await` resolves, the libp2p dial+handshake may already be in flight on the runtime, and `TopologyBehaviour` will emit `PeerReady` on the bootnode the moment the handshake verifies.

`broadcast::Sender::subscribe` only delivers events sent **after** the receiver is created — there is no replay. If both client handshakes complete in the window between `cluster.build().await?` returning and this `bootnode_topo.subscribe()` line being reached (entirely plausible on fast loopback), `seen_handshakes` will sit at 0 for the full `HANDSHAKE_TIMEOUT` and the test fails with the misleading "timed out … (saw 0 of 2)" message.

Subscribe *before* the dial is kicked. Concretely: take the subscription inside `ClusterBuilder::build` (or expose `bootnode_topology()` on the partially-built cluster) and store the `Receiver` on `Cluster` / pass it back to the caller. Then `connect_bootnodes()` and the subscriber start in the correct order.

mfw78:
`crates/swarm/test-utils/src/cluster.rs` L343-352 (`kick_bootnode_dial`) — **Fixed 250 ms sleep before `connect_bootnodes` is racy and pays a cost on every build**

This loop unconditionally adds 250 ms to every cluster build (5 × 50 ms) and is the only thing standing between the client and a "no reachable addresses" rejection from `vertex_net_local::is_dialable`. Two problems:

1. **It's still a race.** Under load (CI with limited CPU, or with `target/debug` + sanitizers), 250 ms may not be enough for libp2p to drain `NewListenAddr`. The redial is silently swallowed (`let _ = client.topology.connect_bootnodes().await;`), so a slow CI run shows up as a "missing handshake" timeout 10 s later, not as a clear "redial failed" error.
2. **It's tax we always pay.** Every `cluster::build()` adds 250 ms even when listen addresses were observed instantly.

The robust fix is to subscribe to topology events (or expose a `wait_for_listen_addr` on the topology handle) and continue as soon as at least one listen address is reported, with a generous wall-clock bound for the CI case. Even just checking `client.topology.listen_addrs().is_empty()` in a `tokio::time::timeout(2 s, async { ... })` poll loop would be both faster on the common path and less fragile.

A "won't fix until the underlying builder is fixed" comment is fine, but please bound the sleep with a *condition*, not a constant.

mfw78:
`crates/swarm/test-utils/src/cluster.rs` L362-371 (`reserve_ephemeral_port`) — **TOCTOU window between `drop(listener)` and libp2p re-bind**

The bind-drop-rebind pattern races every other process *and itself* — for `with_clients(N)` the harness calls this once per node sequentially, so two consecutive calls can return the same port if the OS recycles it (Linux typically does not for ephemeral range, but it is allowed). On loopback under heavy parallel `cargo test` invocations the failure mode is "client #2 fails to bind because the bootnode already grabbed that port".

Cleaner alternatives:

- bind the libp2p listener directly with port `0` and read the OS-assigned port from the first `NewListenAddr` event (this is the canonical libp2p test pattern); or
- hold the `TcpListener` until the moment libp2p is told to listen, then `drop` immediately before calling `start_listening` (still racy, but races only against external processes).

At minimum, surface the bind failure inside `start_listening()` and don't proceed — right now `client.start_listening()?` will error and bubble up, but only *after* the harness has already torn down its placeholder listener, so the diagnostic doesn't tell you the port was stolen.

mfw78:
**F1 (GLOBAL_EXECUTOR stale Handle): deferred to a follow-up.**

Fixing this requires changing the `vertex-tasks` API so `TaskManager::current()` either compares against the live runtime `Handle` or exposes a way to clear `GLOBAL_EXECUTOR` between tests. That is a non-trivial change to a primitive that is consumed across the whole codebase and warrants its own PR rather than being smuggled in here.

The current single-test scope of `tests/bootnode_cluster.rs` is safe today. I have noted the multi-test reuse hazard in the harness module docs so a future second-test-in-the-binary author hits the documented workaround (drop `Cluster` before the runtime tears down) rather than the panic.

Filing as a separate issue.

mfw78:
**F2 (`spawn_node_task` swallows panics): addressed.**

`spawn_node_task` now inspects `JoinError` and surfaces panic messages as the bridged error, rather than masking them as "task dropped without producing a result". The `Err(JoinError::is_panic)` arm extracts the payload (`&'static str` / `String` if available) and wraps it in an `eyre::eyre!` with the task name. See `crates/swarm/test-utils/src/cluster.rs` in 43c97a8.

mfw78:
**F3 (subscribe race against `PeerReady`): addressed.**

`ClusterBuilder::build` now subscribes to the bootnode topology event stream **before** `kick_bootnode_dial` runs, and stashes the receiver on `Cluster` (`bootnode_events` field). The integration test takes it via the new `Cluster::take_bootnode_events()` API, so subscription is guaranteed to happen before any handshake can race ahead. See 43c97a8.

mfw78:
**F4 (fixed 250 ms sleep is a constant): addressed.**

`kick_bootnode_dial` replaced the fixed `sleep(50ms) × 5` loop with a condition-bounded poll. It now retries `connect_bootnodes` until either every client reports `connected_peers_count() > 0` or a 2-second wall-clock budget elapses, with 25 ms between attempts. Fast on the common path (returns as soon as the handshakes complete) and bounded by a real condition under load. See 43c97a8.

mfw78:
**F5 (TOCTOU port reservation): partially addressed.**

The inter-sibling race is fixed: `ClusterBuilder::build` now reserves every port upfront via a `PortReservation` (holding the placeholder `TcpListener`) and only drops each reservation immediately before that node's libp2p listener binds. While we are starting node N, every other node's placeholder is still held, so two siblings cannot collide on the same OS-assigned port even if the kernel recycles it.

The residual race is only against **external** processes (a parallel `cargo test` invocation in a different repo, for example). The canonical `tcp/0` + read-`NewListenAddr` fix would require the node builder to expose its actual post-bind listen addresses through a public API; that is a separate refactor.

See 43c97a8.

PR #146, opened 2026-06-07
docs(agents): codify AGENTS.md, deep guides, and per-area files

mfw78:
## Summary

- Adds a canonical `AGENTS.md` at repo root plus five topic guides under `docs/agents/` (`swarm-protocol`, `rust-idiomatic`, `libp2p-networking`, `wasm`, `api-surface`).
- Adds per-area `AGENTS.md` under `crates/{net,swarm,swarm/net,storage,node,observability}`, `bin/vertex`, and `docs/`. Each is 30-50 lines and reinforces the top-level rules with local context.
- `CLAUDE.md` is a symlink to `AGENTS.md` at every level so Claude Code reads the same instructions every other agent does (Codex, Cursor, OpenHands, humans).
- Adds the Book of Swarm text extract under `docs/swarm/reference/book-of-swarm.txt` for greppable conceptual reference. The PDF is gitignored.

## What's codified

- **Process**: a start-of-task checklist (classify the change, read the relevant guidance, refine scope and spec before code, update AGENTS.md in the same PR if guidance is stale, then implement and verify).
- **V1 wire conformance vs experimentation**: the contract for matching the reference implementation byte-for-byte on the live testnet, with `SwarmHardfork` and `ForkDigest` as the gating mechanism for any divergence.
- **Repo split**: primitives and layer-2 (chunks, BMT, addresses, manifests, feeds, postage, erasure) belong in `nxm-rs/nectar`; vertex owns the node, libp2p stack, storage backend, and binary. Both are under our control so movement across the boundary is a same-org PR.
- **Rust idioms**: prefer `thiserror`, `strum`, `derive_more`, and `auto_impl(&, Arc, Box)` over hand-rolled impls. Sealed traits, type-state, const generics. Anti-Go-ism checklist.
- **libp2p discipline**: per-protocol `NetworkBehaviour`, codecs separate from behaviours, `PeerId` never above the `vertex-swarm-node` boundary, conformance vectors as the wire contract.
- **Wasm**: a `client` node type will run in `wasm32-unknown-unknown`. Crate-by-crate cone, cfg-gating patterns (target_arch vs cargo feature), tokio feature hygiene, transport and storage plan.
- **API surface**: FFI primary (Dart bindings and similar), gRPC for desktop and server operator scripting, wasm-bindgen for browsers. No HTTP+JSON, no `openapi.yml`. This is the deliberate departure from the reference implementation.
- **Workflow rules**: no em-dashes, no `underlay` term, no inline reference-implementation refs in code or operator docs, no Claude attribution in commits, `cargo fmt` + `cargo clippy` required pre-commit, `gh pr checks` after every push.

## Test plan

- [ ] `cargo build --workspace` succeeds (no code changes, but verify the workspace still builds)
- [ ] `cargo test --workspace` is green
- [ ] `CLAUDE.md` symlinks resolve correctly when an agent opens a file in each subdirectory
- [ ] `docs/swarm/reference/book-of-swarm.pdf` does not appear in `git status` (covered by `.gitignore`)

### DIFF
--- a/crates/swarm/AGENTS.md
+++ b/crates/swarm/AGENTS.md
@@ -7,7 +7,7 @@ Root-level rules in `/AGENTS.md` apply here too. The notes below are the area-sp
 ## Sub-area map
 
 - `primitives`, `forks`, `spec`, `identity`: pure data and configuration. `nectar-primitives` is the canonical source for chunk and address types.
-- `peers/peer`: the `SwarmPeer` record with EIP-191 handshake signature.
+- `peers/peer`: the `SwarmPeer` record with the EIP-191 handshake signature (distinct from postage EIP-191 signing, which lives upstream in `nectar-postage`). See the follow-up tracked at `nxm-rs/vertex` for extracting the sign-data primitive to nectar.
 - `peers/peer-manager`, `peers/peer-score`: lifecycle, persistence, scoring.
 - `bandwidth/{core,pricing,pseudosettle}`: per-peer balances in Accounting Units (AU). `swap` and `chequebook` are workspace-disabled.
 - `api`: the trait chain `SwarmPrimitives` to `SwarmNetworkTypes` to `SwarmClientTypes` to `SwarmStorerTypes`. Strictly libp2p-free with the documented `Multiaddr` exception.
@@ -23,7 +23,18 @@ Root-level rules in `/AGENTS.md` apply here too. The notes below are the area-sp
 - libp2p-free: `primitives`, `forks`, `spec`, `identity`, `api` (apart from the `Multiaddr` re-export), `builder` (almost), `bandwidth/*`, `localstore`, `redistribution`, `storer`, `rpc`, `test-utils`, `peers/peer-manager`, `peers/peer-score`.
 - libp2p-aware: `peers/peer` (uses `Multiaddr`, `PeerId`), `node`, `topology`.
 
-When in doubt: if you need a `NetworkBehaviour` or a `Swarm`, you are in `swarm/node` or `swarm/topology`. If you need only `Multiaddr` or `PeerId` you can be elsewhere, but justify it.
+When in doubt: if you need a `NetworkBehaviour` or a `Swarm`, you are in `swarm/node`, `swarm/topology`, or a future composite-behaviour crate for a node type. If you need only `Multiaddr` or `PeerId` you can be elsewhere, but justify it.
+
+### Composite behaviours for node types
+
+`swarm/topology` is the model for a composite `NetworkBehaviour` that owns several sub-protocols (kademlia routing, hive gossip, NAT discovery, reachability). The same pattern is planned for the Client and Storer node types: a single composite behaviour that brings together pricing, swap, pseudosettle, pushsync, retrieval, and similar role-specific protocols, exposed through one type to the libp2p `Swarm` in `swarm/node`.
+
+When such a crate lands (`vertex-swarm-client-behaviour`, `vertex-swarm-storer-behaviour`, or similar), it follows the same rules as `topology`:
+
+- Owns its sub-protocols, composes them with `#[derive(NetworkBehaviour)]`, exposes a single event enum.
+- Does not depend on a specific `Swarm` configuration; `swarm/node` selects the composite for the node type.
+- Lives in `crates/swarm/` (libp2p-aware tier), not in `crates/swarm/net/` (per-protocol tier).
+- Protocol crates under `crates/swarm/net/` stay one-protocol-per-crate. Composition happens here.
 
 ## Nectar boundary
 
--- a/crates/swarm/AGENTS.md
+++ b/crates/swarm/AGENTS.md
@@ -1,0 +1,53 @@
+# AGENTS: crates/swarm/
+
+Swarm domain layer. Defines what a Swarm node is, the network spec, identity, peer records, bandwidth accounting, peer manager, topology, localstore and storer config, redistribution config, and the RPC service surface. The libp2p-aware composition lives in `swarm/node` and `swarm/topology`; the rest is protocol-agnostic.
+
+Root-level rules in `/AGENTS.md` apply here too. The notes below are the area-specific overlay. The wire-protocol crates under `crates/swarm/net/` have their own `AGENTS.md`.
+
+## Sub-area map
+
+- `primitives`, `forks`, `spec`, `identity`: pure data and configuration. `nectar-primitives` is the canonical source for chunk and address types.
+- `peers/peer`: the `SwarmPeer` record with EIP-191 handshake signature.
+- `peers/peer-manager`, `peers/peer-score`: lifecycle, persistence, scoring.
+- `bandwidth/{core,pricing,pseudosettle}`: per-peer balances in Accounting Units (AU). `swap` and `chequebook` are workspace-disabled.
+- `api`: the trait chain `SwarmPrimitives` to `SwarmNetworkTypes` to `SwarmClientTypes` to `SwarmStorerTypes`. Strictly libp2p-free with the documented `Multiaddr` exception.
+- `builder`: layered builders that produce `BuiltBootnode`, `BuiltClient`, `BuiltStorer`.
+- `node`: composes the libp2p `NetworkBehaviour` and exposes `BootNode`, `ClientNode`, `StorerNode`. This is where libp2p shows up.
+- `topology`: libp2p behaviour for peer discovery, kademlia routing, reachability tracking.
+- `localstore`, `storer`, `redistribution`: storer-side configuration and the chunk-store/reserve implementation.
+- `rpc`: tonic-generated gRPC services and a `GrpcServiceProvider` trait.
+- `test-utils`: `MockIdentity`, `MockTopology`, and fixtures.
+
+## libp2p dependency
+
+- libp2p-free: `primitives`, `forks`, `spec`, `identity`, `api` (apart from the `Multiaddr` re-export), `builder` (almost), `bandwidth/*`, `localstore`, `redistribution`, `storer`, `rpc`, `test-utils`, `peers/peer-manager`, `peers/peer-score`.
+- libp2p-aware: `peers/peer` (uses `Multiaddr`, `PeerId`), `node`, `topology`.
+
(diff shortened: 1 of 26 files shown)


crates/swarm/AGENTS.md
```diff
@@ -0,0 +1,53 @@
...
+- `localstore`, `storer`, `redistribution`: storer-side configuration and the chunk-store/reserve implementation.
+- `rpc`: tonic-generated gRPC services and a `GrpcServiceProvider` trait.
+- `test-utils`: `MockIdentity`, `MockTopology`, and fixtures.
+
+## libp2p dependency
+
+- libp2p-free: `primitives`, `forks`, `spec`, `identity`, `api` (apart from the `Multiaddr` re-export), `builder` (almost), `bandwidth/*`, `localstore`, `redistribution`, `storer`, `rpc`, `test-utils`, `peers/peer-manager`, `peers/peer-score`.
+- libp2p-aware: `peers/peer` (uses `Multiaddr`, `PeerId`), `node`, `topology`.
+
+When in doubt: if you need a `NetworkBehaviour` or a `Swarm`, you are in `swarm/node` or `swarm/topology`. If you need only `Multiaddr` or `PeerId` you can be elsewhere, but justify it.
```
mfw78: Umm, but for this it could also be the case in the future for the 'Storer' and 'Client' types of nodes that there will be a need for 'NetworkBehaviour'. ie. for the 'Client' types of nodes, these will be similar to the 'Topology', in that it will bring together multiple different types of behaviours / protocols (i.e. pricing, swap, pushsync, retrieval). So in effect, we should be prepared for that development in the AGENTS.md for the future when it is done.

crates/swarm/AGENTS.md
```diff
@@ -0,0 +1,53 @@
+# AGENTS: crates/swarm/
+
+Swarm domain layer. Defines what a Swarm node is, the network spec, identity, peer records, bandwidth accounting, peer manager, topology, localstore and storer config, redistribution config, and the RPC service surface. The libp2p-aware composition lives in `swarm/node` and `swarm/topology`; the rest is protocol-agnostic.
+
+Root-level rules in `/AGENTS.md` apply here too. The notes below are the area-specific overlay. The wire-protocol crates under `crates/swarm/net/` have their own `AGENTS.md`.
+
+## Sub-area map
+
+- `primitives`, `forks`, `spec`, `identity`: pure data and configuration. `nectar-primitives` is the canonical source for chunk and address types.
+- `peers/peer`: the `SwarmPeer` record with EIP-191 handshake signature.
```
mfw78: We should check if this should still be the case given that some of the EIP-191 stuff was handed off to nectar. As part of this, should review / check it, and if it is the case, then a follow-up task / issue should be filed to remove the relevant section from vertex.

docs/agents/libp2p-networking.md
```diff
@@ -0,0 +1,70 @@
...
+- Per-protocol `Codec` impls live next to the protocol: in the protocol crate (e.g. `crates/swarm/net/handshake/src/codec/`) or in `vertex-net-codec` when truly generic. Do not invent a new codec where `quick_protobuf_codec::Codec::<Msg>::new(MAX)` suffices (see `crates/swarm/net/identify/src/protocol.rs:85`).
+- All protocols except `handshake` are headered streams: the headers exchange runs first, then protocol data. Build new headered protocols on `vertex-swarm-net-headers`' `InnerInbound`, `InnerOutbound`, `HeaderedInbound<T>`, `HeaderedOutbound<T>` (`crates/swarm/net/headers/src/lib.rs`). See `docs/swarm/protocols.md`.
+
+### 4. PeerId vs OverlayAddress
+
+`PeerId` is libp2p transport identity. `OverlayAddress` is the Swarm Kademlia key derived from the Ethereum key. The mapping is owned by `vertex-swarm-node` (with the `PeerRegistry` in `vertex-net-peer-registry`). Never expose `PeerId` above `vertex-swarm-node`: traits in `vertex-swarm-api`, accounting, storer, RPC, and CLI all speak `OverlayAddress`. See `docs/client/architecture.md` for the dependency graph and `crates/swarm/peers/peer-manager/src/manager.rs` for the `DashMap<OverlayAddress, Arc<PeerEntry>>` model.
```
mfw78: Revise what the 'OverlayAddress' is as it's not derived solely from the Ethereum key.

docs/agents/libp2p-networking.md
```diff
@@ -0,0 +1,70 @@
...
+
+### 7. Address management
+
+`vertex-net-local` owns IP scope classification (`loopback`, `private`, `link-local`, `public`), dual-stack capability, and same-subnet detection via `netdev` (cached 60s). `vertex-swarm-topology`'s `LocalAddressManager` (in `nat_discovery`) selects addresses for handshake advertisement, sourcing from `--nat-addr` (highest priority) and listen addresses. Addresses are always `libp2p::Multiaddr` with a trailing `/p2p/{local_peer_id}`. Observed addresses from identify are used only to flip a public-connectivity flag; do not store or advertise NAT-mapped ephemeral ports. See `docs/networking/address-management.md`.
+
+### 8. Handshake and headers gotchas
+
+Handshake is the only non-headered Swarm protocol; it speaks SYN/ACK directly. Every other protocol must complete the headers exchange before any protocol payload moves. When writing a new protocol: wrap the inner codec in `HeaderedInbound<T>` / `HeaderedOutbound<T>`. Handler output arrives as `HeaderedInboundOutput<T>` with a `.data` field. The custom `0x99` multi-multiaddr encoding lives in `vertex-swarm-peer` and is not standard libp2p; do not propagate it into other protocols.
```
mfw78, line 52: OK, the custom `0x99` is some bee-specific ism, and when there are bee specific isms, which are generally around serialisation strategy, when the wire level differs from what would be the idiomatic protobut encoding, such as jamming in JSON that gets serialised to protobuf bytes, or doing custom list serialisation, such as this case, then we should implement the same wiretype as bee to ensure compatibility, but there should be notes in the rustdoc / lib.rs / mod.rs (or whatever specifically) that allows for an easy grep across the code base to group these isms so that we can progressively file upstream corrections / SWIPs to rectify these code smells. We should absolutely document and make it obvious in code that the implementation is a bee code smell, that is required for wire-level compatibility, but we should file an issue for the SWIPs (https://github.com/ethersphere/SWIPs), with a copy that we have at https://github.com/nxm-rs/SWIPs). We don't want to get jammed on trying to get everything done upstream first and then implement as Ethereum Swarm are just too slow. We need to push fast.

crates/swarm/AGENTS.md
mfw78: Addressed in ccac1c6.

Added a "Composite behaviours for node types" subsection that codifies `swarm/topology` as the model for future Client and Storer composite NetworkBehaviours bringing together pricing, swap, pseudosettle, pushsync, retrieval, and similar role-specific protocols. Rules: each composite lives in `crates/swarm/` (libp2p-aware tier, not `crates/swarm/net/`), owns its sub-protocols via `#[derive(NetworkBehaviour)]`, exposes a single event enum, and `swarm/node` selects which composite the node type uses. Protocol crates under `crates/swarm/net/` stay one-protocol-per-crate; composition happens here.

The "When in doubt" rule was widened: `swarm/node`, `swarm/topology`, or a future composite-behaviour crate for a node type are all valid homes for a `NetworkBehaviour`.

crates/swarm/AGENTS.md
mfw78: Verified and addressed in ccac1c6, with follow-up issue #147.

Verification: the handshake EIP-191 signing is still legitimately here. `SwarmPeer` holds the libp2p `Multiaddr` list (via the `0x99` shim), so the struct itself cannot move to nectar without dragging libp2p into nectar's wasm-friendly tier. The postage EIP-191 path is already in `nectar-postage`.

However, the *pure* sign-data construction (byte layout plus EIP-191 prefix, no `Multiaddr` types) is a candidate for extraction so non-node Swarm consumers (Dart clients, light verifiers) can produce and verify peer records without a libp2p stack. Filed #147 to track that refactor.

The AGENTS.md sub-area line now distinguishes handshake EIP-191 (here) from postage EIP-191 (already in `nectar-postage`) and links the follow-up.

docs/agents/libp2p-networking.md
mfw78: Fixed in ccac1c6.

The doc now states the derivation accurately: `OverlayAddress = keccak256(ethereum_address || network_id_le8 || nonce)` via `nectar_primitives::compute_overlay`. Not from the Ethereum key (or address) alone.

Also added the *why*: the same operator identity yields different overlays across networks or with a fresh nonce, which is the property the redistribution game relies on.

docs/agents/libp2p-networking.md
mfw78: Codified in ccac1c6 and anchored in 0e82933.

New "Wire-compat shims and SWIPs" section in `docs/agents/swarm-protocol.md` defines a greppable marker convention and the workflow:

1. Implement the shim byte-faithfully so peers interoperate.
2. Tag the call site, constant, or type with `// BEE-COMPAT(SWIP-NN): <short reason>`. Use `BEE-COMPAT(TBD)` if the SWIP candidate has not been filed yet, and file before merging. `grep -rn "BEE-COMPAT"` across the workspace must surface every shim.
3. File a SWIP candidate at `nxm-rs/SWIPs` (currently tracked in `nxm-rs/vertex` while the SWIPs repo issue tracker is disabled). After internal review, mirror upstream at `ethersphere/SWIPs`. Do not block Vertex work waiting for upstream review.
4. Gate the corrected wire behind a new `SwarmHardfork` variant; `ForkDigest` flips the encoding at activation.
5. When the hardfork ships, remove the `BEE-COMPAT` marker.

Anchored the convention: added `BEE-COMPAT(SWIP-148)` markers at the `0x99` multi-multiaddr prefix in `crates/swarm/peers/peer/src/serde_multiaddr.rs`, and filed #148 as the first SWIP candidate (replace `0x99`-prefixed in-band layout with `repeated bytes`). The section also calls out the variable-width `big.Int.Bytes()` encoding in `vertex-swarm-net-pricing` as the next candidate.

Reviewers should now reject any new wire-compat shim that arrives without a `BEE-COMPAT` marker and a linked SWIP candidate.

Issue #134, opened 2026-06-06
topology: wire nectar's Bin / NeighborhoodDepth into ProximityIndex and routing


mfw78:
## Context

Nectar at the current pin (3c99edc) exposes typed primitives and the canonical neighborhood-depth math:

- `nectar_primitives::Bin` - typed kademlia bin index, 0..=MAX_PO.
- `nectar_primitives::ProximityOrder` - typed proximity metric.
- `nectar_primitives::recompute_neighborhood_depth(&[u8; 32], saturation, low_watermark) -> Bin` - pure-function port of bee's `recalcDepth`.

Per nectar's own docs: "The wrapper type (`NeighborhoodDepth`) stays in the routing layer of each downstream impl; nectar only owns the math."

Vertex main still keys `ProximityIndex` by raw `u8` and recomputes depth inline. This issue tracks switching to the typed nectar surface.

## Scope

1. `crates/swarm/peers/peer-manager/src/proximity_index.rs`: rekey `ProximityIndex` from `u8` to `nectar_primitives::Bin`. Update all callers (`peer-manager/src/lib.rs`, `topology/src/kademlia/routing.rs`, `topology/benches/topology_benchmarks.rs`, `api/src/components/topology.rs`).

2. Add a `NeighborhoodDepth(Bin)` newtype in `crates/swarm/topology/src/kademlia/` with a `recompute(connected_per_bin, saturation, low_watermark) -> Self` constructor that delegates to `recompute_neighborhood_depth`. Document the routing-layer ownership of the wrapper type in the module docs.

3. Replace the existing depth-recompute math in `kademlia/routing.rs` with calls to the new wrapper.

4. Thread `Bin` and `ProximityOrder` through the relevant APIs where they currently use raw `u8` (e.g. `fn proximity(&self, peer: &OverlayAddress) -> u8` in `behaviour.rs:632` and `routing.rs:124`).

## Acceptance criteria

- [ ] `ProximityIndex` keyed by `Bin`; no raw `u8` in its public surface.
- [ ] `NeighborhoodDepth(Bin)` wrapper in the topology routing layer.
- [ ] Depth recomputation delegates to `nectar_primitives::recompute_neighborhood_depth`.
- [ ] Vertex's `proximity` accessors return typed `ProximityOrder`, not `u8`.
- [ ] Existing tests pass; no behavioural change beyond the type swap.

## Supersedes

- #107 (the original attempt, written before nectar exposed the typed surface; would have duplicated it).

Issue #55, opened 2026-02-05
Add IP-level abuse tracking to PeerManager


mfw78:
## Summary

IP-level abuse tracking was removed from the PeerManager to simplify the initial implementation. This issue tracks re-adding it with a better design.

## Motivation

IP-level tracking helps detect:
- **Identity cycling attacks**: Malicious actors changing their nonce/overlay while keeping the same IP
- **Distributed abuse patterns**: Multiple banned overlays originating from the same IP
- **Connection flooding**: Excessive connection attempts from a single IP

## Previous Implementation (Removed)

The previous implementation had several code smells:
- `Vec::remove(0)` in hot path - O(n) instead of O(1)
- O(m×n) linear scan when recording overlay bans (iterating all IPs, then all overlays per IP)
- Public fields in `IpScore` struct exposing implementation details
- Cloning entire `Vec<B256>` when returning overlays for an IP

## Suggested Improvements

1. **Use `VecDeque<B256>`** instead of `Vec<B256>` for known_overlays (O(1) eviction)
2. **Add reverse index** `HashMap<B256, HashSet<IpAddr>>` for O(1) overlay→IP lookup
3. **Private fields** with accessor methods to maintain invariants
4. **Return iterators** instead of cloned collections where possible
5. **Consider integration with libp2p's connection limits** for a more holistic approach

## Acceptance Criteria

- [ ] IP tracking detects identity cycling (same IP, different overlays)
- [ ] Efficient O(1) operations for hot paths
- [ ] Configurable thresholds for ban decisions
- [ ] Metrics/observability for IP-level abuse patterns
- [ ] Unit tests covering attack scenarios

## References

- Original implementation: commit that removes IP tracking (this PR)
- Related: peer scoring in vertex-net-peers

PR #149, opened 2026-06-07
fix(topology): correct neighborhood depth and converge aggressively

mfw78:
## What

Audit of the Kademlia topology against the reference and the Book of Swarm (ch 2.1.3, 2.1.4, 2.2.4, 3.3.4) found the neighborhood-depth computation diverged from the spec, with knock-on effects on hive gossip and (future) storage responsibility. This corrects it and makes bootstrap converge aggressively.

## The bug

`KademliaRouting::recalc_depth` returned the *deepest* bin with `>= nominal` connected peers. It never inspected shallower bins, so a populated deep bin reported a too-deep depth even when shallower bins were empty or unsaturated. That violates the "no empty bin below depth" invariant that thin-table routability depends on, and a wrong depth feeds hive gossip neighbour classification (`proximity >= depth`) and the neighbourhood-vs-balanced split.

## The fix

- Delegate depth to `nectar_primitives::recompute_neighborhood_depth` (the canonical shallowest-unsaturated-frontier + low-watermark-anchor algorithm). The empty-bin clause now falls out for free: an empty bin is unsaturated, so it caps the frontier.
- Drive `saturation` and `low_watermark` from `SwarmSpec` (`saturation_peers` = 8, `neighborhood_low_watermark` = 3 by default), so they are fork/spec-selectable.
- Replace the `depth == 0 ⇒ nominal` allocation cap with a bounded aggressive bootstrap fill toward `bootstrap_target` (18). Using saturation 8 with the old cap deadlocked: no bin could exceed `nominal` while depth was 0, so no bin could reach 8 and depth was pinned at 0 forever. The bounded fill lets bins reach the frontier fast and depth climb in real steps, without exposing a not-yet-established node to unbounded inbound.
- Add a periodic oversaturation prune. Depth-increase events already trim eagerly; this reclaims surplus that accrues in below-depth bins while depth is stable (e.g. from inbound connections), matching the reference network's timer-driven prune.

## Tests

- `test_depth_caps_at_gap`: bin 0 saturated, bins 1..8 empty, bin 8 populated -> depth caps at 1 (old scan reported 8).
- `test_depth_climbs_with_saturated_bins`: bins 0..2 saturated + bin 3 at the watermark -> depth 3.
- `test_depth_stays_zero_below_saturation`: a few peers in one mid bin -> depth stays 0.
- Existing capacity/eviction tests updated for the new depth-0 regime; `bootnode_cluster` convergence test still green.

## Scope notes

- Resolves the periodic-prune gap (audit Finding 2).
- Advances #134 by delegating the depth math to nectar; the remaining typed-surface rekey (`ProximityIndex` keyed by `Bin`, `NeighborhoodDepth` newtype, typed `ProximityOrder`) is a mechanical follow-up tracked separately.

### DIFF
--- a/crates/swarm/api/src/spec.rs
+++ b/crates/swarm/api/src/spec.rs
@@ -7,6 +7,21 @@ use nectar_primitives::{ChunkTypeSet, NetworkId};
 use nectar_swarms::{NamedSwarm, Swarm};
 use vertex_swarm_forks::{ForkCondition, ForkDigest, SwarmHardfork, SwarmHardforks};
 
+/// Default per-bin saturation target driving the neighborhood-depth frontier.
+///
+/// A bin with fewer than this many connected peers marks the shallowest
+/// unsaturated frontier in [`recompute_neighborhood_depth`]. Matches the live
+/// reference network so depth tracks node-for-node.
+///
+/// [`recompute_neighborhood_depth`]: nectar_primitives::recompute_neighborhood_depth
+pub const DEFAULT_SATURATION_PEERS: u8 = 8;
+
+/// Default cumulative count of peers in the deepest bins anchoring the
+/// neighborhood in [`recompute_neighborhood_depth`].
+///
+/// [`recompute_neighborhood_depth`]: nectar_primitives::recompute_neighborhood_depth
+pub const DEFAULT_NEIGHBORHOOD_LOW_WATERMARK: u8 = 3;
+
 /// Parser for Swarm network specifications.
 ///
 /// Handles both preset names ("mainnet", "testnet") and file paths via a single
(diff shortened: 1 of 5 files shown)


mfw78:
Is a periodic pruner really the most optimal / suitable way to go for this?

mfw78:
Good call, you're right that it isn't. I traced what can actually oversaturate a *below-depth* bin (the only bins `eviction_candidates` touches, it scans `0..depth`):

- outbound stops at `target` (`needs_more` = `connected < target`)
- inbound stops at `target + inbound_headroom` (`should_accept_inbound`)

So a below-depth bin is already bounded at `target + headroom`; there's no unbounded accumulation while depth is stable. The one genuine oversaturation event is a **depth increase** (a neighbourhood bin, previously unbounded, becomes below-depth with a small tapered target), and that's already handled eagerly by the existing `new_depth > old_depth` trigger.

That makes the timer the wrong tool on two counts:

1. It mostly fights the intended `headroom` slack. `surplus` is measured against `target`, so it trims a stable bin from `target + headroom` down to `target`; inbound refills to `target + headroom`; next tick trims again -> connect/evict churn on peers the headroom buffer deliberately admitted.
2. It doesn't address the one genuinely-unbounded surface anyway: neighbourhood bins (`>= depth`, `target = usize::MAX`, always-accept). `eviction_candidates` skips those entirely.

So I've dropped the periodic prune. The eager depth-increase trim stays and handles the real event; below-depth bins remain bounded by the inbound ceiling; the headroom stays as intended slack. The neighbourhood-unbounded inbound is a separate, real concern (a neighbourhood inbound cap via the reserved `RejectionReason::NeighborhoodFull`, tied to #55) and not something a below-depth timer solves, so I'll track it there rather than paper over it here.

#149 is now just the depth-correctness fix plus the aggressive bounded bootstrap. Force-pushed.

PR #167, opened 2026-06-08
feat(topology): self-confirm globally-routable IPv6 listen addresses

### LINKED ISSUE #164: AutoNAT v2: prioritize testing/confirming IPv6 external-address candidates
Follow-up to #156 (AutoNAT v2) and #160 (verified-external advertisement, IPv6-first ordering).

## Problem

When AutoNAT v2 verifies our own external-address candidates, the libp2p v2 client (`~/.cargo/git/checkouts/rust-libp2p-*/*/protocols/autonat/src/v2/client/behaviour.rs`) selects which candidate to test purely by observation-frequency score (`untested_candidates` sorts by `score`); it has no address-family preference and exposes only `with_max_candidates` / `with_probe_interval`. IPv6 addresses are usually globally routable, so confirming an IPv6 external address first is more likely to give us a usable, advertisable address sooner, which composes with #160 (which already advertises IPv6-first once confirmed).

## Options (no upstream API for this today)

1. Pre-confirm a globally-routable IPv6 address directly via `ToSwarm::ExternalAddrConfirmed` (the same mechanism the topology behaviour already uses for static NAT addresses), gated on a real signal (a public-scope IPv6 listen address, which is almost always routable), so AutoNAT only needs to verify the harder IPv4/NAT case. Most feasible, reuses existing wiring.
2. Bias the candidate score by emitting IPv6 candidates with higher frequency. Hacky, races with organically-observed IPv4 candidates, non-deterministic.
3. A small custom prioritization behaviour that holds candidates and feeds them to the AutoNAT client IPv6-first. Deterministic but duplicates the client's queue and couples us to libp2p internals.

Recommendation: pursue option 1; if deterministic ordering becomes a hard requirement, prefer a small upstream PR adding a candidate comparator over maintaining option 3 locally.

## Acceptance

- Where we hold both an IPv4 and an IPv6 external-address candidate, the IPv6 one is confirmed/advertised first in the common case.
- No regression to the AutoNAT v2 dial-back protocol or to bee interop (AutoNAT v2 is additive; bee runs v1).

mfw78:
Closes #164

## What

When the topology behaviour sees a `NewListenAddr` carrying a globally-routable IPv6 unicast address it actually binds, it self-confirms that address as external by emitting `ToSwarm::ExternalAddrConfirmed`, the same mechanism already used for static NAT addresses. This implements Option 1 from #164.

## Why

IPv6 has no NAT: a global unicast address bound on a local interface is reachable end-to-end without a dial-back. The only caveat is a stateful firewall dropping unsolicited inbound packets, which is operator configuration and would equally fail an AutoNAT dial-back. This applies only to listen addresses we bind, never to observed or relayed candidates, and only to IPv6: a public-scope IPv4 listen address may still sit behind NAT and is left for AutoNAT to verify.

The effect is that an IPv6 listen address reaches the verified advertisement tier immediately (composing with #160's IPv6-first ordering), letting the AutoNAT v2 client spend its score-ranked candidate budget on the harder IPv4/NAT case. This is the practical lever for #164's IPv6 prioritization goal: the libp2p autonat v2 client selects candidates purely by observation-frequency score and exposes no address-family preference.

## How

New `vertex_net_local::is_globally_routable_ipv6(addr)` returns true only for global unicast IPv6 (false for IPv4 even when public-scope, for non-global IPv6, and for family-less addresses). The topology behaviour calls it in the `NewListenAddr` arm.

## Tests

- `test_is_globally_routable_ipv6` (net-local), plus a doctest
- `global_ipv6_listen_addr_self_confirmed_ipv4_not` (topology): a global IPv6 listen address is self-confirmed while public IPv4, ULA, and unspecified IPv6 are not

`cargo test`: net-local 55 + 3 doctests, topology 137. clippy (lib + tests) and fmt clean.

Docs updated in `docs/networking/address-management.md` (AutoNAT/UPnP section and the IPv6-vs-IPv4 note).

### DIFF
--- a/crates/net/local/src/lib.rs
+++ b/crates/net/local/src/lib.rs
@@ -11,5 +11,6 @@ pub mod system;
 pub use capabilities::{LocalCapabilities, advertise_filter};
 pub use scope::{
     AddressFamily, AddressScope, IpCapability, classify_multiaddr, family_order, is_dialable,
+    is_globally_routable_ipv6,
 };
 pub use system::{add_subnet, remove_subnet, same_subnet};
--- a/crates/net/local/src/scope.rs
+++ b/crates/net/local/src/scope.rs
@@ -85,6 +85,34 @@ pub fn classify_multiaddr(addr: &Multiaddr) -> Option<AddressScope> {
     extract_ip(addr).and_then(classify_ip)
 }
 
+/// Whether a multiaddr is a globally-routable IPv6 unicast address.
+///
+/// IPv6 has no NAT: a global unicast address bound on a local interface is
+/// reachable end-to-end by any peer, so a node listening on one can treat it as
+/// externally confirmed without a dial-back. The only caveat is a stateful
+/// firewall dropping unsolicited inbound packets; that is the operator's
+/// configuration, not an addressing question, and the same firewall would also
+/// fail an AutoNAT dial-back. Apply this only to listen addresses the node
+/// actually binds, never to observed or relayed addresses.
+///
+/// Returns `false` for IPv4 (which may sit behind NAT even when public-scope),
+/// for non-global IPv6 (loopback, unique-local, link-local, unspecified), and
+/// for family-less addresses (DNS).
+///
+/// ```
+/// use vertex_net_local::is_globally_routable_ipv6;
+///
+/// let global: libp2p::Multiaddr = "/ip6/2606:4700:4700::1111/tcp/1634".parse().unwrap();
+/// let ula: libp2p::Multiaddr = "/ip6/fd00::1/tcp/1634".parse().unwrap();
+/// let public_v4: libp2p::Multiaddr = "/ip4/8.8.8.8/tcp/1634".parse().unwrap();
+/// assert!(is_globally_routable_ipv6(&global));
+/// assert!(!is_globally_routable_ipv6(&ula));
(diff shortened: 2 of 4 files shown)


mfw78:
Closing without merging. On review, self-confirming a globally-routable IPv6 listen address as an external address is fail-open and not worth the marginal gain.

A global unicast IPv6 listen address is already advertised before this change: tier 2 of the advertised set is `addresses_for_scope`, which includes every public-scope listen address, and #160 already orders IPv6 ahead of IPv4 within each tier. Peers already receive the IPv6 address as a dial hint.

What this PR added on top was to move that address from the listen tier to the verified tier and mark it dial-back-confirmed when it is not. Stateful firewalls that default-deny inbound IPv6 are common (RFC 6092 recommends exactly that for residential CPE), so a firewalled address would land in the verified tier and stay there. Emitting `ExternalAddrConfirmed` also tells libp2p the address is settled, so the AutoNAT v2 client stops dial-back-testing it, which is the probe that would have caught the firewall. There is no clean expire-on-failed-dial-back path either.

The goal in #164 (advertise/try IPv6 sooner) is substantially met by #160's tier-2 IPv6-first ordering. Faking verification is worse than leaving AutoNAT to verify. Keeping #164 open for a future honest implementation (bias AutoNAT to test IPv6 candidates first while still requiring the dial-back).

PR #171, opened 2026-06-08
feat(swarm-builder): add config-gated download verification wrapper

mfw78:
Adds `VerifyingChunkProvider<P>`, a wrapper that verifies downloaded chunks before returning them.

Content integrity is verified by default (reconstruct the chunk from the retrieved bytes and confirm its derived address matches the requested address); stamp verification is opt-in via `ChunkVerifyConfig { verify_content, verify_stamp }` (default: content on, stamp off). The rationale is that if a downloaded chunk passes content verification it is correct regardless of the stamp.

Self-contained new module `crates/swarm/builder/src/verify.rs` plus the export in `lib.rs`; does not touch `providers.rs`/`launch.rs`/`rpc.rs`/`config.rs` (wired up by a follow-up builder unit). Stamp verification currently returns a clear 'unavailable' error rather than silently passing, since no stamp-signature helper is threaded through yet.

Part of the Client node type work.

### DIFF
--- a/crates/swarm/builder/src/verify.rs
+++ b/crates/swarm/builder/src/verify.rs
@@ -1,29 +1,36 @@
 //! Config-gated verification wrapper for downloaded chunks.
 //!
-//! Content integrity is cheap to check once the bytes are in hand: reconstruct
-//! the chunk from the returned data and confirm its derived address matches the
-//! requested one. This wrapper makes that check the default. Stamp verification
-//! is a separate, opt-in concern and is off by default.
+//! Content integrity is established before a chunk reaches this wrapper: the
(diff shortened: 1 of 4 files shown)


crates/swarm/builder/src/verify.rs
```diff
@@ -0,0 +1,232 @@
...
+}
+
+impl Default for ChunkVerifyConfig {
+    fn default() -> Self {
+        Self {
+            verify_content: true,
+            verify_stamp: false,
+        }
+    }
+}
```
mfw78, line 48: Can the default derivation be done using auto impl and/or derive_more with an annotation for the specific bools for what their defaults should be?

PR #172, opened 2026-06-08
feat(swarm-chequebook): re-enable crate with serde_json-free cheque codec

mfw78:
Re-enables `vertex-swarm-bandwidth-chequebook`, the leaf crate of the SWAP settlement stack. The crate was workspace-disabled because `SignedCheque::to_json`/`from_json` depended on `serde_json`, which is gone from the workspace.

## What changed

The cheque travels on the swap protocol as a JSON object embedded in a protobuf `bytes` field and must stay byte-identical to the live network. Rather than re-add `serde_json` or switch to protobuf, the codec is hand-rolled and fixed-shape.

The encoder emits PascalCase keys in declaration order (`Chequebook`, `Beneficiary`, `CumulativePayout`, `Signature`), lowercase `0x`-hex addresses, a bare decimal JSON number for the payout across the full U256 range, and standard base64 for the signature. The decoder is a strict field extractor, not a general JSON parser, and rejects unknown shapes, quoted payouts, and malformed base64.

Note on the field shape: the live wire format uses PascalCase keys and a bare numeric payout, which is what the conformance vectors pin. The previous `serde(rename)`-driven shape and the `test_json_uses_camel_case` assertion did not match the wire, so that test is replaced by `test_json_field_shape` and the pinned vectors.

`ChequeError` is reworked into precise `thiserror` variants with `strum::IntoStaticStr` for clean `reason` metric labels. `serde` and `serde_json` are dropped from the crate; `alloy-signer`/`sol-types`/`primitives` and `nectar-contracts` are kept. A minimal standard-alphabet base64 implementation is local to the crate to keep the wasm cone dependency-free.

## Wire conformance

`tests/json_conformance.rs` pins the exact JSON bytes for three cheques (sequential signature bytes, max U256 payout, zero payout with empty signature), generated from the reference encoder, plus round-trip and negative-decode tests. Editing these to match a code change would mean the codec is wrong.

The encoder is tagged `BEE-COMPAT(TBD)` since JSON-inside-protobuf is a wire shim to retire upstream once the swap payload moves to a binary codec.

## Scope

This is unit S1 of the SWAP re-enable. Only the chequebook member and dependency lines are uncommented in the root `Cargo.toml`; `net/swap` and `bandwidth/swap` stay disabled for sibling units.

## Verification

- `cargo fmt --all`
- `cargo clippy --all-targets --all-features -- -D warnings` (crate clean)
- `cargo test -p vertex-swarm-bandwidth-chequebook` (14 tests pass)
- `cargo build` (full workspace builds with the crate re-enabled)


mfw78 (commented):
I think we should actually use serde_json for this instead of us reproducing JSON encoding. We should be very careful about using JSON, and only tolerate it when we need to do so for wire equivalence.

mfw78:
Switched the cheque wire codec to `serde_json` over a fixed-order wire struct, dropping the hand-rolled encoder and shape-specific extractor, and replaced the local base64 module with the `base64` crate.

The encoding stays byte-identical to the live network: PascalCase keys in declaration order, lowercase 0x-hex addresses, a bare decimal JSON number for the payout (carried as a `RawValue` so it spans the full 256-bit range), and standard base64 for the signature. The conformance vectors are unchanged and still pass.

JSON is tolerated here only for wire equivalence. I added `serde_json` and `base64` to the workspace and recorded the settled wire-equivalence exception in `crates/swarm/AGENTS.md` so the next agent does not have to relitigate it.

PR #173, opened 2026-06-08
feat(swarm-node): wire chunk push receipts and implement client put/get

mfw78:
Implements the client node upload core: the public push handle method, receipt completion, and real `SwarmClient::put`/`get` bodies.

`ClientHandle::push_chunk` registers a pending push keyed by chunk address, emits `ClientCommand::PushChunk`, and awaits the storer's receipt over a oneshot channel. This mirrors the existing `retrieve_chunk` plumbing, including `complete_push`/`fail_push` alongside `complete_retrieval`/`fail_retrieval`.

`ClientService` now resolves pending pushes. A `ReceiptReceived` event completes the waiter with the receipt fields (signature, nonce, storage radius, peer). A `PushFailed` event surfaces the storer's rejection reason as the new `RetrievalError::PushRejected` variant rather than dropping it, so callers see why a push was rejected.

`SwarmClient::get` selects the closest peers via `topology().closest_to`, retrieves the chunk, decodes it, and verifies the returned bytes hash to the requested address before returning. `SwarmClient::put` selects the closest peer, sends the chunk's serialized bytes together with the pre-computed postage stamp (supplied as the storage proof), and waits for the receipt.

Unit tests cover the push command emission, receipt completion, and rejection paths using a mock mpsc channel.

Known follow-up outside this slice: pushsync outbound failures (error receipts, dial/timeout) are emitted by the connection handler as `ProtocolError`, which does not carry the chunk address, so they cannot resolve a pending push today. The `PushFailed` `ClientEvent` is wired in `ClientService` but is not yet emitted by the behaviour for pushsync failures. Closing that gap requires the handler/behaviour to emit `PushFailed { address, .. }` on pushsync outbound failure, which belongs to the handler plumbing slice.

`get` currently decodes content-addressed chunks only; the address verification guarantees a wrong chunk is never returned. Single-owner retrieval and config-gated download verification are handled by the download slice.

### DIFF
--- a/crates/swarm/node/src/client_service.rs
+++ b/crates/swarm/node/src/client_service.rs
@@ -7,12 +7,13 @@
 use std::collections::HashMap;
 use std::sync::Arc;
 
+use alloy_primitives::Signature;
 use nectar_primitives::ChunkAddress;
 use parking_lot::Mutex;
 use tokio::sync::{mpsc, oneshot};
 use tracing::{debug, warn};
 use vertex_swarm_net_pseudosettle::PaymentAck;
-use vertex_swarm_primitives::OverlayAddress;
+use vertex_swarm_primitives::{Bin, NeighborhoodDepth, Nonce, OverlayAddress};
 use vertex_tasks::{GracefulShutdown, SpawnableTask};
 
 use crate::protocol::{ClientCommand, ClientEvent};
@@ -40,17 +41,17 @@ pub struct RetrievalResult {
 /// Result of a chunk push.
 ///
 /// Carries the receipt fields returned by the storer after a successful
-/// pushsync round-trip.
+/// pushsync round-trip, parsed into their canonical typed forms.
 #[derive(Debug)]
 pub struct PushResult {
     /// The peer that issued the receipt.
     pub peer: OverlayAddress,
-    /// The receipt signature from the storer.
-    pub signature: bytes::Bytes,
-    /// The nonce used in signing the receipt.
-    pub nonce: bytes::Bytes,
-    /// The storer's storage radius.
-    pub storage_radius: u8,
+    /// The receipt signature from the storer over the chunk address and nonce.
+    pub signature: Signature,
+    /// The nonce mixed into the receipt signature.
+    pub nonce: Nonce,
+    /// The storer's storage radius (its neighborhood-depth boundary).
+    pub storage_radius: NeighborhoodDepth,
 }
 
 /// Outcome delivered to a pending push: a receipt on success or an error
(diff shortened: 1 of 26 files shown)


crates/swarm/node/src/client_service.rs
```diff
@@ -36,6 +37,26 @@ pub struct RetrievalResult {
...
+/// Result of a chunk push.
+///
+/// Carries the receipt fields returned by the storer after a successful
+/// pushsync round-trip.
+#[derive(Debug)]
+pub struct PushResult {
+    /// The peer that issued the receipt.
+    pub peer: OverlayAddress,
+    /// The receipt signature from the storer.
+    pub signature: bytes::Bytes,
```
mfw78: This should be done using an alloy signature primitive

crates/swarm/node/src/client_service.rs
```diff
@@ -36,6 +37,26 @@ pub struct RetrievalResult {
...
+    /// The nonce used in signing the receipt.
+    pub nonce: bytes::Bytes,
```
mfw78: If this is a nonce, we should look at how the nonce is actually structured. is this a B256 or similar? We should never be using arbitrary bytes length for this, we should be using types.

crates/swarm/node/src/client_service.rs
```diff
@@ -36,6 +37,26 @@ pub struct RetrievalResult {
...
+    /// The storer's storage radius.
+    pub storage_radius: u8,
```
mfw78: Should the storage radius actually be a separate type? Similar to how we changed ProximityOrder, Bin and NeighborhoodDepth

crates/swarm/node/src/client_service.rs
```diff
@@ -52,6 +73,9 @@ pub enum RetrievalError {
     /// Chunk not found.
     #[error("Chunk not found: {0}")]
     NotFound(ChunkAddress),
+    /// Chunk push was rejected by the storer.
+    #[error("Push failed: {0}")]
+    PushRejected(String),
```
mfw78, line 61: I'm really not sure about having Strings in this. If the string is actually returned in the error in the wire, then I suppose so.

crates/swarm/node/src/swarm_client.rs
```diff
@@ -72,31 +75,80 @@ impl<T, A> Client<ClientComponents<T, A>, ()> {
...
+                        .map_err(|e| SwarmError::InvalidChunk {
+                            address: Some(*address),
+                            reason: e.to_string(),
+                        })?;
+                    return Ok(chunk);
+                }
+                Err(e) => {
+                    last_error = Some(SwarmError::network_msg(e.to_string()));
+                }
+            }
```
mfw78, line 110: Should this be how we actually handle it? This tries nodes sequentially, should we actually not instead run these in parallel or something? Consider libp2p best practices / P2P best practices.

crates/swarm/node/src/swarm_client.rs
```diff
@@ -72,31 +75,80 @@ impl<T, A> Client<ClientComponents<T, A>, ()> {
...
+        let closest = self
+            .components
+            .topology()
+            .closest_to(&address, CLOSEST_PEER_COUNT);
+
+        let peer = closest.into_iter().next().ok_or(SwarmError::NoStorer {
+            chunk_address: address,
+        })?;
+
+        let stamp = bytes::Bytes::copy_from_slice(storage.as_ref());
```
mfw78: Is this how we should actually be doing this? Stamps should be a rust type, no?

PR #174, opened 2026-06-08
refactor(swarm): type stamps, receipts, signatures, nonces, and storage radius

mfw78:
## Why

Domain values must flow as types, never raw `Bytes`/`&[u8]`/`u8`, at Rust API boundaries; raw bytes are tolerated only inside wire `codec.rs` files (and the gRPC byte boundary). Stamps, push receipts, signatures, nonces, and the storage radius were inconsistently raw across the codebase. This PR establishes the typed substrate that the four open PRs (#170-#173) will conform onto.

## StorageRadius vs NeighborhoodDepth

"Storage radius" is a node's reserve / area-of-responsibility radius: demand-side, and a remote value when read from a pushsync receipt. "Neighbourhood depth" is the local node's connectivity boundary: supply-side and local-only. Book of Swarm 3.3.4 confirms the two can diverge. Typing the pushsync receipt's `storage_radius` as `NeighborhoodDepth` would be a category error, so this PR introduces a new, distinct `StorageRadius(Bin)` newtype.

`StorageRadius` mirrors `NeighborhoodDepth`'s shape (`ZERO`, `new(Bin)`, `bin()`, `get()`) but is intentionally not cross-comparable with `Bin` or `NeighborhoodDepth`, and carries no `contains`-style membership method, since that connotation belongs only to the local connectivity boundary. The crate-level "keep the proximity types distinct" doc block now covers all four roles: `ProximityOrder`, `Bin`, `NeighborhoodDepth`, and `StorageRadius`.

## Typed substrate

- `vertex-swarm-primitives` gains `StorageRadius` and re-exports the postage `Stamp` from nectar; `nectar-postage` is wired into the workspace.
- `ChunkRetrievalResult.stamp`, `RetrievalResult.stamp`, and retrieval/pushsync `Delivery.stamp` are now `Stamp`.
- `SwarmChunkSender::send_chunk` / `send_chunk_unchecked` take the chunk and its `Stamp` together (the stamp was previously absent from the trait) and return a `PushReceipt` (renamed from `ChunkSendReceipt`) carrying `storer`, `signature: Signature`, `nonce: Nonce`, and `storage_radius: StorageRadius`.
- `SwarmStorer::storage_radius` returns `StorageRadius` (was `u8`); its misleading "neighborhood depth threshold" comment is fixed.
- Pushsync `Receipt` carries typed `signature: Signature`, `nonce: Nonce`, and `storage_radius: StorageRadius`. Bytes/ints convert to typed values only inside the codec, with precise decode error variants for malformed signature, nonce, and radius.
- The client behaviour command/event plumbing (`ClientCommand`, `ClientEvent`, `HandlerCommand`, `HandlerEvent`) carries typed values end to end, dropping the mid-pipeline `as u32`/`as u8` radius juggling and byte parsing.

Codec error deliveries and error receipts are decodable from their `err` string alone, so a remote error message that leaves the signature/nonce/stamp fields empty still decodes (the typed fields are parsed strictly only on the success path).

## StorageRadius migration follow-up

`StorageRadius` is vertex-only for now and is marked `// TODO(nectar): migrate StorageRadius upstream`. It should move to nectar-primitives alongside the other reserve/responsibility types; an issue can be filed to track that move.

## Downstream

The four open PRs #170-#173 will conform onto this typed substrate.

## Verification

`cargo fmt --all`, `cargo clippy --lib --all-features -- -D warnings`, `cargo clippy --tests --benches --all-features -- -D warnings -A clippy::unwrap_used -A clippy::expect_used -A clippy::indexing_slicing`, and the touched-crate tests all pass; `cargo build --all-features` succeeds.


crates/swarm/net/pushsync/src/error.rs
```diff
@@ -11,8 +11,20 @@ vertex_net_codec::protocol_error! {
         #[error("invalid chunk address: {0}")]
         InvalidAddress(String),
```
mfw78: Does this need to be a string?

crates/swarm/net/pushsync/src/error.rs
```diff
@@ -11,8 +11,20 @@ vertex_net_codec::protocol_error! {
...
 
-        /// Storage radius exceeds maximum (255).
-        #[error("invalid storage radius: {0} exceeds u8 range")]
+        /// Malformed postage stamp in the delivery.
+        #[error("invalid stamp: {0}")]
+        InvalidStamp(String),
```
mfw78: Does this need to be a string?

crates/swarm/net/pushsync/src/error.rs
```diff
@@ -11,8 +11,20 @@ vertex_net_codec::protocol_error! {
...
+
+        /// Malformed receipt signature.
+        #[error("invalid signature: {0}")]
+        InvalidSignature(String),
```
mfw78: Shouldn't this be #[from] alloy::Signature or something?

PR #181, opened 2026-06-08
feat(swarm-chequebook): on-chain chequebook backend behind `chain` feature

mfw78:
Adds the on-chain interaction layer for SWAP cheques: chequebook deploy, cheque cashout, and balance reads via an alloy provider over the `nectar_contracts` bindings.

## What changed

`crates/swarm/bandwidth/chequebook/` gains a `chain` module behind a new crate feature `chain` (default off). The module exposes:

- `ChequebookChain`: an async trait with `balance`, `liquid_balance`, `liquid_balance_for(beneficiary)`, `paid_out(beneficiary)`, `deploy_chequebook(issuer, ...)` via the factory `deploySimpleSwap`, and `cash_cheque` / `cash_cheque_beneficiary` taking a `SignedCheque`.
- `ProviderChequebookChain<P>`: an alloy-provider-backed implementation over the `nectar_contracts` `IChequebook` / `IChequebookFactory` bindings.
- `ChainConfig`: factory and BZZ token addresses, derivable from the `nectar_contracts` deployment constants via `ChainConfig::for_deployments`. The layer takes addresses from config and never hardcodes a single network; mainnet and testnet addresses both flow through the same path.

`alloy-provider` and `alloy-contract` are added as optional workspace dependencies, matching the existing alloy `2.0` line. They are gated behind the `chain` feature so the default and wasm dependency cones stay lean: `cargo tree` confirms neither crate is pulled without the feature.

The crate stays libp2p-free. Errors are a `thiserror` enum with `strum::IntoStaticStr` for clean metric `reason` labels.

## Testing

Unit tests cover the call and transaction encoding the bindings produce for `cashCheque`, `cashChequeBeneficiary`, and `deploySimpleSwap` (selector plus argument round-trip), and exercise read decoding (`balance`, `paid_out`) against an alloy mock provider returning canned ABI-encoded values.

A live on-chain cashout (sending a signed transaction to a node and confirming a receipt) is deliberately out of CI scope; this is called out in the test module comment.

Verified locally: `cargo test -p vertex-swarm-bandwidth-chequebook --features chain`, default-feature build (no alloy-provider pulled), and `--features chain` build all pass.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -145,6 +145,29 @@ dependencies = [
 ]
 
 [[package]]
+name = "alloy-contract"
+version = "2.0.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "3ae8b60d71b92824e095b4003ff01fd2bc923017b7568997c5f459240e83499c"
+dependencies = [
+ "alloy-consensus",
+ "alloy-dyn-abi",
+ "alloy-json-abi",
+ "alloy-network",
+ "alloy-network-primitives",
+ "alloy-primitives",
+ "alloy-provider",
+ "alloy-rpc-types-eth",
+ "alloy-sol-types",
+ "alloy-transport",
+ "futures",
+ "futures-util",
+ "serde_json",
+ "thiserror 2.0.18",
+ "tracing",
+]
+
+[[package]]
 name = "alloy-dyn-abi"
 version = "1.6.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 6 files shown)


mfw78:
Superseded by an incoming chain-service stack. Per review, the chain must not live inside the chequebook crate: it is a node-wide concern (swap/chequebook now; postage batch listener, redistribution/staking, price oracle later) and the default light node must build and run (upload/download) with zero chain/alloy-provider in its dependency cone, including wasm32.

Replacement design (after a 3-architect analysis of bee, critically, without copying):
- `vertex-chain-api`: wasm-safe traits + data only (ChainReader/TransactionSender/ChequebookChain/ChainConfig), no alloy-provider.
- `vertex-chain-service`: native-only alloy-provider impl (provider + transaction service + block-paged log listener + contract clients + a long-lived ChainService). The only crate that pulls alloy-provider/native-tls.
- chequebook reverts to a pure cheque codec; this PR's `chain.rs` becomes the chequebook contract client in `vertex-chain-service`, and its calldata tests are salvaged.
- The provider crate is an optional dependency of `bin/vertex` (under a `chain` feature) and the storer builder path only; library crates depend on the traits. A CI cone guard (`just check-cone`) proves the default + wasm cone is chain-free.

Recommend closing this in favour of the stacked PRs that extract the chain service and bolt chequebook on as a consumer.

PR #184, opened 2026-06-09
docs(wasm): chain code is wasm-compatible, drop the chain-absence cone guard

mfw78:
## What

Drops the `just check-cone` chain-absence guard introduced in the first revision of this PR and corrects `docs/agents/wasm.md`. After reviewer feedback, the cone guard's premise was wrong: it forbade `alloy-provider`, `alloy-contract`, and `vertex-chain-*` from the default `vertex` cone and the wasm cone, but that chain code is wasm-compatible. The deliverable is now the doc correction only.

## Why the guard was wrong

`alloy-provider`, `alloy-contract`, and the alloy signer/transport stack build for `wasm32-unknown-unknown` when their features are selected to avoid native-only deps (`native-tls`/`openssl`, a default `reqwest` TLS backend, threaded `rayon`). Forbidding chain code from the wasm and light-node cones created a structural bottleneck that pushed later PRs into hand-reimplementing pieces of alloy, which is wasted effort. The rule should be "the wasm-targeted crates compile for wasm with our chosen features", not "chain code is absent".

## Why not a wasm-buildability guard instead

A guard that builds the designated wasm crates for `wasm32-unknown-unknown` would be the right replacement, but it cannot pass cleanly today and would build zero crates: every designated wasm-cone crate (`vertex-swarm-primitives`, `-spec`, `-forks`, `-api`, `-identity`) transitively pulls `wasm-bindgen-rayon` through `nectar-primitives`, which fails to compile for plain `wasm32-unknown-unknown` without a threaded-wasm toolchain (`atomics` + `bulk-memory` + `build-std`). `vertex-swarm-spec`/`-api` additionally pull `vertex-tasks`, whose `tokio::select!` usage needs the wasm tokio feature set trimmed. These are upstream (nectar) and tracked (vertex-tasks) blockers, out of scope here. Shipping a guard that builds nothing would be a fake check, so this PR drops the recipe and lands only the honest doc correction. The wasm-buildability guard should land as a follow-up once the `wasm-bindgen-rayon` pull is fixed upstream in nectar and `vertex-tasks` is trimmed.

## Doc correction

`docs/agents/wasm.md`:

- New "Chain code in wasm" section: chain code is wasm-compatible and welcome in the wasm cone when features are chosen to avoid native-only deps. Do not architect a "no chain in wasm" boundary. Strict primitive crates should aim for `no_std` where sensible without over-engineering.
- Corrects the stale Status claim that the `no_std` crates "compile to wasm today". They do not yet, and the blockers are wasm-incompatible deps (`wasm-bindgen-rayon` via `nectar-primitives`, `vertex-tasks` tokio features), not chain code.

## CI wiring

No `.github/workflows/` change is included. The earlier revision's CI job could not be pushed because the available token lacks the `workflow` scope. There is no guard recipe to wire now anyway; the wasm-buildability follow-up will need a maintainer with `workflow` scope (or an SSH push) to add the CI job.


docs/agents/wasm.md
```diff
@@ -8,7 +8,7 @@ A wasm client unlocks light-client embeddings in dapps, indexer UIs, and the Nex
 
 ### Status
 
-- Already no_std capable with a `default = ["std"]` feature: `vertex-swarm-primitives`, `vertex-swarm-spec`, `vertex-swarm-forks`, `vertex-swarm-api`. These compile to wasm today.
+- Already no_std capable with a `default = ["std"]` feature: `vertex-swarm-primitives`, `vertex-swarm-spec`, `vertex-swarm-forks`, `vertex-swarm-api`. Their dependency cone resolved for `wasm32-unknown-unknown` with `--no-default-features` is free of chain code and of `reqwest`, and `just check-cone` enforces that. A plain `cargo build` for wasm does not yet succeed: `nectar-primitives` pulls `wasm-bindgen-rayon`, which needs a threaded-wasm toolchain (`atomics` + `bulk-memory` + `build-std`), and `vertex-swarm-spec`/`vertex-swarm-api` pull `vertex-tasks`, whose `tokio::select!` usage needs the wasm tokio feature set trimmed per the tokio audit below. Both are tracked work, not a regression.
```
mfw78: I don't think we have to be free of chain code for wasm. chain code in wasm is absolutely fine, but for things that are strict primitives, then these should be able to be no_std likely, but don't be too crazy on this. what has been designed so far has essentially created a structural bottleneck that has caused subsequent PRs to effectively duplicate / reimplement alloy, which is a waste of fucking tokens and time.

mfw78:
Pivoted per review. Dropped the `just check-cone` chain-absence guard and kept only the `docs/agents/wasm.md` correction.

I verified the wasm situation in `nix develop` (which provides the `wasm32-unknown-unknown` std): `cargo build --target wasm32-unknown-unknown --no-default-features -p vertex-swarm-primitives -p vertex-swarm-forks` fails because `nectar-primitives` unconditionally pulls `wasm-bindgen-rayon`, and every designated wasm-cone crate inherits that. So a real wasm-buildability guard would build zero crates right now. Rather than ship a fake check, I dropped the recipe; the wasm-buildability guard is a follow-up once `wasm-bindgen-rayon` is fixed upstream in nectar and `vertex-tasks` is trimmed.

Note: this branch was never merged with current `main` (it predates the chequebook/swap merge), so GitHub shows it slightly behind base; the net change against `main` is the small wasm.md edit. Happy to rebase onto `main` if preferred.

PR #185, opened 2026-06-09
refactor(chain-api): use alloy types, drop reinvented chain abstraction

mfw78:
The chain crate had grown a parallel copy of alloy: a `ChainReader` trait over reads and `eth_call`, a `TransactionSender` over send/confirm, `TxRequest` and `TxReceipt` data types, and a `LogFilter`. An alloy `Provider` with its fillers already does all of that, and does it in wasm. This replaces the reinvention with a thin layer over alloy that holds only what alloy does not cover for a Swarm node.

Consumers now take a shared `alloy_provider::Provider` (and `ChainConfig` for the addresses) and call it directly. A node with no chain configured holds no provider.

Deleted: `reader.rs`, `sender.rs`, the `tx.rs` duplicates (`TxReceipt`, `TxStatus`, `LogFilter`, the old `TxRequest`), and `disabled.rs` (`DisabledChain` only existed to implement the deleted traits).

The crate is now:

- `ChainConfig` keyed on `alloy_chains::NamedChain` rather than a bare `u64`, carrying the `nectar_contracts` addresses, with `mainnet()` (Gnosis) and `testnet()` (Sepolia) constructors.
- `ChainError` and `TxError`: typed `thiserror` enums that carry alloy's `TransportError` and `PendingTransactionError` through `#[from]` (no lossy string-only variants) with `strum::IntoStaticStr` `reason` labels.
- `ProviderExt`: an extension trait on `alloy_provider::Provider<Ethereum>` with a blanket impl for the two pending-transaction operations alloy has no built-in for, `resend` and `cancel`. Both rebuild via `TransactionRequest::from_recovered_transaction` so the replacement keeps the original sender and nonce, then bump the fee. `recover_pending` is dropped from the provider surface: a `Provider` holds no record of a previous run's broadcasts, so that is application-persisted state, and the owning component reconstructs the hashes and calls `resend`/`cancel` on each.
- `TxRequest`: a newtype over `alloy_rpc_types_eth::TransactionRequest` that attaches a `&'static str` description for logs and metrics. It derefs to the inner request.

The chequebook chain trait is removed from chain-api: the on-chain chequebook client is an implementation detail of SWAP settlement and belongs in the chequebook chain crate (#186), so chain-api no longer depends on `vertex-swarm-bandwidth-chequebook` for `SignedCheque`.

`vertex-swarm-bandwidth-chequebook` now takes the settlement chain as an `alloy_chains::NamedChain` rather than a `u64` for EIP-712 domain construction, so callers name the chain instead of restating a magic number.

`alloy-provider`, `alloy-network`, and `alloy-rpc-types-eth` are added to the workspace with `default-features = false` so the `Provider` trait comes in without reqwest or native TLS, keeping the crate wasm-compatible. `docs/design/chain-service.md` is rewritten: chain code is wasm-compatible with careful feature selection, the chain crate is a thin alloy-extension layer rather than an alloy wrapper, and the chequebook chain client lives in the chequebook crate.

After the cuts the crate is deliberately small; that is the intended outcome. The api/service crate split is being reconsidered in the downstream PRs (#186/#187); the crate keeps the `vertex-chain-api` name for now.

Verified locally: `cargo fmt --all --check`, `cargo clippy --all-targets --all-features -- -D warnings` (chain-api and chequebook), `cargo test -p vertex-chain-api -p vertex-swarm-bandwidth-chequebook`, and `cargo build` (full workspace) all pass.


crates/chain/api/src/chequebook.rs
```diff
@@ -0,0 +1,72 @@
...
+
+use crate::ChainError;
+
+/// On-chain chequebook operations for the SWAP settlement service.
+///
+/// Read methods take the chequebook contract address explicitly so a single
+/// implementation can serve queries against any peer's chequebook, not just the
+/// local one. Write methods return a [`TxHash`] for the caller to confirm
+/// through the [`crate::TransactionSender`].
+#[async_trait::async_trait]
```
mfw78: This is great, this line is precisely how we should be using the async_trait macro. Create an issue to flag whole-of-codebase review to ensure that this is the case.

crates/chain/api/src/chequebook.rs
```diff
@@ -0,0 +1,72 @@
...
+    /// caller and the rest to `recipient`. Uses `cashCheque` with both the
+    /// beneficiary signature (carried by the cheque) and the issuer/authorizing
+    /// signature `sig`.
+    async fn cash_cheque(
+        &self,
+        cheque: &SignedCheque,
+        recipient: Address,
+        payout: U256,
+        sig: Bytes,
+    ) -> Result<TxHash, ChainError>;
```
mfw78: Fundamentally, I don't think that the Chequebook should exist anywhere in a chain trait that is used. This is an IMPLEMENTATION detail of a protocol / function within Vertex, and isn't related to the generalised case of handling chain functionality. As such, this is a prime candidate to be moved to a dedicated chequebook crate, such as vertex-swarm-bandwidth-chequebook or similar.

crates/chain/api/src/config.rs
```diff
@@ -0,0 +1,88 @@
...
+//! left to the consumer so this crate stays free of the spec dependency.
+
+use alloy_primitives::Address;
+use nectar_contracts::{ChequebookFactory, StoragePriceOracle, Token};
+
+/// Contract addresses and chain identity for a chain service.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct ChainConfig {
+    /// EIP-155 chain id (100 for Gnosis mainnet, 11155111 for Sepolia testnet).
+    pub chain_id: u64,
```
mfw78: This shouldn't be a u64, we should lean on alloy primitives / alloy types for this. I'm pretty sure that there is an alloy-chain (or similar) crate that provides a type that handles a "Chain" like this.

crates/chain/api/src/error.rs
```diff
@@ -0,0 +1,105 @@
...
+    Reverted(String),
+
+    /// The chain service is disabled for this node configuration.
+    ///
+    /// Returned by chain-off code paths (for example a light node, a bootnode,
+    /// or a wasm client) so a consumer can distinguish "no chain configured"
+    /// from a genuine transport failure.
+    #[error("chain service is disabled")]
+    Disabled,
+}
```
mfw78: This in general looks stupid because it doesn't include #[from] kind of annotations and uses strings which is lossy and/or lacks specificity which we should use as this is in the api, so it should be more "library" like.

crates/chain/api/src/reader.rs
mfw78, line 1: To be honest, this looks like AI slop, when basically all of this can be done by just sharing an alloy provider. Alloy providers are able to be in WASM, so doing this is pretty stupid

crates/chain/api/src/sender.rs
```diff
@@ -0,0 +1,45 @@
...
+/// recovery of pending transactions across restarts. Services inject it as
+/// `Arc<dyn TransactionSender>` and never touch a signer or a nonce directly.
+#[async_trait::async_trait]
+#[auto_impl::auto_impl(&, Arc)]
+pub trait TransactionSender: Send + Sync {
+    /// Broadcast `request` and return its hash without waiting for inclusion.
+    async fn send(&self, request: TxRequest) -> Result<TxHash, TxError>;
+
+    /// Wait for `hash` to confirm and return its receipt summary.
+    async fn confirm(&self, hash: TxHash) -> Result<TxReceipt, TxError>;
```
mfw78: This is duplicative when compared to an alloy provider

crates/chain/api/src/sender.rs
```diff
@@ -0,0 +1,45 @@
...
+
+    /// Cancel a pending transaction by replacing it with a zero-value self-send
+    /// at the same nonce and a higher fee. Returns the cancellation's hash.
+    async fn cancel(&self, hash: TxHash) -> Result<TxHash, TxError>;
+
+    /// Re-attach to transactions left pending by a previous run.
+    ///
+    /// Called once at startup so the sender can resume monitoring (and, if
+    /// needed, re-price) transactions it broadcast before a restart.
+    async fn recover_pending(&self) -> Result<(), TxError>;
```
mfw78: Check to see if there are any idiomatic alloy based solutions that allow for doing this instead of reinventing the wheel potentially. In general though, resend, cancel, and recover_pending have reasonable functionality and if there are no alloy analogues, these could be kept, though likely it would be better to do an extension trait on an alloy provider.

crates/chain/api/src/tx.rs
```diff
@@ -0,0 +1,149 @@
...
+
+use alloy_primitives::{Address, B256, TxHash, U256};
+use bytes::Bytes;
+
+/// A request to submit a transaction to the chain.
+///
+/// The sender implementation owns nonce selection, gas estimation, and fee
+/// pricing. This type carries only the caller's intent and bounds.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct TxRequest {
```
mfw78, line 18: This looks like it is absolutely duplicative of an alloy type, doing this is pointless, just use an alloy tx request type, and use it as a newtype with potentially a description if we want to allocate a description that we determine for a tx request.

crates/chain/api/src/tx.rs
```diff
@@ -0,0 +1,149 @@
...
+            to: None,
+            data: init_code,
+            value: U256::ZERO,
+            gas_limit: None,
+            min_gas_limit: None,
+            tip_boost_percent: 0,
+            description,
+        }
+    }
+}
```
mfw78, line 34: This is straight out duplicating from alloy, it's incredibly wasteful!

crates/chain/api/src/tx.rs
mfw78, line 1: Most of this is an absolute duplication of primitives / types that are available within alloy (and that are wasm compatible)

crates/swarm/bandwidth/chequebook/tests/json_conformance.rs
```diff
@@ -15,7 +15,9 @@ use alloy_signer_local::PrivateKeySigner;
 use bytes::Bytes;
 use serde_json::Value;
 use vertex_swarm_bandwidth_chequebook::{ChequeExt, SignedCheque, cheque::Cheque};
-use vertex_swarm_spec::init_mainnet;
+
+/// Gnosis Chain mainnet EIP-155 id, the cheque signing chain.
+const MAINNET_CHAIN_ID: u64 = 100;
```
mfw78, line 22: This is stupid as it should be a NamedChain from alloy-chains

docs/design/chain-service.md
```diff
@@ -0,0 +1,59 @@
...
+
+This note records how that service is structured and why. It supersedes the earlier approach in #181, which embedded an alloy-provider-backed chequebook backend inside the chequebook crate behind a cargo feature. That put the provider in the wrong place: chain access leaked into a settlement primitive, and the cone purity of a default light node depended on a feature flag staying off rather than on a crate boundary.
+
+## Goal
+
+A light node is the default node type, and a client node must be able to run on `wasm32-unknown-unknown`. The default and wasm builds must contain zero chain transport in their dependency cone: no `alloy-provider`, no `alloy-contract`, no `reqwest`, no native-tls, no tokio net features. A storer, by contrast, requires the chain. The structure has to satisfy both from the same workspace without a consumer opting into a feature to stay lean.
```
mfw78: This is wrong. alloy-provider, alloy-contract etc are fine, so long as their features are carefully selected so that they do not become wasm incompatible

docs/design/chain-service.md
```diff
@@ -0,0 +1,59 @@
...
+
+`vertex-chain-service` (later PR, native-only) holds the `alloy-provider`-backed implementations of those traits: the reader over a real RPC transport, the transaction sender with nonce management and fee bumping, the chequebook backend over the `nectar_contracts` bindings, and the health probe. It is an optional dependency of `bin/vertex` and of the storer builder path, and of nothing else.
+
+## Cone purity by crate boundary, not by feature
+
+The provider lives in a separate crate that library code never names. Because nothing in the default or wasm cone has `vertex-chain-service` in its dependency graph, those builds cannot pull a provider in even by accident. There is no feature to forget to turn off. A library crate that needs chain access takes a `vertex-chain-api` trait object (`Arc<dyn ChainReader>`, `Arc<dyn TransactionSender>`, `Arc<dyn ChequebookChain>`) and never sees the implementation. The binary and the storer builder are the only places that select the concrete service and inject it.
+
+This is verified the usual way: `cargo build -p vertex-chain-api --target wasm32-unknown-unknown` builds, and `cargo tree` over the api crate shows no provider crates.
```
mfw78: This seems too restrictive in that we should absolutely be able to use alloy types but just make sure that we don't bleed features from alloy that affect wasm compatibility.

crates/swarm/bandwidth/chequebook/src/cheque.rs
mfw78, line 1: chain id being used like this is stupid as it could just simply be the NamedChain or similar from alloy-chains.

mfw78 (commented):
Extensively review the comments that have been left, and fix / abandon the stupidity of this 'cone' which is restricting the use of alloy providers. alloy providers / contract etc are fine so long as they don't use features that are incompatible / have no compatibility layer with wasm.

crates/chain/api/src/config.rs
```diff
@@ -31,33 +31,33 @@ pub struct ChainConfig {
 }
```
mfw78, line 34: OK, this is legitimately perhaps somewhat nice, but isn't this exact or similar (more extensive functionality) already present in nectar-swarms?

mfw78: Good catch. ChainConfig now derives the settlement chain from nectar-swarms via NamedSwarm::chain rather than hardcoding it; the new for_swarm(NamedSwarm)/from_swarm(Swarm) constructors pair that chain with the matching nectar-contracts deployment. Only the contract addresses remain sourced here, since nectar-swarms does not carry deployment constants.

PR #186, opened 2026-06-09
feat(chain-service): native alloy-backed chain implementation

mfw78:
PR 2 of the chain-service extraction, stacked on #185 (feat/chain-api). GitHub will retarget the base to main once #185 merges.

This adds vertex-chain-service, the native-only implementation of the wasm-safe vertex-chain-api trait surface introduced in #185. It supersedes the chain.rs from #181 (whose calldata encode/decode tests are salvaged here).

It is the only crate that pulls an Ethereum RPC stack (alloy-provider with reqwest + native-tls; native-tls because the rustls root-certs license is outside the allow-list). Nothing is wired into the node, builder, or binary yet (that is PR 3), so the default vertex binary cone and the wasm client cone stay chain-free.

## What is in the crate

- AlloyChainProvider over any alloy Provider implements ChainReader and ChainHealth. Reads map onto eth_chainId, eth_blockNumber, eth_getBlockByNumber, eth_getBalance, eth_call, and eth_getLogs. is_synced compares the head block timestamp to wall-clock. connect_validated reads the chain id at startup and errors if it does not match the expected one.
- TransactionService is the node-wide TransactionSender. Nonce is the max of the node's pending count and the highest persisted in-flight nonce, serialized behind a per-sender mutex. Gas and fee pricing are delegated to alloy fillers; a replacement sets an explicitly bumped fee before submission so the node accepts it. Confirmation is nonce-indexed: a pending transaction resolves as superseded the moment a higher nonce confirms for the same sender. Pending transactions persist through a PendingTxStore over the vertex-storage Database trait and are re-watched on startup; the persisted record carries the original call so a resend rebuilds it rather than turning into a cancel.
- ChequebookContract implements ChequebookChain using nectar_contracts SolCall calldata, reads via ChainReader::call and writes via TransactionSender::send.
- LogListener is a generic block-paged eth_getLogs engine with a persisted cursor, page size, tail, stall timeout, and backoff. Contract-agnostic, the base for future postage and staking listeners.
- ChainHandle bundles the read and send surfaces with the address book; ChainService is the SpawnableTask that recovers pending transactions on start then runs a health and metrics poll loop emitting chain.synced, chain.block_number, chain.tx_sent, chain.tx_confirmed, and chain.tx_reverted.

## Testing

No live chain. Reads, nonce computation, and nonce-indexed confirm/supersede are exercised against alloy's mocked transport; the send path uses a wallet-filled mocked provider; the pending store and log-listener paging use in-memory backends. A real on-chain cashout is out of CI scope, noted in the test-module comments.

cargo fmt, clippy (all targets, all features, -D warnings), cargo test -p vertex-chain-service (19 tests), and a full cargo build are green. The chain markers (vertex-chain-service, alloy-provider, native-tls, openssl-sys) are absent from the default vertex binary cone and the wasm-safe cone.


crates/swarm/bandwidth/chequebook/src/lib.rs
```diff
@@ -42,8 +48,12 @@
 //! standard base64. This whole JSON path is slated for protobuf replacement,
 //! tracked in issue #183.
 
+#[cfg(feature = "chain")]
+pub mod chain;
 pub mod cheque;
 
+#[cfg(feature = "chain")]
+pub use chain::ChequebookContract;
```
mfw78: I think we can most probably _not_ have this cfg feature, and instead, we should have a feature of "swap-chequebook" which can be cfg gated. So, when compiling "light clients", we can effectively "disable" defaults, or have it by default that swap-chequebook or whatever we call it is "off", and that it has to be specifically enabled. Then, as an example, if we use the feature "storer", it would compile and include swap-chequebook in the defaults for the cfg gating.

crates/swarm/bandwidth/chequebook/src/chain.rs
```diff
@@ -0,0 +1,302 @@
...
+                cumulativePayout: cheque.cheque.cumulativePayout,
+                beneficiarySig: Bytes::copy_from_slice(cheque.signature.as_ref()),
+                callerPayout: payout,
+                issuerSig: Bytes::copy_from_slice(issuer_sig.as_ref()),
+            },
+            "cash_cheque",
+        )
+        .await
+    }
+}
```
mfw78, line 159: The use of 'read_call' and 'send_call' really looks like it is doing some heavy lifting that is likely already able to be done by sol! or importing bytecode that has already been compiled. Check in the nectar-contracts to see if the bytecode is around and if we can use this to then interact with the subsequent contracts. It would be preferably instead of reinventing / duplicating.

crates/swarm/bandwidth/chequebook/src/lib.rs
mfw78: Done in 37118ad: renamed the chequebook `chain` feature to `swap-chequebook` (off by default), so light clients compile without an Ethereum provider. The consuming crates enable it through a `storer` role feature (wired in the stacked PR).

crates/swarm/bandwidth/chequebook/src/chain.rs
mfw78: Done in 37118ad: `read_call`/`send_call` are gone; reads and writes now go through `alloy_contract::CallBuilder` over the `nectar-contracts` `sol!` interfaces (`.call()` / `.send()`). `nectar-contracts` ships only interfaces (no bytecode), so chequebook deployment stays a `deploySimpleSwap` factory call, now driven through the contract builder.

PR #190, opened 2026-06-09
docs(swarm-builder): add client upload/download examples

mfw78:
Add two runnable examples under `crates/swarm/builder/examples/` that show how to build a Swarm Client with the node builder and exercise the chunk surface, as part of making the Client node type first-class.

`upload_chunk` builds a client through `DefaultClientBuilder`, constructs a content chunk, signs a postage stamp for it, pairs the two as a `StampedChunk`, and pushes it via `SwarmChunkSender::send_chunk`, printing the resulting `PushReceipt` (storer, signature, nonce, storage radius).

`download_chunk` builds a client with content verification on, retrieves a chunk by address through the verifying chunk provider, and prints the served-by overlay and chunk address.

Both examples build offline (`cargo build --examples`) and clearly comment that the actual upload/download needs connected peers; they report the offline branch gracefully rather than erroring, and a real embedder would supply persistent storage and a populated routing table.

The examples follow the same construction path as the binary: `ClientConfig::new(...)` then `DefaultClientBuilder::from_config(...).build(ctx)`, with a minimal `InfrastructureContext` an embedder would implement.

One small library change: `ClientRpcProviders::chunks()` is now public so an embedder (FFI, gRPC, or an example) can reach the chunk sender/provider without going through the gRPC surface. The field was previously private with no accessor.

Verification: `cargo fmt --all`, `cargo clippy -p vertex-swarm-builder --lib --all-features -- -D warnings`, `cargo clippy -p vertex-swarm-builder --examples --all-targets -- -D warnings`, `cargo build --examples`, and `cargo doc -p vertex-swarm-builder --all-features --no-deps` all pass.

### DIFF
--- a/crates/swarm/builder/examples/download_chunk.rs
+++ b/crates/swarm/builder/examples/download_chunk.rs
@@ -1,0 +1,91 @@
+//! Build a Swarm client, wait until it can route, and download a chunk.
+//!
+//! The flow mirrors what an FFI or gRPC embedder does: build a client through
+//! `DefaultClientBuilder`, spawn its event loop, dial bootnodes, await a
+//! deterministic readiness gate, then retrieve a chunk by address.
+//!
+//! Readiness is `TopologyHandle::wait_until_routable`, which resolves the
+//! moment a storer is connected. That is the state from which a retrieval has a
+//! peer to ask; it is event-driven, never a timed guess. A fuller readiness
+//! surface (target depth, neighborhood saturation) is tracked in
+//! <https://github.com/nxm-rs/vertex/issues/194>.
+//!
+//! Run with: `cargo run -p vertex-swarm-builder --example download_chunk`
+
+use std::path::{Path, PathBuf};
+use std::sync::Arc;
+
+use vertex_node_api::InfrastructureContext;
+use vertex_swarm_api::{
+    ChunkAddress, HasTopology, SwarmChunkProvider, SwarmNodeType, SwarmTopologyCommands,
+};
+use vertex_swarm_builder::{ChunkVerifyConfig, ClientConfig, DefaultClientBuilder};
+use vertex_swarm_identity::Identity;
+use vertex_swarm_node::args::{ChainConfig, NetworkConfig};
+use vertex_swarm_spec::init_dev;
+use vertex_tasks::TaskExecutor;
+
+/// Minimal [`InfrastructureContext`]: a task executor for the client's services
+/// and a data directory for its database.
+struct ExampleContext {
+    executor: TaskExecutor,
+    data_dir: PathBuf,
+}
+
+impl InfrastructureContext for ExampleContext {
+    fn executor(&self) -> &TaskExecutor {
+        &self.executor
+    }
+
+    fn data_dir(&self) -> &Path {
+        &self.data_dir
+    }
+}
+
+#[tokio::main]
+async fn main() -> Result<(), Box<dyn std::error::Error>> {
+    let executor = TaskExecutor::current();
+    let spec = init_dev();
+    let identity = Arc::new(Identity::random(spec.clone(), SwarmNodeType::Client));
+
+    let verify = ChunkVerifyConfig {
+        verify_content: true,
+        verify_stamp: false,
+    };
+    let config = ClientConfig::new(
+        spec,
+        identity,
+        NetworkConfig::default(),
+        Default::default(),
+        verify,
+        ChainConfig::default(),
+    );
+    let ctx = ExampleContext {
+        executor: executor.clone(),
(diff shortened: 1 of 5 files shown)


crates/swarm/builder/examples/download_chunk.rs
```diff
@@ -0,0 +1,105 @@
...
+            println!("  chunk address: {}", result.chunk.address());
+        }
+        Err(err) => {
+            // Offline (no peers) or a genuine retrieval failure both land here.
+            println!("Retrieval did not complete: {err}");
+            println!("Join a network with a peer holding this chunk to fetch it.");
+        }
+    }
+
+    Ok(())
```
mfw78: How would this know that it can go ahead and retrieve straight away? Kademlia would still be stabilising. There needs to be some way added in to create some "predictable"/deterministic (non-heuristic based) method to determine if we are at a suitable kademlia population to start retrieval.

crates/swarm/builder/examples/upload_chunk.rs
```diff
@@ -0,0 +1,131 @@
...
+        }
+        Err(SwarmError::NoStorer { chunk_address }) => {
+            println!(
+                "No storer peer connected for {chunk_address}; join a network to complete the upload"
+            );
+        }
+        Err(err) => {
+            eprintln!("Upload failed: {err}");
+        }
+    }
```
mfw78: Same deal, how does it know that it is stable in order to be able to push in chunks?

mfw78 (commented):
In general, these are OK, but they neglect that there is some warmup time required, and that we should establish what is a deterministic method to establish whether the node is warm or not (and make it NOT heuristically based).

Beyond this, 130+ lines of code for a simple, contrived example of this is poor user experience, and an issue should be filed to follow-up with making this UX/DX more idiomatic / rust friendly with sound defaults, so that these tools in effect have more utility / are easier to adopt.

crates/swarm/builder/examples/download_chunk.rs
mfw78: Addressed. The example now builds the client, spawns the event loop, dials bootnodes, then awaits a deterministic readiness gate before retrieving.

The gate is `TopologyHandle::wait_until_routable`, added in this PR (fb95177). It resolves the moment the node is connected to at least one storer, which is the deterministic state from which a retrieval has a peer to ask: `retrieve_chunk` picks the closest storers from the routing table, so with none connected there is nowhere to route. It is event/state-driven, not timed: it returns immediately if a storer is already connected and otherwise waits for the `PeerReady` event that brings one in (the subscription is taken before the state read so a storer connecting in the gap cannot be missed; a lagged receiver re-reads state). No sleep, no "wait N seconds".

This is the minimal honest condition for routing. A fuller readiness surface (target depth, neighborhood saturation, a node-type-aware "warm" signal driven by `DepthChanged`) is tracked in #194 and referenced from the example header.

crates/swarm/builder/examples/upload_chunk.rs
mfw78: Addressed, same mechanism as the download example. The upload now builds the client, spawns the event loop, dials bootnodes, then awaits `TopologyHandle::wait_until_routable` before pushing.

`send_chunk` routes to the closest storers and fails with `NoStorer` when none are connected, so the gate resolving on "at least one storer connected" is exactly the deterministic precondition for a push to have a target. It is event-driven off `PeerReady`, never time-based. Fuller readiness surface tracked in #194.

mfw78:
Addressed the review.

Deterministic readiness: added `TopologyHandle::wait_until_routable` (fb95177), an awaitable, event/state-driven gate that resolves once the node is connected to at least one storer. That is the precise, non-heuristic precondition for routing: both `send_chunk` and `retrieve_chunk` pick the closest storers from the routing table, so with none connected a push returns `NoStorer` and a retrieval has nowhere to ask. The gate returns immediately if a storer is already connected and otherwise waits for the `PeerReady` event; the subscription is taken before the state read so a storer connecting in the gap cannot be missed, and a lagged receiver re-reads authoritative state. No `sleep`, no warmup guess. The primitive lives on the topology handle so FFI and gRPC embedders get it too, not just the examples.

Both examples now follow the real embedder flow: build client, spawn the event loop, dial bootnodes, await `wait_until_routable`, then act. Running against a live network is still required to actually move a chunk, which is faithful.

Length: trimmed the boilerplate. The remaining lines are the doc header, the minimal `InfrastructureContext`, and (upload only) the stamp-signing helper. Each example is now focused on build -> await readiness -> act -> print the typed result.

A fuller readiness surface (target depth, neighborhood saturation, node-type-aware warmth via `DepthChanged`) is a larger change than this PR should carry; tracked in #194 and referenced from both example headers.

PR #191, opened 2026-06-09
feat(swarm-node): wire swap protocol into client behaviour

mfw78:
Wire the SWAP cheque-based settlement wire protocol into the client behaviour, behind an optional `swap` cargo feature, mirroring the existing pseudosettle plumbing.

## What this does

Adds a `swap = ["dep:vertex-swarm-net-swap"]` feature to `vertex-swarm-node` that pulls `vertex-swarm-net-swap` as an optional dependency and composes the `/swarm/swap/1.0.0/swap` protocol into the client handler and combined upgrade.

The handler advertises swap in the active full protocol set (clients and storers, not bootnodes), emits a cheque on `ClientCommand::SendCheque`, and surfaces a received cheque as `SwapEvent::ChequeReceived`. Exchange-rate negotiation rides the headers exchange already implemented in the wire crate; the node carries the rate value onto the wire and reports the peer rate back, leaving rate policy to the settlement service.

A dedicated event channel set via `ClientBehaviour::set_swap_events` routes swap events to a settlement service with strong types (`SignedCheque`, typed peer, typed `U256` rate), never raw wire bytes. Events are still emitted as `ClientEvent::SwapChequeReceived` / `SwapChequeSent` for other consumers, matching how pseudosettle surfaces both paths.

The `SwapEvent` shape (`ChequeSent { peer, peer_rate }`, `ChequeReceived { peer, cheque, peer_rate }`) matches the contract the future `vertex-swarm-bandwidth-swap` settlement service consumes.

## Scope

Touches only the node crate. No settlement service, no builder wiring, no accounting; those land in separate units. All swap types and wiring live behind `#[cfg(feature = "swap")]`.

## Verification

`cargo fmt --all`, `cargo clippy -p vertex-swarm-node --all-targets --all-features -- -D warnings`, and `cargo test -p vertex-swarm-node` (default and `--features swap`) all pass. The node crate builds in both feature configurations.

### DIFF
--- a/crates/swarm/node/src/protocol/behaviour.rs
+++ b/crates/swarm/node/src/protocol/behaviour.rs
@@ -107,15 +107,15 @@ impl ClientBehaviour {
         self.pseudosettle_event_tx = Some(tx);
     }
 
-    /// Set the sender for swap events.
+    /// Route swap events to the given sink.
     ///
-    /// When set, swap-related events will be sent to this channel in addition
+    /// Once routed, swap-related events will be sent to this channel in addition
     /// to being emitted as [`ClientEvent`]. The node wires this up when a swap
     /// settlement service is present.
     #[cfg(feature = "swap")]
     // Wired by the node builder when the swap settlement service is present.
     #[allow(dead_code)]
-    pub(crate) fn set_swap_events(&mut self, tx: mpsc::UnboundedSender<SwapEvent>) {
+    pub(crate) fn route_swap_events(&mut self, tx: mpsc::UnboundedSender<SwapEvent>) {
         self.swap_event_tx = Some(tx);
     }
 
--- a/crates/swarm/node/src/protocol/behaviour.rs
+++ b/crates/swarm/node/src/protocol/behaviour.rs
@@ -21,6 +21,8 @@ use tokio::sync::mpsc;
 use tracing::{debug, warn};
 use vertex_swarm_primitives::OverlayAddress;
 
+#[cfg(feature = "swap")]
+use super::SwapEvent;
 use super::{
     ClientCommand, ClientEvent, PseudosettleEvent,
     handler::{ClientHandler, Config as HandlerConfig, HandlerCommand, HandlerEvent},
@@ -78,6 +80,9 @@ pub(crate) struct ClientBehaviour {
     pending_events: VecDeque<ToSwarm<ClientEvent, HandlerCommand>>,
     /// Optional sender for pseudosettle events.
     pseudosettle_event_tx: Option<mpsc::UnboundedSender<PseudosettleEvent>>,
+    /// Optional sender for swap events.
+    #[cfg(feature = "swap")]
+    swap_event_tx: Option<mpsc::UnboundedSender<SwapEvent>>,
 }
 
 impl ClientBehaviour {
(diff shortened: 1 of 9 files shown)


mfw78 (commented):
OK, so this merges in a bunch of things, and can send / receive cheques, but is this actually plumbed in at all into the accounting side of things, with then sound logic around when one should be sending a cheque, and also sound logic about when one should expect a cheque / disconnect a peer that isn't paying? ie. there should be thorough accounting as well, and it should merge seemlessly with pseudosettle.

mfw78:
Good question, and it is a deliberate layering split. This unit (S5) is the swap WIRE plumbing only, exactly as pseudosettle is layered: the node crate carries the wire events and command, and the accounting policy lives in the settlement provider plus the builder accounting wiring.

What S5 delivers: it composes the `/swarm/swap` protocol into the client handler and upgrade, advertises `/swarm/swap/1.0.0/swap` in the full protocol set, emits a cheque on `ClientCommand::SendCheque`, and surfaces inbound/outbound results as strongly typed `SwapEvent::ChequeReceived`/`ChequeSent`. A dedicated `set_swap_events` channel routes these to the settlement service with strong types only (typed peer, the full `SignedCheque`, and the peer's exchange rate from the headers exchange), never raw wire bytes. This crate carries no thresholds and makes no settlement decisions.

The accounting policy you are asking about is delivered downstream, not here. When to send a cheque (debt crossing the payment threshold, including the early-payment trigger), when to expect a cheque, and disconnecting a peer that exceeds the disconnect threshold without paying are decided by the swap `SwarmSettlementProvider` in `crates/swarm/bandwidth/swap` (S4), driven by the per-peer balance state from `crates/swarm/bandwidth/core` Accounting and the thresholds on `SwarmAccountingConfig`. The provider is connected to this wire layer by the builder accounting wiring (S6), the same way pseudosettle's policy in `crates/swarm/bandwidth/pseudosettle` is connected to its wire events.

I verified the surface is sufficient for that policy: `create_swap_actor` and `SwapService::handle_event` in S4 already consume `SwapEvent` and hold a `ClientCommand` sender for `SendCheque`, and `SignedCheque` carries chequebook, beneficiary, cumulative payout, and signature (with `recover_signer`), so the provider can validate, compute the incremental amount, credit the balance, and cash on-chain. No surface change was needed; I only added a module-level rustdoc note making the wire-vs-policy boundary explicit.

Kept S5 scoped to wire plumbing so it stays mergeable; the policy and on-chain validation land in S4 and S6 as separate PRs.

crates/swarm/node/src/protocol/behaviour.rs
```diff
@@ -100,6 +107,18 @@ impl ClientBehaviour {
...
 
+    /// Set the sender for swap events.
+    ///
+    /// When set, swap-related events will be sent to this channel in addition
+    /// to being emitted as [`ClientEvent`]. The node wires this up when a swap
+    /// settlement service is present.
+    #[cfg(feature = "swap")]
+    // Wired by the node builder when the swap settlement service is present.
+    #[allow(dead_code)]
+    pub(crate) fn set_swap_events(&mut self, tx: mpsc::UnboundedSender<SwapEvent>) {
```
mfw78: I really don't think we should be prefixing functions with `set_` as this isn't idiomatic rust. Fix this one here, and then file an issue to do a codebase wide analysis of this.

mfw78: Renamed to `route_swap_events` so it reads as routing events to a sink rather than a field setter. Filed #195 for the codebase-wide `set_` audit covering the remaining wiring and config methods (including `set_pseudosettle_events`).

PR #193, opened 2026-06-09
feat(ffi): add embeddable client FFI surface

mfw78:
Add the `vertex-ffi` crate, the primary public API for embedding a Vertex Swarm client into a native host (Dart and Flutter, Swift, Kotlin and JNI, C++).

`VertexClient` owns a native multi-thread runtime and a running client node built through the highest-level builder entry point (`DefaultClientBuilder::from_config(...).build(ctx)`). The host builds one from a flat config, then uploads pre-stamped chunks and downloads chunks by address. Uploads reconstruct a `StampedChunk` from address plus wire bytes plus stamp (the address disambiguates the chunk variant), and downloads return the chunk's wire bytes, its stamp, and the serving peer.

Inputs and outputs cross the boundary as byte vectors, strings, and primitives only. The strong domain types (`StampedChunk`, `ChunkAddress`, `Stamp`, `PushReceipt`) are reconstructed immediately inside Rust and never leak across the boundary, mirroring the gRPC chunk service.

The C ABI and the per-language bindings are generated from the `api` module by flutter_rust_bridge via `flutter_rust_bridge.yaml`. There is no hand-maintained parallel C header. A placeholder `frb_generated` module keeps `cargo build -p vertex-ffi` self-contained, so a checkout builds and tests on its own and CI never runs the codegen binary; only a host that actually invokes the bindings regenerates the glue. A small `build.rs` registers the `frb_expand` cfg the `#[frb]` macro emits so the workspace `unexpected_cfgs` lint stays clean.

The crate is native-only: the runtime-bearing dependency is gated to non-wasm targets, since the browser embedding path is wasm-bindgen, a separate surface. No `serde_json`, `reqwest`, `axum`, or `hyper` anywhere in the cone.

The errors are a flat `thiserror` enum with `strum::IntoStaticStr` carrying pre-formatted strings, so a host never needs a vertex-internal error type.

One supporting change in the builder: a `chunks()` accessor on `ClientRpcProviders` so the embedding host can drive the client's chunk provider directly without standing up the gRPC server.

Documented limitation: one client per process. The node internals resolve their task executor from a process-global slot the first built client populates, so a second concurrent client would spawn its node tasks onto the first client's runtime. This matches the single-node assumption of the CLI; lifting it is a node-infrastructure change outside this crate.

Verification: `cargo fmt --all --check`, `cargo clippy -p vertex-ffi --all-targets -- -D warnings` (also with `--features chain`), `cargo build` workspace-wide, and `cargo test -p vertex-ffi` all pass.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -103,6 +103,7 @@ Per-area `AGENTS.md` files apply when you edit code in that directory.
 | `crates/storage/AGENTS.md` | Storage abstraction and redb backend. |
 | `crates/node/AGENTS.md` | Protocol-agnostic node infrastructure. |
 | `crates/observability/AGENTS.md` | Logging, tracing, metrics infra. |
+| `crates/ffi/AGENTS.md` | Native FFI surface for embedding a client. |
 
 ## Doc map
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -84,6 +84,17 @@ dependencies = [
 ]
 
 [[package]]
+name = "allo-isolate"
+version = "0.1.27"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "449e356a4864c017286dbbec0e12767ea07efba29e3b7d984194c2a7ff3c4550"
+dependencies = [
+ "anyhow",
+ "atomic 0.5.3",
+ "backtrace",
+]
+
+[[package]]
 name = "allocator-api2"
 version = "0.2.21"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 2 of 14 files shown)


mfw78:
We should research the frb and check what idiomatic methods we can use to create these types that can be used at the boundary / have our existing types used at the boundary. It would be far nicer to not have to reimplement a bunch of them at the FFI boundary. Also, we should make sure that we can in effect actualise on streaming, and reduce memory allocation, such as the case of uploading / downloading large files (many chunks), and clear these out non-sequentially.

Issue #229, opened 2026-06-11
topology: post-#225 cleanups (iterator supply API, trim tie-break, connection_limits, stale dialing doc)


mfw78:
## Summary

Small, independent cleanups identified during the #225 reviews. None changes behaviour on the network; each is a single focused PR.

## Items

1. **Honest iterator API for dialable supply.** `dialable_overlays_in_bin_excluding` takes a count and internally over-fetches to compensate for exclusions. Replace the count-based contract with an iterator (or `impl Iterator`) so callers take what they need and the over-fetch heuristic disappears along with the `saturating_mul(2)` it required.
2. **Prefix-diversity tie-break in trim selection.** When a bin holds a surplus, eviction currently ranks by score alone. Among equal scores, prefer evicting peers that share a long address prefix with a retained peer, so the kept set stays spread across the bin's sub-tries. This mirrors the balanced-selection goal on the way out, not just on the way in.
3. **Compose libp2p `connection_limits` behaviour.** Global connection caps are currently implicit in topology accounting. Compose `libp2p::connection_limits` into the swarm so transport-level limits (total, per-peer, pending) are enforced independently of kademlia bin logic, and topology only reasons about per-bin capacity.
4. **Rewrite `docs/networking/peer-dialing-strategy.md`.** The document predates the depth-deadlock fix and describes the old taper and trim behaviour. Rewrite once the connection-profiles work lands, or sooner if it misleads.

## References

Review follow-up from #225.

mfw78:
All four cleanups merged:

- Honest iterator API for dialable supply (item 1): #232 (refactor(swarm-peers): expose dialable supply as an iterator).
- Prefix-diversity tie-break in trim selection (item 2): #235 (feat(swarm-topology): prefer prefix-diverse retention in trim selection).
- Compose libp2p `connection_limits` (item 3): #233 (feat(swarm-node): enforce transport connection limits via libp2p connection_limits).
- Rewrite `docs/networking/peer-dialing-strategy.md` (item 4): #231 (docs(networking): rewrite peer dialing strategy for saturation-floored allocation).

Closing as complete.

Issue #228, opened 2026-06-11
topology: connection profiles, topology phases, dial-rate shaping, and pullsync readiness gate


mfw78:
## Summary

Umbrella for the connection-strategy direction agreed in the #225 reviews: a node should establish a solid, true neighbourhood depth as fast as reasonably practicable, with the aggressiveness chosen by node role. A client wants to be usable as soon as possible; a storer can take slightly longer but must settle its neighbourhood quickly so pullsync (#77) can start.

## Design sketch

**ConnectionProfile.** A named bundle of topology pacing parameters (dial concurrency, tick interval, per-tick dial budget, bootstrap fill level) selected by `SwarmNodeType` with a CLI override: `aggressive` as the client default, `balanced` as the storer default, `conservative` for constrained environments. Profiles only set numbers; they do not change logic.

**TopologyPhase.** An explicit phase machine replacing the implicit "bootstrap mode" predicate:

- `Bootstrap`: known table is thin; fill every bin towards the bootstrap level, dial at the profile's maximum rate.
- `Converging`: depth is climbing; prioritise dials that raise the unsaturated frontier.
- `Stable`: depth has been steady; maintenance pacing, inbound mostly suffices.

Transitions are observable (log + metric) so operators can see where startup time goes.

**Dial-rate shaping.** Replace fixed per-tick dial quotas with a GCRA token bucket per profile, so bursts after gossip influx are absorbed without hammering the network and without waiting a full tick when supply arrives.

**Pullsync readiness gate.** Storage radius is a distinct value from connection depth. Pullsync should begin when the neighbourhood (bins at and above depth) is saturated and stable, not when total connectivity looks good. Define the readiness predicate now so #77 can consume it.

## Existing issues this folds in or touches

- #64 (per-protocol stream timeouts configurable): the profile is the natural home for the topology-side knobs; protocol stream timeouts stay separate but should follow the same profile-selection pattern.
- #65 (hive timeouts configurable): same.
- #76 (client service protocol event handlers): the client profile defines when the client surface reports itself ready.
- #77 (pullsync): consumes the readiness gate above.

## Scope

- `crates/swarm/topology` (phase machine, pacing, profile plumbing).
- `crates/node` / `bin/vertex` (CLI flag placement per `crates/node/AGENTS.md`).
- `docs/networking/peer-dialing-strategy.md` is stale after #225 and should be rewritten as part of this work.

## References

Review follow-up from #225.

mfw78:
[...]

Issue #227, opened 2026-06-11
topology: depth hysteresis and split bootstrap_target from oversaturation band


mfw78:
## Summary

Two structural follow-ups from the #225 reviews on how neighbourhood depth and per-bin capacity interact.

## 1. Depth hysteresis

Depth is recomputed instantaneously from the connected-peer bin distribution on every connect and disconnect. A single disconnection in a bin sitting exactly at saturation drops depth immediately, which retargets the taper, which can trigger trims and dials that oscillate. The fix in #225 removed the feedback loop that made this fatal, but the instantaneous recompute still allows depth to flap under normal churn.

Proposal: require the new depth to be stable for a short window (or for the saturation deficit to exceed one peer) before lowering depth. Raising depth can remain immediate, since over-connection is harmless and the trim floor protects the climb.

## 2. Split `bootstrap_target` into its two meanings

`bootstrap_target` currently serves two distinct roles:

- the per-bin fill level used while the table is in bootstrap mode (no meaningful depth yet), and
- the floor for the trim surplus and the inbound ceiling band at any depth, where it acts as an oversaturation bound.

These coincide today at 18 (matching the reference network's oversaturation level), but they are conceptually different knobs: one is "how aggressively to fill while blind", the other is "how many peers above target a bin may retain". Connection profiles (see the connection-modes issue) will want to vary the first without touching the second.

Proposal: introduce `oversaturation_peers` for the retention/inbound band and keep `bootstrap_target` purely as the bootstrap fill level, with a construction-time invariant `bootstrap_target <= oversaturation_peers` and both floored at saturation, following the make-invalid-states-unrepresentable pattern from #225.

## Scope

- `crates/swarm/topology/src/kademlia/{limits,routing}.rs`.
- Capture-replay regression tests as in the #225 fix.

## References

Review follow-up from #225.

mfw78:
Both items merged:

- Depth hysteresis (item 1): #236 (feat(swarm-topology): hysteresis on neighborhood depth lowering).
- Split `bootstrap_target` from the oversaturation band (item 2): #234 (feat(swarm-topology): split bootstrap_target from oversaturation band).

Follow-up hardening on the depth/readiness clocks is tracked separately in #242. Closing as complete.

Issue #226, opened 2026-06-11
topology: replace gossip pre-verification swarm with verify-on-first-dial


mfw78:
## Summary

Vertex currently verifies every gossiped peer record by dialling the peer on a separate ephemeral-identity swarm and running a full handshake before the peer is admitted to the known table. This design was flagged by all three reviews of the depth-deadlock fix (#225) as the largest remaining structural smell in the discovery path. It should be replaced by collapsing verification into the first real dial.

## Problems with the current design

1. **Double handshake.** Every peer that we eventually connect to pays two complete handshakes: one on the verification swarm and one on the real swarm. This doubles ECDSA work for both sides and doubles our connection visibility on the network.
2. **Cold-start delay.** The known table only fills as fast as the verification swarm can churn through the gossip backlog, which adds minutes of warm-up before candidate selection has real supply. The depth climb measured on mainnet only starts once verification throughput catches up with gossip intake.
3. **Re-verification churn against bee 2.8.0.** Bee 2.8.0 re-signs gossip records on every broadcast (fixed upstream in ethersphere/bee#5493), so the same peer arrives with a fresh signature repeatedly and is re-verified each time. Without a cooldown this wastes verification capacity on peers we already know.
4. **Ephemeral identity is a fingerprint.** A node that handshakes with a throwaway identity and immediately disconnects is trivially identifiable as a vertex prober and contributes nothing to the network.

## Proposed direction

Admit gossiped records into the known table as an unverified tier. Candidate selection may dial unverified peers; a successful real handshake both verifies and connects in one round trip. A failed dial demotes or expires the record exactly as it does today. Records that update an existing verified peer only need signature validation, not a dial. Add a per-peer cooldown so re-signed records from bee 2.8.0 peers do not trigger re-processing more than once per interval.

This removes the verification swarm entirely, removes the warm-up delay, halves handshake cost on the happy path, and makes our connection behaviour indistinguishable from a normal node.

## Scope

- `crates/swarm/topology` (gossip intake, candidate selection eligibility).
- `crates/swarm/peers/peer-manager` (known-table tiers, record update path).
- Docs: `docs/swarm/hive-gossip.md`, `docs/networking/peer-management.md`.

## References

Review follow-up from #225. Related upstream behaviour: ethersphere/bee#5493 (gossip re-signing churn).

PR #230, opened 2026-06-11
fix(topology): break the depth deadlock between trim, taper, and candidate supply

### LINKED ISSUE #225: topology: depth permanently stuck at 0 - trim evicts below saturation, taper targets below saturation, candidate supply polluted by connected peers
A 15-minute mainnet client run (2026-06-10, JSON log capture) shows kademlia depth climbing to 7 in 49 seconds and then collapsing to 0, where it stays frozen for the rest of the run with `connected=111` and zero further discovery dials. The depth maths itself is correct and matches the reference (saturation 8, low watermark 3, `nectar_primitives::recompute_neighborhood_depth`); the failure is a self-defeating interaction between the depth-change trim, the tapered connection allocation, and the dial-candidate supply.

## Observed timeline (all numbers from the capture)

**Phase 1, climb (22:54:44 to 22:55:33).** Once the dialable pool warms up, dialling bursts (82 then 64 discovery dials per minute), bins fill, and depth climbs 0 -> 1 -> 2 -> 3 -> 4 -> 5 -> 7.

**Phase 2, self-destruction (22:54:56 to 22:55:33).** Every `kademlia depth changed` event immediately fires `trim_overpopulated_bins`, which evicts active peers down to the new depth's taper targets (`DepthAwareLimits::surplus`, `kademlia/limits.rs`). At depth 7 the taper allocates `total_target * (bin+1) / weight_sum` = `160 * (bin+1) / 28`: bin 0 gets 5, bin 1 gets 11, bin 2 gets 17. The post-trim populations match exactly: bin0=5, bin1=11, bin2=17. Three eviction waves removed 58 active peers (17, 10, 31). The moment bin 0 dropped to 5, below `saturation_peers = 8`, depth recomputed 7 -> 0. The trim destroys the saturation that produced the depth.

**Phase 3, permanent freeze (22:55:42 to end of capture).** `evaluate_connections` allocates with `effective_depth = max(connected_depth, estimate_depth_from_known)` (`routing.rs`). The known table is large by now, so the estimate (~9) keeps the taper engaged even though connected depth is 0, and the taper targets equal the trimmed populations exactly, so `needs_more` is false in every bin. Zero discovery dials for the remaining 9.5 minutes. There is no fixed point with depth > 0 while any shallow bin's taper target sits below saturation.

## Reference divergence

The reference implementation prunes only the surplus above `overSaturationPeers = 18` per bin (`binPruneCount` in bee `pkg/topology/kademlia/kademlia.go` returns `size - oversaturationAmount`) and never prunes toward an allocation target, so bins can never drop below saturation and depth is monotone under pruning. Vertex trimmed to taper targets as low as 5 at depth 7 and 3 at depth 9+.

This is a second instance of the depth/allocation coupling documented on `DEFAULT_BOOTSTRAP_TARGET` (`kademlia/limits.rs`: "Must be >= SwarmSpec::saturation_peers() or depth can never advance past 0"), reached through the trim and taper paths the bootstrap fix does not cover.

## Third defect, found while verifying the fix live

`PeerManager::dialable_overlays_in_bin` did not exclude already-connected peers. Connected peers are hot, dialable, and most recently used, so once a bin held more connected peers than the requested candidate count, the returned slice consisted entirely of already-connected overlays, the `CandidateSelector` rejected every one (`connected_peers.exists`), and the bin starved forever despite 100+ known unconnected peers. Observed in a 12-minute verification run: every bin froze the moment `connected >= deficit` (bin 0 at 5 connected against a deficit of 3, with 128 known), `evaluated connection candidates` events stopped entirely, and depth never climbed.

## Fix

1. **Trim floor.** Trimming reclaims peers only above `max(target, bootstrap_target = 18)`, mirroring the reference network's prune-above-oversaturation behaviour. Pruning can no longer undo saturation, so depth is monotone under trim.
2. **Taper floor.** Balanced-bin taper targets are floored at `saturation_peers` (the previous `nominal = 3` floor is below saturation), threaded from `SwarmSpec` at behaviour construction so allocation and depth recomputation always agree.
3. **Candidate supply.** `dialable_overlays_in_bin_excluding` lets candidate selection exclude the connected set, so saturating bins keep refilling toward their floors.

Verified on mainnet: depth climbs 0 -> 9 monotonically in 88 seconds, survives seven trim events during the climb, and holds at 9 for the remainder of a 12-minute run with shallow bins pinned at the trim floor (`bins 0:18 1:18 2:18 3:17 4:18`, `connected=163`).

## Related follow-ups (not addressed here)

- The first ~4.5 minutes of every run produce almost no dials because the dialable pool is gated on gossip verification, which is serialized and dominated by re-verification churn (2.8.0 peers re-sign their gossip records on every broadcast, so known peers fail the signature-equality check and get re-dialled; fixed upstream for 2.9 in ethersphere/bee#5493).
- Trimming synchronously on every depth change amplified the oscillation (three eviction waves in 37 seconds mid-climb); the reference trims periodically. Hysteresis is worth considering separately.
- `select_neighborhood_candidates` iterates deepest-first, so during bootstrap the shallow bins that gate depth are last in line for dial slots.

mfw78:
Closes #225

## Problem

On mainnet, neighbourhood depth stayed at 0 indefinitely while the node held 150+ connections. Forensics on a 15-minute JSON capture showed three coupled bugs forming a deadlock:

1. **Trim cut below saturation.** The eviction surplus was computed against the linear taper target alone, so shallow bins were trimmed down to allocations (5, 11, 17 peers) below the saturation level (8) that depth recomputation requires. Every climb triggered a trim that immediately undid it.
2. **Taper floor below saturation.** The per-bin allocation `total_target * (bin + 1) / weight_sum` was floored only at `nominal` (3), so shallow-bin targets could never support a saturated bin in the first place.
3. **Candidate supply polluted by connected peers.** Dialable candidate selection drew from the known table without excluding already-connected overlays, so saturating bins returned only peers we were already connected to and starved their own refill.

## Fix

- Floor the taper allocation at saturation, and floor the bootstrap allocation at saturation so a `bootstrap_target` below saturation is unrepresentable (no runtime assert needed).
- Trim surplus is measured against `max(target, bootstrap_target)`, and only over the evictable population (handshaking and active); in-flight dials hold capacity but cannot be evicted, so charging their slots to active peers cut bins below saturation.
- Inbound admission gets an explicit ceiling band `(target + inbound_headroom).max(bootstrap_target)` so the dial path, inbound path, and trim floor describe one coherent band, mirroring the reference network's dial-to-saturation / accept-to-oversaturation behaviour.
- Depth recomputation reads saturation from the limits themselves instead of re-reading the spec, so there is a single source of truth; the node builder threads the spec's `saturation_peers` into the limits at construction.
- Both candidate selection paths exclude connected, in-progress, and queued overlays from the dialable supply, with the peer manager exposing `dialable_overlays_in_bin_excluding`. This also fixes an `attempt to add with overflow` panic in the over-fetch arithmetic when callers passed `usize::MAX`.

## Verification

- Capture-replay regression test reproducing the mainnet bin distribution (depth 7, bins 26/25/23/16/9/11/13/14/9/3) and asserting trim no longer collapses depth.
- Live mainnet run before review changes: depth climbed 0 to 9 in 88 seconds and held for 10+ minutes with bins pinned at the oversaturation band and 163 connections. A second live run after the review changes is recorded in the PR conversation.
- Three independent expert reviews (libp2p, Swarm protocol, Rust idiom) returned approve-with-changes; all pre-merge findings are applied here. Structural follow-ups are filed as #226, #227, #228, and #229.

AI Assistance: this PR was developed with the help of an AI coding agent (Claude Code); all changes were reviewed and verified by the author.

### DIFF
--- a/crates/swarm/api/src/lib.rs
+++ b/crates/swarm/api/src/lib.rs
@@ -85,8 +85,8 @@ pub use self::reporting::{
 };
 pub use self::rpc::RpcProviders;
 pub use self::spec::{
-    DEFAULT_NEIGHBORHOOD_LOW_WATERMARK, DEFAULT_SATURATION_PEERS, StaticSwarmSpecProvider,
-    SwarmSpec, SwarmSpecParser, SwarmSpecProvider, SwarmToken,
+    DEFAULT_SATURATION_PEERS, StaticSwarmSpecProvider, SwarmSpec, SwarmSpecParser,
+    SwarmSpecProvider, SwarmToken,
 };
 pub use self::swarm::{SwarmClient, SwarmStorer};
 pub use self::types::{
--- a/crates/swarm/api/src/lib.rs
+++ b/crates/swarm/api/src/lib.rs
@@ -85,7 +85,8 @@ pub use self::reporting::{
 };
 pub use self::rpc::RpcProviders;
 pub use self::spec::{
-    StaticSwarmSpecProvider, SwarmSpec, SwarmSpecParser, SwarmSpecProvider, SwarmToken,
+    DEFAULT_NEIGHBORHOOD_LOW_WATERMARK, DEFAULT_SATURATION_PEERS, StaticSwarmSpecProvider,
+    SwarmSpec, SwarmSpecParser, SwarmSpecProvider, SwarmToken,
 };
 pub use self::swarm::{SwarmClient, SwarmStorer};
 pub use self::types::{
--- a/crates/swarm/peers/peer-manager/src/manager.rs
+++ b/crates/swarm/peers/peer-manager/src/manager.rs
@@ -1107,8 +1107,7 @@ mod tests {
         let bin = Bin::from(test_overlay(0).proximity(&test_overlay(1)));
         let excluded = test_overlay(1);
 
-        let dialable =
-            pm.dialable_overlays_in_bin_excluding(bin, usize::MAX, |o| *o == excluded);
+        let dialable = pm.dialable_overlays_in_bin_excluding(bin, usize::MAX, |o| *o == excluded);
         assert!(!dialable.contains(&excluded));
     }
 
--- a/crates/swarm/peers/peer-manager/src/manager.rs
+++ b/crates/swarm/peers/peer-manager/src/manager.rs
@@ -276,8 +276,25 @@ impl<I: SwarmIdentity> PeerManager<I> {
 
     /// Get dialable overlay addresses from a specific bin (not banned, not in backoff).
     pub fn dialable_overlays_in_bin(&self, bin: Bin, count: usize) -> Vec<OverlayAddress> {
+        self.dialable_overlays_in_bin_excluding(bin, count, |_| false)
+    }
+
+    /// Get dialable overlay addresses from a bin, skipping excluded overlays.
+    ///
+    /// `exclude` lets the caller remove overlays that are dialable but useless
+    /// as candidates, typically peers it is already connected to. Without the
+    /// exclusion, connected peers (hot and dialable) fill the returned slice
+    /// first and a bin with more connections than `count` yields no usable
+    /// candidates even when many unconnected peers are known.
+    pub fn dialable_overlays_in_bin_excluding(
+        &self,
+        bin: Bin,
+        count: usize,
+        exclude: impl Fn(&OverlayAddress) -> bool,
+    ) -> Vec<OverlayAddress> {
         self.index.filter_bin(bin, count, |overlay| {
             !self.banned_set.contains_key(overlay)
+                && !exclude(overlay)
                 && self.peers.get(overlay).is_some_and(|e| e.is_dialable())
         })
     }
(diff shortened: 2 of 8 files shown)


mfw78:
Second live mainnet verification, run against the branch including the review-response commit (pre-rebase binary; the subsequent rebase onto main only adapted construction to the new `TopologyBehaviourBuilder` and peer-hub APIs, the topology logic is unchanged):

- Fresh datadir, mainnet, started 00:39 UTC.
- Warm-up (gossip verification gating the dialable pool, tracked separately in #226) lasted about 4 minutes, then depth climbed 0 to 7 in under 2 minutes with 152 connections.
- 10-minute stability window sampled every 30 s: depth min 8, max 8, final 8, with 199 cumulative peer-connected events. No collapse, no oscillation.

This reproduces the first verification (depth 0 to 9 in 88 s, stable 10+ minutes) and additionally exercises the new inbound ceiling band from the review changes.

PR #238, opened 2026-06-11
feat(swarm-topology): explicit topology phase machine

mfw78:
## Summary

Replaces the implicit "bootstrap mode" predicate (depth == 0 inside `DepthAwareLimits::target()`) with an explicit, observable `TopologyPhase` machine. The `depth == 0` allocation branch in the limits is untouched: the phase machine names and observes the lifecycle state, becoming the canonical answer to "what phase are we in" for consumers and operators, so the upcoming dial-shaping and pullsync-readiness work can hook into it.

## Phase machine

`TopologyPhase` (`Bootstrap`/`Converging`/`Stable`, exported from `vertex-swarm-topology`) lives in `kademlia/phase.rs` as a pure state machine (`PhaseTracker`): callers supply the observation time, so the transition table is unit-tested with simulated clocks. The phase is re-derived deterministically from observable routing state rather than edge-triggered:

- `Bootstrap`: published depth is 0. The known table is too thin to anchor a neighborhood; every bin fills toward the bootstrap target.
- `Converging`: depth is above 0 but moved within the stability window, or the neighborhood has not reached the saturation threshold (same condition as `ReadinessSnapshot::is_saturated`: connected peers in bins inside the depth boundary total at least the spec's saturation target).
- `Stable`: depth above 0, unmoved for the stability window, neighborhood saturated.

A `Stable` node falls back to `Converging` when depth moves or saturation is lost, and all the way to `Bootstrap` only when depth returns to 0. Depth churn that leaves the derived phase unchanged commits nothing, so there is no transition spam within the window.

The stability window is a `KademliaConfig` knob (`with_phase_stability_window`, default 60s, deliberately no CLI flag).

## Evaluation points

The routing layer re-derives the phase at the existing depth-publication points (handshake completion and connection close in the behaviour) and on the periodic connection-evaluator tick, which alone observes the time-driven `Converging` -> `Stable` settle on a quiet table. The evaluator task now receives the topology event channel so timer-driven transitions broadcast like every other.

## Observability

- One `info!` line per transition: `from`, `to`, `depth`, `time_in_phase`, recorded under the tracker lock so concurrent evaluations cannot interleave out of order.
- `topology_phase_changes_total{from,to}` counter (labels via `strum::IntoStaticStr`, snake_case), distinct from the existing per-connection `topology_phase_transitions_total`.
- One-hot `topology_phase{phase}` gauge published from startup (Bootstrap) and updated on every transition.
- `TopologyEvent::PhaseChanged { from, to, depth }` broadcast on the topology event stream; `TopologyHandle::wait_until` re-evaluates on it.
- `ReadinessSnapshot` now carries `phase` and `time_in_phase`, read from authoritative routing state.

## Tests

- Transition table in `phase.rs` under simulated clocks: bootstrap -> converging on first depth climb, converging -> stable after the window with saturation, converging holds without saturation, stable -> converging on depth change and on saturation loss, collapse to bootstrap at depth 0, and no transition spam under depth churn within the window.
- Routing-level lifecycle tests through real connect/disconnect sequences (zero window isolates the saturation condition).
- The three-node `vertex-swarm-node` convergence test asserts every node reports `Bootstrap` on the readiness surface, mirroring the depth-0 result (a three-node cluster cannot saturate a neighborhood, so deeper phases are out of reach there).

Part of #228

### DIFF
--- a/crates/swarm/node/src/node/client.rs
+++ b/crates/swarm/node/src/node/client.rs
@@ -310,6 +310,7 @@ impl<I: SwarmIdentity + Clone> ClientNode<I> {
             TopologyEvent::PeerDisconnected { .. } => {}
             TopologyEvent::PeerRejected { .. } => {}
             TopologyEvent::DepthChanged { .. } => {}
+            TopologyEvent::PhaseChanged { .. } => {}
             TopologyEvent::DialFailed { .. } => {}
             TopologyEvent::PingCompleted { .. } => {}
         }
--- a/crates/swarm/node/tests/bootnode_cluster.rs
+++ b/crates/swarm/node/tests/bootnode_cluster.rs
@@ -23,7 +23,7 @@ use tokio::time::timeout;
 use vertex_swarm_api::{SwarmTopologyPeers as _, SwarmTopologyState as _, SwarmTopologyStats as _};
 use vertex_swarm_primitives::{Bin, NeighborhoodDepth, all_bins};
 use vertex_swarm_test_utils::cluster::{ClusterBuilder, NodeRole};
-use vertex_swarm_topology::TopologyEvent;
+use vertex_swarm_topology::{TopologyEvent, TopologyPhase};
 
 /// Cap on wall-clock time the test will wait for handshakes to complete.
 const HANDSHAKE_TIMEOUT: Duration = Duration::from_secs(10);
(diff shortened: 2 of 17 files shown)


mfw78:
Rebased onto current main (atop #234, #235, #237). Conflicts in \`readiness.rs\`, \`handle.rs\`, \`config.rs\`, \`args.rs\`, and \`routing.rs\` resolved by keeping both sides: the phase machine (\`phase\`/\`time_in_phase\`, \`phase_stability_window\`, \`evaluate_phase\`/\`phase_status\`) now sits alongside the #237 neighborhood-readiness clock (\`neighborhood_stable_for\`, \`neighborhood_stability_window\`, \`update_neighborhood_stability\`). The two stability clocks use different time bases, so the phase machine keeps \`web_time::Instant\` (aliased \`PhaseInstant\`) while the neighborhood clock keeps \`tokio::time::Instant\`; both are imported.

The only red CI job is \`test\`, failing solely on \`kademlia::routing::tests::test_eviction_tiebreak_keeps_bin_prefix_spread\`. That failure is pre-existing on \`main\` (run 27323560560 is red on the identical test) and comes from the #235 trim logic, not from anything in this PR. All phase-machine and neighborhood-readiness tests pass; clippy, fmt, doc, and wasm are green.

PR #261, opened 2026-06-12
ci(pages): deploy the wasm swarm demo to github pages

### LINKED ISSUE #252: wasm-playground: revive or remove the dead browser demo bin
`bin/wasm-playground` is excluded from the workspace and cannot compile standalone.

It references path crates that no longer exist in this repository (`bmt`, `postage`, `ethers-signers`); those primitives moved to nectar. As a result the bin is dead: it is not a workspace member, so CI never builds it, and its path dependencies do not resolve.

We should decide its fate: either revive it into the workspace by re-pointing it at the current crate graph (the nectar primitives plus `vertex-util-runtime` for time and randomness), or delete it.

The sibling `bin/swarm-wasm-lib` is in the same situation and is already flagged in `docs/agents/wasm.md`.

mfw78:
This is the final phase of the browser Swarm client demo: it makes the deployed demo actually connect to live mainnet, ships the GitHub Pages deploy, and removes the dead wasm bins.

## What the browser showed

A headless-browser run against this build (served with COOP/COEP) boots the wasm module, mounts the live topology panel, mints an ephemeral overlay, resolves the mainnet bootnodes over DoH, and dials them over secure websockets. The dial reaches the live `libp2p.direct` AutoTLS endpoint: the wss/TLS WebSocket connection opens and the libp2p noise/multistream upgrade is attempted (the events log shows `protocol negotiation failed`, i.e. the upgrade after the socket opened, not a transport rejection). Completing the upgrade to a connected peer needs network reachability to the bee nodes behind that endpoint, which the CI sandbox network does not have. The connect path itself is sound end to end; the deployed page in a real browser is where peers appear and the bins fill.

## The fixes (at the source, not in the demo)

A clean wasm build hid several runtime faults that only fire once the client runs. Each is fixed where it lives:

- Wall clocks go through `web-time` instead of `std::time`, which is the unsupported-platform stub on wasm32 and panics on first read. This covers the GCRA rate limiter, the gossip intake clock, and the client protocol handler. `nectar`'s `Timestamp` and the postage timestamp are fixed upstream in [nxm-rs/nectar#52](https://github.com/nxm-rs/nectar/pull/52) (released as 0.2.0); the workspace nectar rev is bumped here. libp2p's connection-handler delays go through `futures-timer`, so its `wasm-bindgen` browser-clock feature is turned on.
- Timers go through a new `vertex_tasks::time::sleep` (tokio on native, `gloo-timers` in the browser); tokio's timer driver does not run on wasm32. It is used for the topology evaluator, gossip, the behaviour's dial pacing, and the peer-manager tick. The browser timer future is `!Send`, so the affected struct fields and tasks are cfg-gated to drop the `Send` bound and spawn on the browser event loop on wasm.
- A browser client has no listeners, so the listen-address IP-capability heuristic left it unable to dial anything; it now reports `Dual` on wasm, since the browser dials either address family outbound.
- The DoH resolver and the snapshot fallback rewrite the AutoTLS `/ip4/.../tls/sni/<host>/ws` leaves into the `/dns4/<host>/tcp/<port>/tls/ws` form the browser websocket transport dials by hostname.

## The rayon thread pool

It is not the blocker it looked like. The connect/topology path never parallelizes, so `initThreadPool` is never called and no rayon worker-URL is ever requested. The demo still needs the threaded-wasm link recipe (shared importable memory plus `build-std`), because `wasm-bindgen-futures` schedules its task queue with `Atomics.waitAsync`, which requires a `SharedArrayBuffer`. That recipe lives in `bin/swarm-demo/.cargo/config.toml`; `Trunk.toml` keeps `filehash` off (stable module names) and `wasm-opt` off (binaryen needs `--enable-threads`).

## Pages workflow

`.github/workflows/pages.yml` builds the demo with `--public-url /vertex/` (the Pages subpath for `nxm-rs/vertex`) on nightly + `wasm32-unknown-unknown` + `rust-src`, installs trunk, and publishes `dist/` through `actions/upload-pages-artifact` + `actions/deploy-pages`. The workflow validity is CI-checkable; the deploy runs post-merge. The demo is the whole Pages site, served from https://nxm-rs.github.io/vertex/.

## Dead-bin cleanup

`bin/swarm-wasm-lib` and `bin/wasm-playground` referenced deleted bmt/postage/ethers-signers crates and are removed, along with their `[workspace.exclude]` entries. Native `cargo build --all-features` stays green.

Closes #252


mfw78:
Closing as subsumed, not merged. Every change on this branch (the `.github/workflows/pages.yml` deploy, `bin/swarm-demo`, the wasm runtime fixes, the DoH bootnode discovery wiring) landed on main through the squash-merge of #262 (711a16b), which was branched from this branch and carried it forward together with the `Version::V1Lazy` upgrade, the `nxm-rs/rust-libp2p` fork, and the dnsaddr-doh rewrite cleanup. Rebasing this branch onto main yields an empty diff (every change is already present in a newer form), so there is nothing left to merge. Tracked in #252.

PR #262, opened 2026-06-12
fix(swarm-node): lazily negotiate the wasm wss upgrade so the client clears multistream-select

mfw78:
## What

The browser Swarm client (`bin/swarm-demo`, `vertex-swarm-node`'s wasm `build_swarm`) opened its secure WebSocket to the live mainnet bootnodes but failed at libp2p protocol negotiation immediately after the socket opened, surfacing as `dial_failed: protocol negotiation failed` and a zero connected-peer count.

This switches the wasm upgrade from strict `Version::V1` to `Version::V1Lazy`, which clears the multistream-select security round that strict V1 could not complete over the browser WebSocket.

Built on top of #261 (the wasm demo plus Pages deploy), which carries the prerequisites and is not yet on `main`. Merge order: #261 then this.

## Exact negotiation error captured in-browser

With `Version::V1`, the captured console error was:

```
Dialer: Proposed protocol protocol = /noise
Connection attempt to peer failed with Transport([(/dns4/<host>.libp2p.direct/tcp/1635/tls/ws/p2p/<id>,
  Other(Custom { kind: Other, error: Other(Right(Left(Right(Select(ProtocolError(IoError(Kind(BrokenPipe)))))))) }))])
error = Failed to negotiate transport protocol(s): [... Multistream select failed: ... broken pipe ...]
```

The failure is `Select(ProtocolError(IoError(BrokenPipe)))`: a broken pipe inside the multistream-select security round, right after the dialer proposed `/noise`.

## Bee-vs-vertex protocol comparison

The offered protocol identifiers all match the live network, confirmed by reading the live node's libp2p configuration and the pinned `rust-libp2p` rev `3e72d4c`:

- Security: live node offers libp2p-TLS and Noise (TLS preferred); the browser client offers Noise only. Multistream-select intersects on `/noise`. Match.
- Muxer: both offer `/yamux/1.0.0`. The live node supports early muxer negotiation via the Noise extension but falls back to multistream muxer selection; the pinned `rust-libp2p` noise only carries the WebTransport certhashes extension, so it negotiates yamux over multistream, which the live node accepts. Match.
- Handshake app protocol: both use `/swarm/handshake/15.0.0/handshake`. Match.

So this was never a protocol-string mismatch. I confirmed it directly: a raw browser `WebSocket` to the same live node, driven by hand, completes multistream-select (`/multistream/1.0.0` then `/noise` echoed back) and the full Noise XX handshake (the node returns its 284-byte XX message 2). The network path and the live node are healthy from this browser.

## Root cause

Instrumenting `WebSocket` in the page showed the browser client's own socket lifecycle: open, the node's `/multistream/1.0.0` echo arrives, the client sends its 28-byte multistream proposal, then the `libp2p-websocket-websys` `Connection` is dropped one millisecond later (`close(1000, "connection dropped")`, the transport's own `Drop` string), before reading the confirmation.

The browser `WebSocket` is message-framed, and `libp2p-websocket-websys`'s `AsyncWrite::poll_flush` reports completion only once the socket's `bufferedAmount` drains, which it observes through a periodic timer rather than synchronously. Strict `Version::V1` writes the proposal and then parks on a flush-then-await round trip (`SendProtocol` to `FlushProtocol` to `AwaitProtocol`) waiting for the peer's `/noise` confirmation. Under the browser executor that synchronous flush barrier leaves the security negotiation parked mid-handshake and the connection is torn down with a broken pipe.

## The fix

`Version::V1Lazy`. The dialer offers a single security (`/noise`) and a single muxer (`/yamux/1.0.0`), so the lazy variant settles on each optimistically (`Negotiated::expecting`) and folds the confirmation into the first read of the following handshake bytes, removing the synchronous flush barrier. The responder still replies with a regular V1 response, so this stays wire-compatible with the live network. The change is confined to the wasm `build_swarm` sibling; the native TCP path is `#[cfg(not(target_arch = "wasm32"))]` and untouched. Native `cargo build`/`clippy -D warnings` on `vertex-swarm-node` stay green, and the wasm demo rebuilds with the threaded toolchain.

## Live verification status

V1Lazy provably advances the handshake: after the change, the in-browser error moves from `Select(ProtocolError(...))` (stuck inside multistream-select) to `Apply(Io(Kind(BrokenPipe)))` with `Handshake failed: broken pipe` (now inside the Noise handshake), and the console trace shows the client encrypting and emitting its first Noise message. The security negotiation round is cleared.

A confirmed connected peer (count > 0) was not yet reached in this sandbox. The residual failure is the same browser-WebSocket flush behavior one layer deeper, in the Noise handshake send, and it reproduces under the demo's threaded `+atomics` build (forced by `nectar-primitives` pulling `wasm-bindgen-rayon`). That residual is in the pinned `rust-libp2p` `websocket-websys` flush path under the wasm executor and is tracked as follow-up rather than bumping the pinned rev here. This PR lands the multistream-select fix with the captured evidence above; the user should run the final live verification on a desktop browser.

Part of #252


crates/net/dnsaddr-doh/src/parse.rs
```diff
@@ -102,11 +102,78 @@ pub fn is_browser_dialable_wss(addr: &Multiaddr) -> bool {
...
+            // The raw IP host and the SNI component are folded into the leading
+            // `/dns4` host above.
+            Protocol::Ip4(_) | Protocol::Ip6(_) | Protocol::Sni(_) => {}
+            other => rewritten.push(other),
+        }
+    }
+
+    rewritten
+}
+
```
mfw78: Is this required anymore given the changes that we made upstream in our forked copy of libp2p?

crates/net/dnsaddr-doh/src/resolver.rs
```diff
@@ -127,8 +129,14 @@ fn resolve_into<'a, F: TxtFetcher>(
...
-            } else if is_browser_dialable_wss(&addr) && seen_leaves.insert(addr.to_string()) {
-                leaves.push(addr);
+            } else if is_browser_dialable_wss(&addr) {
+                // Rewrite the network's `/ip4/.../tls/sni/<host>/ws` AutoTLS form
+                // into the `/dns4/<host>/tcp/<port>/tls/ws` shape the browser
+                // websocket transport can dial, deduplicating on the dialable form.
+                let dialable = to_browser_dialable_wss(&addr);
+                if seen_leaves.insert(dialable.to_string()) {
+                    leaves.push(dialable);
+                }
```
mfw78, line 137: Is this required anymore given our forked changes to rust-libp2p?

crates/swarm/node/src/node/builder.rs
```diff
@@ -247,10 +247,23 @@ where
...
+/// flush as complete only once the socket's buffered amount drains, which it
+/// observes through a periodic timer rather than synchronously. Under the
+/// browser executor the strict-`V1` flush-then-await round trip leaves the
+/// security negotiation parked mid-handshake, and the connection is torn down
+/// before the peer's protocol confirmation is read. The dialer offers a single
+/// security (`/noise`) and a single muxer (`/yamux/1.0.0`), so the lazy variant
+/// settles on each optimistically and folds the confirmation into the first
+/// read of the following handshake bytes. The responder still replies with a
+/// regular `V1` response, so this stays wire-compatible; it only removes the
+/// synchronous flush barrier the browser transport cannot satisfy.
```
mfw78, line 266: Given that the V1Lazy was done before we repaired the issues with upstream rust-libp2p, is it required anymore for us to deviate / be explicit about this?

crates/net/dnsaddr-doh/src/parse.rs
mfw78: No longer needed. Removed in 10fe96e (on the #263 fork-pin branch this stacks into). The forked `libp2p-websocket-websys` now parses the AutoTLS `/ip4/.../tls/sni/<host>/ws` leaf directly (it opens the secure socket against the `/sni` host and keeps the `/tcp` port), so `to_browser_dialable_wss` is redundant. Note it is only removable together with the fork: on #262 alone the unfixed transport still rejects the `/sni` form, which is why the removal lands with the fork pin in #263.

crates/net/dnsaddr-doh/src/resolver.rs
mfw78: Same as the parse.rs thread: removed in 10fe96e. The resolver now keeps the original `/ip4/.../tls/sni/<host>/ws` leaf, which the forked transport dials directly. This also matches the gossiped-peer dial path, which already connects to AutoTLS peers in that exact form with no rewrite.

crates/swarm/node/src/node/builder.rs
mfw78: V1Lazy is not actually a deviation. The native `.with_tcp(...)` path and the upstream `with_websocket`/`with_relay` builder phases all use `Version::V1Lazy` internally (libp2p/src/builder/phase/{tcp,websocket,relay}.rs); the original explicit `V1` was the outlier, so switching to V1Lazy made the wasm path match native. The fork's flush fix means `V1` would also work now, but V1Lazy is the standard SwarmBuilder choice, so keeping it. No change needed.

Issue #287, opened 2026-06-13
retrieval/pushsync: verify chunks at relay boundaries and report malformed chunks to peer scoring


mfw78:
## Problem

The codecs already verify chunk integrity on decode (BMT/SOC address reconstruction via `StampedChunk::reconstruct` in `crates/swarm/net/retrieval/src/codec.rs` and `crates/swarm/net/pushsync/src/codec.rs`), but **nothing reports the outcome to peer scoring, and nothing re-verifies on the outbound side**. Today a peer can serve corrupted chunks or fail every request with zero score impact:

- `crates/swarm/node/src/protocol/handler.rs` (retrieval response handling ~482-507, pushsync receipt handling ~509-531) logs errors but never calls `PeerManager::report_peer()`.
- `crates/swarm/node/src/client_service.rs` sees `RetrievalFailed` / `PushFailed` events and does not report either.
- The scoring vocabulary already exists in `crates/swarm/api/src/reporting.rs`: `RetrievalSuccess`/`RetrievalFailure` (±2.0), `PushSuccess`/`PushFailure` (±2.0), `InvalidData` (-10.0), `MaliciousBehavior` (-50.0) — it is simply not wired from these protocols.
- Outbound `ServeChunk` / relayed deliveries accept a `StampedChunk` from the application layer and encode it without re-verification (`handler.rs` ~680-700).

## Required behaviour

1. **Verify at every relay boundary.** When acting as a forwarder:
   - retrieval: verify the chunk received from upstream **before** relaying it downstream to the original requester;
   - pushsync: verify a chunk pushed to us **before** relaying it onward toward the neighbourhood.
   Inbound decode-time verification already covers the receive edge; add an outbound verification gate (or a type-state guarantee, e.g. only a `VerifiedStampedChunk` can be encoded) so a malformed chunk can never be emitted by us — we must not launder another peer's garbage into our own reputation problem.
2. **Score the sender adversely.** A peer that delivers a chunk failing address/stamp reconstruction must be reported with `InvalidData` (or `MaliciousBehavior` for repeat/structural cases). Plain failures/timeouts report `RetrievalFailure`/`PushFailure`; successes report the positive events so honest peers climb.
3. **Single reporting authority.** Route everything through `PeerManager::report_peer()` like accounting violations already do (`crates/swarm/bandwidth/core/src/accounting/mod.rs:129-137`).

## Acceptance

[...]

Issue #130, opened 2026-06-06
ratelimiter: add self-throttle pattern for outbound accounting-gated protocols


mfw78:
## Context

Pushsync and retrieval initiate outbound work that consumes our credit with the remote peer (via pseudosettle / swap accounting). If we initiate faster than the remote replenishes our allowance, our requests get refused and we waste roundtrips. Lighthouse hit the same problem on its consensus RPC and solved it with a `SelfRateLimiter` that delays our own outbound requests until the rate-limiter accepts them.

We need the same shape. Inbound rate limiting on these protocols is already handled by accounting (see closed #127, #128, #129). The remaining gap is on the outbound side: throttle ourselves so we stay under the remote's negotiated allowance for us.

## Proposed primitive

Add a `SelfRateLimiter<K>` in `crates/net/ratelimiter` that:

- Wraps a `KeyedRateLimiter<K>` keyed by `PeerId`.
- Exposes `try_send(peer, cost)` returning `Result<(), DelayUntil>` where `DelayUntil(Duration)` carries the wait hint from the GCRA bucket (the `TooSoon(Duration)` variant already in the limiter).
- Provides a `DelayQueue` integration so callers can park rejected requests and get woken when the bucket can satisfy them.
- Lets the protocol crate provide its own cost function (token count derived from request size / kind).

This is the outbound dual of the inbound shape that landed in #105 and is modeled after `lighthouse_network/src/rpc/self_limiter.rs` (Lighthouse's outbound throttle).

## Acceptance criteria

- [ ] `SelfRateLimiter<K>` available in `vertex-net-ratelimiter`.
- [ ] Unit tests cover: accept under budget, reject + delay over budget, wake on bucket replenish, per-key independence, `clear(&K)` on disconnect.
- [ ] Documentation in the crate-level `lib.rs` explains the inbound vs outbound symmetry (inbound = our `KeyedRateLimiter`, outbound = `SelfRateLimiter`).

## Consumers

Filed separately:

- pushsync outbound self-throttle: TBD link
- retrieval outbound self-throttle: TBD link

Both depend on this primitive.

mfw78:
Consumer context: the typed client upload/download path landed in #170-#174. The accounting-aware peer selection/gating counterparts are now filed at #175 (retrieval) and #176 (pushsync), distinct from this self-throttle primitive but designed to compose with it (selection picks affordable peers, SelfRateLimiter paces requests to each).

mfw78:
[...]

PR #292, opened 2026-06-13
refactor(swarm-node): per-request response channels for outbound retrieval and pushsync

### LINKED ISSUE #177: pushsync: outbound failure hangs push_chunk (no PushFailed emitted)
## Summary

An outbound pushsync failure does not resolve `ClientHandle::push_chunk`; the caller blocks until the connection drops.

## Detail

Landed with the typed upload path in #170-#174. On the push side:

- `ClientHandle::push_chunk` (`crates/swarm/node/src/client_service.rs`) registers a pending push keyed by chunk address and awaits a oneshot.
- The receipt path resolves it: `ReceiptReceived` -> `complete_push`. The failure path is `ClientEvent::PushFailed { address, .. }` -> `fail_push`.
- But an outbound pushsync error in `crates/swarm/node/src/protocol/handler.rs` emits `HandlerEvent::Error { protocol: "pushsync", .. }` with no chunk address. That surfaces as `ClientEvent::ProtocolError` (`crates/swarm/node/src/protocol/behaviour.rs`), whose handler in `client_service.rs` only logs (`// TODO: Handle specific errors`). It never calls `fail_push`.

Net effect: on outbound failure, `PushFailed` is never emitted, the pending oneshot is never resolved, and `push_chunk` hangs until the connection drops and the sender is dropped (yielding `Cancelled`).

## Fix

Emit `ClientEvent::PushFailed { peer, address, error }` from the outbound pushsync failure path so `fail_push` resolves the pending push promptly. This requires carrying the chunk address through the outbound failure in `crates/swarm/node/src/protocol/{handler.rs,behaviour.rs}` (the handler currently drops it in the generic `Error` variant for pushsync).

## Acceptance criteria

- [ ] Outbound pushsync failure carries the chunk address through the handler.
- [ ] `ClientEvent::PushFailed { address, .. }` is emitted on outbound failure.
- [ ] `push_chunk` returns an error promptly on outbound failure rather than waiting for the connection to drop.
- [ ] Test covers the outbound-failure-resolves-push case.

### LINKED ISSUE #178: retrieval: restructure outbound request rendezvous for concurrent fan-out (racing futures, no correlation id)
## Summary

Retrieval responses are correlated by chunk address alone, so two concurrent requests for the same address collide. As a result `SwarmClient::get` fans out sequentially instead of concurrently.

## Detail

Landed with the typed download path in #170-#174. In `crates/swarm/node/src/client_service.rs`:

- `pending_retrievals: HashMap<ChunkAddress, oneshot::Sender<RetrievalResult>>` keys pending requests by chunk address.
- `complete_retrieval` / `fail_retrieval` remove by address.

Two in-flight retrievals for the same chunk address (the natural fan-out case: ask the closest N peers in parallel) overwrite each other's oneshot in the map: the second `insert` drops the first sender (spurious `Cancelled`), and a failure event from peer A removes peer B's entry, discarding B's success. To stay correct, `NetworkChunkProvider::retrieve_chunk` (`crates/swarm/builder/src/providers.rs`) tries peers sequentially, and `SwarmClient::get` cannot fan out.

## Approach: no correlation id needed; make each request a self-contained future

An explicit correlation id is unnecessary. The wire protocol already has natural correlation: each outbound request opens its own libp2p substream and the response arrives on that same stream. The handler events already carry the peer (`ChunkReceived { peer, address, .. }`, `RetrievalFailed { peer, address, .. }` in `crates/swarm/node/src/protocol/events.rs`). The only collision point is the rendezvous map in `ClientHandle`.

Preferred fix: delete the pending maps entirely. Thread the `oneshot::Sender<RetrievalResult>` through `ClientCommand::RetrieveChunk` into the per-connection handler, which stores it with the outbound substream state and resolves it directly when the stream completes (success or failure). Each request becomes a self-contained future; fan-out is then plain bounded future racing (`select_ok` / `FuturesUnordered`) at the `swarm_client` layer. This also removes the bug class behind #177: that hang exists precisely because the failure path loses the address needed to find the map entry, and with no map there is nothing to lose.

Fallback (minimal) fix: re-key `pending_retrievals` by `(peer, address)`. The completion events already carry both. Same-peer duplicate requests for one address must then be coalesced or rejected.

Constraints either way:

- Racing costs real AU on every branch: a losing branch still consumed the remote's bandwidth and our reserved balance. Accounting prepare/release (RAII actions) must fire when the losing future is dropped, and race width must be bounded with a stagger (2-3 peers, delayed second attempt) rather than blasting all closest peers at once. Interacts with the outbound self-throttle work (#130/#132).
- `fail_retrieval` currently discards the error string; racing wants per-branch errors so the caller can distinguish not-found from timeout.

## Acceptance criteria

- [ ] Concurrent retrievals of the same address from different peers do not collide (no shared address-keyed rendezvous, or rendezvous keyed by `(peer, address)`).
- [ ] `retrieve_chunk` fans out to a bounded, staggered set of closest peers and resolves on first verified success, cancelling the rest.
- [ ] Dropped (losing) request futures release their accounting reservations.
- [ ] Per-branch errors are preserved and surfaced to the caller.
- [ ] Test covers concurrent retrieval of the same address from multiple peers, including one branch failing while another succeeds.

mfw78:
## Problem

Outbound retrieval and pushsync correlated responses to callers through `ClientHandle` maps keyed by chunk address alone (`pending_retrievals`, `pending_pushes`). Three consequences:

- Two in-flight requests for the same address overwrote each other's channel, and a failure from one peer evicted another peer's pending entry, so fan-out was impossible and `SwarmClient::get` walked peers sequentially (#178).
- `ClientEvent::RetrievalFailed` / `PushFailed` were never emitted by any code path: handler failures surfaced only as generic `ProtocolError`, which nothing routed back to the pending maps. Any wire-level failure (delivery error, receipt rejection, dial upgrade error, timeout) left the caller's channel in the map forever, so both `get` and `put` could hang indefinitely (#177, and the same bug class on the retrieval side).
- `SwarmClient::put` only ever tried the single closest candidate.

## Change

Thread the caller's oneshot sender through `ClientCommand` and `HandlerCommand` into `ClientOutboundInfo`. The info travels with the outbound substream and returns with its completion (`FullyNegotiatedOutbound`) or failure (`DialUpgradeError`), so the substream is the correlation and no shared rendezvous state exists. The pending maps and their `complete_*`/`fail_*` plumbing are deleted.

Every termination path now resolves the caller: chunk delivery, wire-level delivery error, receipt and receipt rejection, dial upgrade error, unknown peer (new `RetrievalError::NotConnected`), queue overflow, and connection teardown (channel drop surfaces as `Cancelled`). The handler emits typed `RetrievalFailed`/`PushFailed` events carrying peer and address, which the behaviour forwards as the previously dead `ClientEvent` variants, ready for peer scoring (#287).

With requests self-contained, `SwarmClient::get` races the closest candidates with a staggered start: the best candidate is queried immediately, each 500ms tick (or an attempt failure, instantly) adds the next, and the first verified chunk wins; losing branches are dropped with their channels. The stagger bounds the accounting cost of the race, pending the self-throttle work (#130/#132). `SwarmClient::put` now walks all candidates sequentially instead of giving up after the first; pushes are deliberately not raced since every branch would deliver and be charged for the chunk.

`futures-timer` (already pinned workspace-wide with the `wasm-bindgen` feature for the browser clock) moves from the wasm-only section to a regular dependency for the stagger timer, keeping `vertex-swarm-node` wasm-clean.

## Tests

- `concurrent_retrievals_for_same_address_resolve_independently`: the exact aliasing failure of the old design.
- `get_starts_next_candidate_on_failure_and_returns_first_success`: failure frees a slot immediately.
- `get_staggers_in_a_second_candidate_while_the_first_is_silent`: stagger brings in candidate two and its response resolves the race.
- `push_chunk_fails_when_push_is_rejected` reworked as the #177 regression test: rejection resolves the caller instead of hanging.
- Existing selector/affordability and failure-path tests reworked onto the response channels.

Closes #177
Closes #178

### DIFF
--- a/crates/swarm/node/src/protocol/behaviour.rs
+++ b/crates/swarm/node/src/protocol/behaviour.rs
@@ -175,7 +175,7 @@ impl ClientBehaviour {
                     });
                 } else {
                     debug!(%peer, "Unknown peer for retrieval");
-                    let _ = response.send(Err(crate::RetrievalError::NotConnected));
+                    let _ = response.send(Err(crate::ChunkTransferError::NotConnected));
                 }
             }
             ClientCommand::PushChunk {
@@ -193,7 +193,7 @@ impl ClientBehaviour {
                     });
                 } else {
                     debug!(%peer, "Unknown peer for push");
-                    let _ = response.send(Err(crate::RetrievalError::NotConnected));
+                    let _ = response.send(Err(crate::ChunkTransferError::NotConnected));
                 }
             }
             ClientCommand::ServeChunk {
--- a/crates/swarm/node/src/protocol/behaviour.rs
+++ b/crates/swarm/node/src/protocol/behaviour.rs
@@ -161,32 +161,39 @@ impl ClientBehaviour {
                     debug!(%peer, "Unknown peer for pricing announcement");
                 }
             }
-            ClientCommand::RetrieveChunk { peer, address } => {
+            ClientCommand::RetrieveChunk {
+                peer,
+                address,
+                response,
+            } => {
                 if let Some(&peer_id) = self.overlay_peers.get(&peer) {
                     debug!(%peer_id, %peer, %address, "Retrieving chunk");
                     self.push_event(ToSwarm::NotifyHandler {
                         peer_id,
                         handler: libp2p::swarm::NotifyHandler::Any,
-                        event: HandlerCommand::RetrieveChunk { address },
+                        event: HandlerCommand::RetrieveChunk { address, response },
                     });
                 } else {
                     debug!(%peer, "Unknown peer for retrieval");
+                    let _ = response.send(Err(crate::RetrievalError::NotConnected));
                 }
             }
             ClientCommand::PushChunk {
                 peer,
                 address,
                 chunk,
+                response,
             } => {
                 if let Some(&peer_id) = self.overlay_peers.get(&peer) {
                     debug!(%peer_id, %peer, %address, "Pushing chunk");
                     self.push_event(ToSwarm::NotifyHandler {
                         peer_id,
                         handler: libp2p::swarm::NotifyHandler::Any,
-                        event: HandlerCommand::PushChunk { chunk },
+                        event: HandlerCommand::PushChunk { chunk, response },
                     });
                 } else {
                     debug!(%peer, "Unknown peer for push");
+                    let _ = response.send(Err(crate::RetrievalError::NotConnected));
                 }
             }
             ClientCommand::ServeChunk {
@@ -360,6 +367,28 @@ impl ClientBehaviour {
                     signature,
                     nonce,
                     storage_radius,
+                }));
+            }
+            HandlerEvent::RetrievalFailed {
+                overlay,
+                address,
+                error,
+            } => {
+                self.push_event(ToSwarm::GenerateEvent(ClientEvent::RetrievalFailed {
+                    peer: overlay,
+                    address,
+                    error,
+                }));
+            }
+            HandlerEvent::PushFailed {
+                overlay,
+                address,
+                error,
+            } => {
+                self.push_event(ToSwarm::GenerateEvent(ClientEvent::PushFailed {
+                    peer: overlay,
+                    address,
+                    error,
                 }));
             }
             HandlerEvent::Error {
--- a/crates/swarm/node/src/protocol/events.rs
+++ b/crates/swarm/node/src/protocol/events.rs
@@ -25,21 +25,21 @@ use vertex_swarm_net_pseudosettle::PaymentAck;
 use vertex_swarm_net_swap::SignedCheque;
 use vertex_swarm_primitives::{OverlayAddress, StampedChunk, StorageRadius, SwarmNodeType};
 
-use crate::client_service::{RetrievalError, RetrievalResult};
+use crate::client_service::{ChunkTransferError, RetrievalResult};
 
 /// Channel on which an outbound retrieval request resolves.
 ///
 /// The sender travels with the request from the caller through the behaviour
 /// and handler into the outbound substream state, so the response (or any
 /// failure along the way) resolves the caller directly. Dropping the sender
-/// anywhere on that path surfaces as [`RetrievalError::Cancelled`].
-pub type RetrievalResponseTx = oneshot::Sender<Result<RetrievalResult, RetrievalError>>;
+/// anywhere on that path surfaces as [`ChunkTransferError::Cancelled`].
+pub type RetrievalResponseTx = oneshot::Sender<Result<RetrievalResult, ChunkTransferError>>;
 
 /// Channel on which an outbound chunk push resolves.
 ///
 /// Same lifecycle as [`RetrievalResponseTx`]: the storer's receipt or the
 /// failure that prevented it resolves the caller directly.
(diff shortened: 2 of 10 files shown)


crates/swarm/node/src/protocol/behaviour.rs
```diff
@@ -161,32 +161,39 @@ impl ClientBehaviour {
...
                     debug!(%peer_id, %peer, %address, "Pushing chunk");
                     self.push_event(ToSwarm::NotifyHandler {
                         peer_id,
                         handler: libp2p::swarm::NotifyHandler::Any,
-                        event: HandlerCommand::PushChunk { chunk },
+                        event: HandlerCommand::PushChunk { chunk, response },
                     });
                 } else {
                     debug!(%peer, "Unknown peer for push");
+                    let _ = response.send(Err(crate::RetrievalError::NotConnected));
```
mfw78: Is this error here correct? or should it actually be a PushError or something like that?

crates/swarm/node/src/protocol/events.rs
```diff
@@ -18,11 +18,29 @@
...
+/// and handler into the outbound substream state, so the response (or any
+/// failure along the way) resolves the caller directly. Dropping the sender
+/// anywhere on that path surfaces as [`RetrievalError::Cancelled`].
+pub type RetrievalResponseTx = oneshot::Sender<Result<RetrievalResult, RetrievalError>>;
+
+/// Channel on which an outbound chunk push resolves.
+///
+/// Same lifecycle as [`RetrievalResponseTx`]: the storer's receipt or the
+/// failure that prevented it resolves the caller directly.
+pub type PushResponseTx = oneshot::Sender<Result<PushReceipt, RetrievalError>>;
```
mfw78: I know before I left comments around the RetrievalError, can the RetrievalError / PushsyncError be potentially unified to simplify error handling, or are they actually sharply distinct?

crates/swarm/node/src/protocol/behaviour.rs
mfw78: Good catch on the naming. The variant itself is correct: NotConnected is operation-agnostic (it means "no active connection to the target peer"), so sending it on a push is right. The real problem was the TYPE name. The enum was already the shared outcome type for both retrieve_chunk and push_chunk, so calling it RetrievalError made the push path read wrong. It is now ChunkTransferError, with the variants grouped by which operation can produce them (shared, retrieval-specific NotFound, push-specific PushRejected).

crates/swarm/node/src/protocol/events.rs
mfw78: At the node level they are now unified: there is a single ChunkTransferError that both get (retrieve_chunk) and put (push_chunk) resolve through, since the shared variants (ChannelClosed, NotConnected, Cancelled, Protocol) plus the two operation-specific ones (NotFound, PushRejected) cover both paths. That was the misnamed RetrievalError, now renamed.

The wire-level codec errors stay distinct on purpose: vertex_swarm_net_retrieval::RetrievalError speaks delivery semantics (NotFound, InvalidChunk) while vertex_swarm_net_pushsync::PushsyncError speaks receipt semantics (InvalidStorageRadius, receipt fields). Their failure vocabularies genuinely differ and each lives in its own protocol crate; they already share the vertex_swarm_net_headers::ProtocolError base where it matters, so merging them would only blur two different protocols.

PR #295, opened 2026-06-13
refactor(swarm-bandwidth): introduce typed accounting unit Au

### LINKED ISSUE #288: accounting: introduce a typed AccountingUnit (AU) and eliminate raw u64/i64 accounting arithmetic
## Problem

Accounting semantically operates in **accounting units (AU)**, but the code represents balances, prices, thresholds and settlement amounts as raw `u64`/`i64`/`AtomicU64`/`AtomicI64`. Nothing in the type system distinguishes an AU amount from a byte count, a token amount in wei, or a rate-limiter token cost — they freely mix in arithmetic.

Inventory of raw-typed AU values (non-exhaustive):

- `crates/swarm/bandwidth/core/src/accounting/peer.rs:16-21` — `balance: AtomicI64`, `reserved_balance`/`shadow_reserved_balance: AtomicU64`, `surplus_balance: AtomicI64`, `payment_threshold`/`disconnect_threshold: u64`
- `crates/swarm/bandwidth/core/src/accounting/mod.rs:118-121` — `current_balance - (price as i64)`, `disconnect_threshold as i64` casts
- `crates/swarm/bandwidth/core/src/config.rs:28-32` — `payment_threshold`, `refresh_rate`, tolerance/early-payment percents as `u64`
- `crates/swarm/bandwidth/pricing/src/fixed.rs:43-44` — `(max_po - proximity + 1) * base_price` as bare `u64` (no overflow guard)
- `crates/swarm/bandwidth/pseudosettle/src/lib.rs:144-173` + `service.rs:159-173,254-274` — allowance/settlement arithmetic in `u64`/`i64`, incl. `refresh_rate.saturating_mul(elapsed)` which can saturate to an effectively infinite allowance
- `crates/swarm/bandwidth/swap/src/service.rs:268-285,382-385` — `U256` cumulative payout ⇄ `u64` AU increment conversion at an untyped boundary

## Proposal

Introduce a dedicated `AccountingUnit` (AU) newtype in `vertex-swarm-api` (or `bandwidth/core`) and migrate the accounting surface to it:

- `AU(u64)` for non-negative amounts (prices, thresholds, reserves, allowances) and a signed companion (`AuBalance(i64)` or `AU(i128)`) for balances — pick sign semantics once, document them (positive = peer owes us).
- Checked/saturating arithmetic methods on the newtype (`checked_mul` for `rate × seconds`, price × proximity-factor) so overflow is handled in one audited place instead of ad-hoc casts.
- Explicit, named conversions at the only two legitimate boundaries: bytes→AU (pricing) and AU→wei/`U256` (swap exchange rate). `as` casts between AU and anything else should no longer typecheck.
- Atomics wrapped (`AtomicAu`) or kept internal to `PeerState` so the public API only speaks AU.
- Serde/display as plain integers — no wire-format change.

## Acceptance

- [ ] No bare `u64`/`i64` AU-denominated fields or function params remain in `bandwidth/*`, pricing, pseudosettle, swap service, or the affordability/selection path (`crates/swarm/node/src/selection.rs`).
- [ ] `rate × elapsed` and price multiplication use checked arithmetic with explicit saturation policy.
- [ ] The `U256`⇄AU conversion in swap is a single named function with overflow error.
- [ ] Clippy/compile enforces the boundary (mixing AU with raw ints fails to build).

mfw78:
Stacked on #292; do not merge before it. The base of this branch is `refactor/outbound-request-rendezvous`.

Closes #288.

## What this does

Accounting semantically operates in accounting units (AU), but balances, prices, thresholds, and settlement amounts were raw `u64`/`i64`/`AtomicU64`/`AtomicI64`, freely mixed in arithmetic with byte counts and wei. This introduces an `Au` newtype in `vertex-swarm-api` and migrates the accounting surface so mixing AU with a raw integer no longer compiles.

## Design: one type, signed

`Au` is a single signed newtype over `i64`. Balances need a sign, and the established convention is positive = peer owes us, negative = we owe the peer, so a signed canonical type is the natural fit. Non-negative quantities (prices, thresholds, reserves, allowances, payments) are still carried as `Au` but constructed through `Au::from_amount(u64)` so the non-negativity is an intentional, named step rather than a bare cast. I did not split into a two-type `Au` plus `AuBalance`: every method either reads a balance or compares a price against a balance, so a single signed type kept the call sites simpler and the trait surface uniform without losing safety.

`Add`/`Sub`/`Neg`/`Sum` and the assign variants come from `derive_more`. Checked and saturating arithmetic live on the type, and `checked_scale(factor)` is the single audited multiplication. There is deliberately no `Mul` by a raw integer. `Debug`, `Display`, `Clone`, `Copy`, `PartialEq`, `Eq`, `PartialOrd`, `Ord`, `Hash`, and serde (transparent) all render as a plain integer, so the persisted snapshot and any serialized form stay byte-identical. AU is an internal representation only and never reaches the wire as anything other than the same plain integers as before.

## Overflow bugs fixed

The `rate * elapsed` allowance in pseudosettle previously used a raw `saturating_mul`, which on overflow saturated to `u64::MAX`, an effectively infinite allowance that would forgive arbitrary debt. It now goes through `Au::checked_scale` with an explicit saturation-then-cap policy, so the allowance is always bounded by the actual debt and never wraps into a wrong value. The proximity price formula `(max_po - proximity + 1) * base_price` in pricing also used an unguarded multiply; it now uses `checked_scale` and saturates to the maximum price on overflow, which simply fails affordability rather than mispricing the chunk. The disconnect-threshold tolerance markup and the swap early-payment trigger use the same checked scaling.

## Boundary conversions

The only sanctioned AU crossings are named single functions: the bytes-to-AU pricing step, and the `U256` to/from AU helpers at the swap cheque boundary and the pseudosettle wire boundary. The swap conversion preserves the existing `AmountOverflow` error so over-large cheque increments are rejected, not wrapped.

## Crates migrated

`vertex-swarm-api` (new `Au`, the `SwarmPeerState`/`SwarmSettlementProvider`/`SwarmAccountingConfig`/`SwarmPeerBandwidth`/`SwarmBandwidthAccounting`/`SwarmPricing`/`PeerAffordability` traits), `vertex-swarm-bandwidth` (accounting core, peer state, actions, error, config, no-op accounting and settlement), `vertex-swarm-bandwidth-pricing`, `vertex-swarm-bandwidth-pseudosettle`, `vertex-swarm-bandwidth-swap`, and `vertex-swarm-node` (selection plus the `swarm_client` test impls) and `vertex-swarm-builder` (the affordability wiring test).

## Tests

Added `checked_scale` overflow coverage on the newtype, a pseudosettle test proving the allowance no longer overflows into a tiny value, a pricing test proving the price saturates rather than wrapping, and an AU to wei round-trip in swap. Existing accounting tests were adapted to `Au`.

## Deferred

The deeper pseudosettle robustness items (backward-clock handling, an atomic settle-amount snapshot, settlement-race fixes) are out of scope here and remain a separate follow-up. This PR is the typing migration plus only the overflow fixes the typing forces.


crates/swarm/bandwidth/swap/src/service.rs
mfw78, line 1: Shouldn't the Au type just have an Into<> or something like this for U256 to simplify this?

mfw78: Good call. The conversion now lives on the type as standard traits in crates/swarm/api/src/accounting.rs: TryFrom<U256> for Au and TryFrom<Au> for U256, with an AuConversionError carrying the offending value.

Both directions are TryFrom rather than From/Into because neither is total: Au is a signed i64 while U256 is unsigned and far larger. A U256 can exceed i64::MAX (rejected, not wrapped), and a negative Au (we owe the peer) has no U256 representation (rejected, not clamped to a bogus value). A blanket infallible Into would have to silently lie in those cases.

The swap call sites drop the hand-rolled au_from_wire/wire_from_au helpers: the receive side is now increment.try_into()? and the send side U256::try_from(amount), with From<AuConversionError> for SwapSettlementError preserving the existing AmountOverflow path (and a new NegativeAmount variant for the signed direction). pseudosettle keeps its own helpers for now; AGENTS.md is updated to point at the new trait impls.

PR #297, opened 2026-06-13
refactor(swarm-net): stop processing the adversarial retrieval/pushsync error string

### LINKED ISSUE #296: retrieval/pushsync: collapse remote wire-error strings into a typed bounded WireError enum
## Problem

Retrieval and pushsync carry the remote peer's failure as a raw `String` on the wire and internally: bee writes any Go `err.Error()` into `pb.Delivery.Err` (retrieval.go:429) and `pb.Receipt.Err` (pushsync.go:189), and Vertex models these as `Delivery::Error(String)` (`crates/swarm/net/retrieval/src/codec.rs:68`) and `Receipt.error: Option<String>` (`crates/swarm/net/pushsync/src/codec.rs:76`). Passing unbounded remote strings around internally is a memory/DoS vector and a poor error surface. Collapse them into a typed enum at the codec boundary, parsed similarly to how we classify bee/vertex user-agent strings.

## Findings from bee analysis (v2.7.0-rc5)

Two facts shape the design:

1. Bee never matches on these wire strings. Both the retrieval requester and the pushsync pusher/forwarder wrap a non-empty `Err` opaquely in `p2p.ChunkDeliveryError` and never `errors.Is`/`errors.As` against it; they branch only on locally-produced sentinels. So byte-exact emission is NOT required for interop. The enum is for our own inbound classification (retry/scoring/metrics); outbound we may emit our own clean strings.
2. The wire `Err` is effectively unbounded: capped only by the 128 KiB p2p frame, embeds 64-hex addresses, and nests across every forwarding hop (each prepends `retrieve chunk: ` / `handler: push to closest chunk %s: ` / `delivery of chunk failed: `). A malicious or deeply-forwarded peer can approach the frame limit. We must truncate unknown strings to a small cap (~256 bytes) at parse.

Stable strings worth typing (exact-match sentinels):
- `no peer found` (topology.ErrNotFound) -> NoRoute
- `storage: not found` (storage.ErrNotFound) -> NotFound
- `invalid chunk` (swarm.ErrInvalidChunk) -> InvalidChunk
- `context deadline exceeded` / `context canceled` -> Timeout
- `connection not initialized yet`, `failed to lock` -> Accounting
- `node warmup time not complete` (ErrWarmup, pushsync, nested) -> Warmup
- `could not push chunk` (ErrNoPush, pushsync, nested) -> NoPush

Everything else is `fmt.Errorf`-wrapped (dynamic): match the stable prefix where useful, otherwise fall through to a bounded `Unknown`.

Out of scope here (handled in #293): `ErrShallowReceipt` and `ErrOutOfDepthStoring` never cross the wire; bee detects a shallow receipt by local proximity recomputation on the returned receipt, not by string matching. `ErrOutOfDepthStoring` is dead code in bee.

## Proposed design

A single shared, bounded `WireError` enum in `vertex-swarm-primitives` (both protocol codecs already depend on it for `ReconstructError`/`StampedChunk`):

```rust
pub enum WireError {
    NoRoute, NotFound, InvalidChunk, Timeout, Accounting,
    Warmup, NoPush,            // pushsync-specific
    Unknown,                   // raw remainder truncated to <= 256 bytes
}
```

- `WireError::parse(&str)`: exact-match sentinels, then stable-prefix fallthrough, else `Unknown` (truncated).
- `retriable()` helper for the get/put retry loops.
- `as_wire_str()` -> `&'static str` for our own outbound emission (clean, bounded).
- Replace `Delivery::Error(String)` and `Receipt.error: Option<String>` with the typed form; parse on decode, emit on encode.
- Fold into #287's `FailureKind` and the `ChunkTransferError` surface so the node passes the enum, never a `String`.

## Acceptance

- [ ] `Delivery::Error` and the pushsync receipt error are typed `WireError`, not `String`.
- [ ] Inbound decode parses the wire string into `WireError`; unknown strings truncated to a hard byte cap.
- [ ] Outbound encode emits a clean bounded string for our own failures.
- [ ] `retriable()` drives the get/put retry decisions; node layer carries the enum.
- [ ] Tests: each known sentinel parses to the right variant; an oversized unknown string is truncated; round-trip through the codec.
- [ ] No wire-format change (still a protobuf string field); bee interop unaffected (bee does not match on content).

Stacked on #294 (depends on its `FailureKind`/`ChunkTransferError`).

mfw78:
## Approach: never read the error string, and never send one

The `Err` field is adversarial input the remote fully controls, and the reference node never introspects it (it only checks presence, else falls back to `data`/signature validity). So this strips it from both directions.

### Receive: detect failure structurally, never from the string
- Retrieval failure = empty `data` (chunk reconstruction fails).
- Pushsync failure = an empty (non-65-byte) `signature`.

The `err` field is dropped from **both** protos, so an inbound value is skipped by quick-protobuf without allocation.

### Send: reset the stream, don't write a failure frame
A responder that cannot serve resets the substream instead of encoding a failure frame. The reference pusher/requester reads the reset (or EOF) as a failure at every hop:

```go
// pushsync.go:643, used by origin AND forwarder
if err = r.ReadMsgWithContext(ctx, &rec); err != nil { return nil, err } // reset = failed push
```

This also removes the need for the one-byte forwarder marker from the earlier revision of this PR: the forwarder signature-skip only bites when a receipt is actually *read*, and resetting means there is no receipt to read. It matches the reference's own idiom, which resets on failure (`pushsync.go:194`, `pushChunkToPeer`'s deferred `Reset()`).

### Other
- `Delivery::Error` and the pushsync receipt failure are payload-free; the node carries a stringless `ChunkTransferError::Remote`.
- `MAX_MESSAGE_SIZE` tightened 5 MiB → 16 KiB on both protocols (a legitimate frame is under ~4.4 KiB), bounding any transient field allocation ~320x.

## Wire compatibility

Confirmed against bee v2.7.0-rc5 (analysis in #296): bee discriminates success/failure by `Err != ""` first, else by `data`/signature validity, and never parses, length-checks, or compares the content; an empty/absent field is identical in proto3; a reset/EOF on the receipt or delivery read is a failure, not a hang or a false success. So a bee peer still detects our failures and we detect theirs. A test (`decodes_reference_failure_frame_ignoring_err_field`) hand-builds a reference-shaped failure frame carrying an `err` string and asserts we skip it and decode a failure.

## Tests

Codec failure round-trips updated to the payload-free, marker-free form; the structural (empty-signature) failure signal is asserted; the reference-frame skip test proves wire-compat. fmt + clippy `--all-features` + tests green locally across the touched crates.

Stacked on #294. Closes #296.

### DIFF
--- a/crates/swarm/net/proto/proto/retrieval.proto
+++ b/crates/swarm/net/proto/proto/retrieval.proto
@@ -10,9 +10,18 @@ message Request {
   bytes addr = 1;
 }
 
-// Delivery of a chunk with optional error.
+// Delivery of a chunk.
+//
+// An honest failure is signalled structurally by an empty `data` AND an empty
+// `stamp`: the reference reports a retrieval failure with an empty delivery
+// (it does not reset the stream for retrieval), and that is the only signal we
+// read. A non-empty `data` claims a chunk and must reconstruct and validate, or
+// it is malformed (the two are scored differently). Field 3 was a remote error
+// string in the reference; we never read it (it is adversarial input the
+// reference itself does not introspect), so it is omitted here. An inbound
+// field-3 value is skipped without allocation, and we emit nothing on it. A
+// requester detects our failures from the empty `data` (chunk validation fails).
 message Delivery {
   bytes data = 1;
   bytes stamp = 2;
-  string err = 3;
 }
--- a/crates/swarm/net/pushsync/src/codec.rs
+++ b/crates/swarm/net/pushsync/src/codec.rs
@@ -11,8 +11,8 @@ use crate::error::PushsyncError;
 /// Codec for pushsync delivery messages.
 pub(crate) type DeliveryCodec = Codec<Delivery, PushsyncError>;
 
-/// Codec for pushsync receipt messages.
-pub(crate) type ReceiptCodec = Codec<Receipt, PushsyncError>;
+/// Codec for pushsync receipt responses.
+pub(crate) type ReceiptCodec = Codec<ReceiptResponse, PushsyncError>;
 
 /// Delivery of a chunk to be stored.
 ///
@@ -63,17 +63,25 @@ impl ProtoMessage for Delivery {
     }
 }
 
-/// Receipt acknowledging chunk storage.
+/// Length in bytes of a well-formed recoverable signature. A failure response
+/// carries an empty signature instead, which is the structural failure signal.
+const SIGNATURE_LEN: usize = 65;
+
+/// A successful storage receipt.
+///
+/// This type models a custody acknowledgement only. The reference does not sign
+/// its failure responses (a failure is `&pb.Receipt{Err: ...}` with no
+/// signature), so there is no signed failure receipt to model and no need for
+/// placeholder fields. A failure is represented by the [`ReceiptResponse::Failed`]
+/// variant, not by a `Receipt` with zeroed fields.
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub struct Receipt {
     /// The address of the chunk.
     pub address: ChunkAddress,
-    /// Signature from the storer.
+    /// Signature from the storer over the chunk address and nonce.
     pub signature: Signature,
     /// Nonce used in signing.
     pub nonce: Nonce,
-    /// Error message if storage failed.
-    pub error: Option<String>,
     /// The storage radius of the storer node.
     pub storage_radius: StorageRadius,
 }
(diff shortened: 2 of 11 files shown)


crates/swarm/net/proto/proto/retrieval.proto
```diff
@@ -10,9 +10,14 @@ message Request {
...
 }
 
-// Delivery of a chunk with optional error.
+// Delivery of a chunk.
+//
+// A failure is signalled structurally by empty `data`. Field 3 was a remote
+// error string in the reference; we never read it (it is adversarial input the
+// reference itself does not introspect), so it is omitted here. An inbound
+// field-3 value is skipped without allocation, and we emit nothing on it. A bee
+// requester detects our failures from the empty `data` (chunk validation fails).
```
mfw78: If we are acting as a forwarder, and there is a problem with the stream whereby the node we have forwarded the request to is unable to fulfill it for some reason, can we just drop the stream here too and not tie some introspection methodology to data / stamp so we can be strict about our read of these fields?

crates/swarm/net/pushsync/src/codec.rs
```diff
@@ -63,18 +63,23 @@ impl ProtoMessage for Delivery {
...
     pub nonce: Nonce,
-    /// Error message if storage failed.
-    pub error: Option<String>,
-    /// The storage radius of the storer node.
+    /// True if this receipt reports a failure rather than custody. The remote's
+    /// reason string is adversarial input we never read, so it is not carried.
+    pub failed: bool,
+    /// The storage radius of the storer node (placeholder on a failure receipt).
     pub storage_radius: StorageRadius,
 }
```
mfw78, line 87: Why do we have placeholders for failure receipts? do failure receipts actually get signed by the reference implementation, are they really needed? Could we not just have Option<Receipt> and if it returns None, it is interpreted as a failure, or alternative have some Enum that captures some context if it is relevant?

crates/swarm/net/pushsync/src/protocol.rs
```diff
@@ -23,7 +22,12 @@ use crate::{
 };
 
 /// Maximum size of a pushsync message (chunk + stamp + overhead).
-const MAX_MESSAGE_SIZE: usize = 5 * 1024 * 1024; // 5 MB
+// A legitimate delivery is at most span(8) + body(4096) + id/sig for a SOC +
+// stamp(113) + a 32-byte address + framing, under ~4.4 KiB. 16 KiB leaves
+// generous headroom while bounding the worst-case adversarial frame (and thus
+// any transient field allocation) to 16 KiB instead of multiple megabytes.
+// Rejecting larger frames is not wire-visible: a conformant peer never sends one.
+const MAX_MESSAGE_SIZE: usize = 16 * 1024; // 16 KiB
```
mfw78: Could we alternatively define this as a multiple of a chunk size? Effectively there should be an upper bound that can be calculated, and it should actually be less than this. But is it possible to be more exhaustive / strict than even this?

crates/swarm/net/retrieval/src/protocol.rs
```diff
@@ -23,7 +23,12 @@ use crate::{
 };
 
 /// Maximum size of a retrieval message (chunk + stamp + overhead).
-const MAX_MESSAGE_SIZE: usize = 5 * 1024 * 1024; // 5 MB
+// A legitimate delivery is at most span(8) + body(4096) + id/sig for a SOC +
+// stamp(113) + framing, under ~4.4 KiB. 16 KiB leaves generous headroom while
+// bounding the worst-case adversarial frame (and thus any transient field
+// allocation) to 16 KiB instead of multiple megabytes. Rejecting larger frames
+// is not wire-visible: a conformant peer never sends one.
+const MAX_MESSAGE_SIZE: usize = 16 * 1024; // 16 KiB
```
mfw78: Same as for pushsync above

crates/swarm/net/proto/proto/retrieval.proto
mfw78: Good question, and the answer splits cleanly into the two sides of the stream.

Send side: as a forwarder whose downstream cannot fulfil, we already drop the stream and tie nothing to data/stamp. `RetrievalResponder::send_error` resets the substream and writes no frame, so there is no introspection on our outbound failure path.

Receive side: the empty-data check stays, but for a reason that is not introspection of an adversarial field. The reference reports a retrieval failure by sending an empty delivery (it does not reset the stream for retrieval), so a peer's honest "not found" arrives as an empty frame. We have to distinguish that honest failure from a malformed chunk, because they score differently downstream: an honest empty-data failure is a protocol failure, while data that is present but does not reconstruct is invalid data and is penalised far harder. Collapsing them would punish peers for honest not-found responses.

To get the strict read you are after, I tightened it: a clean failure now requires `data` AND `stamp` both empty. Once `data` is non-empty the frame claims a chunk and must reconstruct and validate against the requested address, or it is malformed. So we are strict about our read of these fields, and we never introspect the dropped error string.

crates/swarm/net/pushsync/src/codec.rs
mfw78: You are right, and I have reworked it. Failure receipts are not signed by the reference: a failure is a receipt with an error string and no signature, so there is nothing signed to model and the zeroed placeholders were not warranted.

`Receipt` is now success-only: real `address`, `signature`, `nonce`, and `storage_radius`, with the `failed` flag and the `failure()` placeholder constructor removed. The decoded response is a small local enum, `ReceiptResponse { Stored(Receipt), Failed }`, implementing `ProtoMessage`. Decode: a non-65-byte (typically empty) signature decodes to `Failed`; otherwise the success fields are parsed strictly into `Stored(Receipt)`. Encode: we never send a failure on the wire (the responder resets the stream instead), so the `Failed` arm just emits an empty default frame and is documented as round-trip-only; nothing reads it.

The ripple moved with it: the pushsync outbound `Output`, `ClientOutboundOutput::Pushsync`, and `on_pushsync_receipt` (which used to check `receipt.failed`) now match on `ReceiptResponse`. The handler takes the chunk address from the outbound request info, so the `Failed` arm no longer needs an address baked into the message.

crates/swarm/net/pushsync/src/protocol.rs
mfw78: Done. `MAX_MESSAGE_SIZE` is now computed from the chunk-size constants rather than a round 16 KiB, and it is tight.

The largest legitimate `data` is a single-owner chunk: `SPAN_SIZE` (8) + `HASH_SIZE` (32, the owner id) + a 65-byte recoverable signature + `DEFAULT_BODY_SIZE` (4096) = 4201 bytes. The pushsync `Delivery` frames that as a 32-byte `address` + `data` + `stamp` (`STAMP_SIZE` = 113). Adding a fixed 64-byte allowance for protobuf field tags, length varints, and the outer length-delimited frame prefix:

`HASH_SIZE + (SPAN_SIZE + HASH_SIZE + 65 + DEFAULT_BODY_SIZE) + STAMP_SIZE + 64 = 32 + 4201 + 113 + 64 = 4410` bytes.

The const is built from those named constants (`nectar_primitives::bmt::{SPAN_SIZE, HASH_SIZE, DEFAULT_BODY_SIZE}` and `nectar_postage::STAMP_SIZE`, both already direct deps of this crate) with the arithmetic in a comment, so it is exact and self-documenting and strictly below 16 KiB. The signature length (65) has no nectar export, so it is a local named const. Legitimate frames all fit: a content chunk has a smaller body, an encrypted chunk's per-wire body is still bounded by `DEFAULT_BODY_SIZE`, and a SOC is the worst case used above. A success receipt is far smaller.

crates/swarm/net/retrieval/src/protocol.rs
mfw78: Same treatment as the pushsync thread above, with one less field. The retrieval `Delivery` carries no address of its own, so the bound is just `data` + `stamp` + framing:

`(SPAN_SIZE + HASH_SIZE + 65 + DEFAULT_BODY_SIZE) + STAMP_SIZE + 64 = 4201 + 113 + 64 = 4378` bytes.

Computed from the same named nectar constants with the arithmetic in a comment, exact and strictly under 16 KiB. Content, SOC, and encrypted frames (body bounded by `DEFAULT_BODY_SIZE`) all fit, and an empty-delivery failure is trivially smaller.

Issue #132, opened 2026-06-06
retrieval: outbound self-throttle to stay under remote's pseudosettle allowance


mfw78:
## Context

Same shape as #131 (pushsync), applied to retrieval. Our outbound retrieval requests cost credit with the serving peer; pacing keeps us below the negotiated allowance for that peer and avoids wasted roundtrips when the peer refuses to serve.

## Proposed mechanism

- `SelfRateLimiter<PeerId>` from #130.
- Cost per request: typically a flat estimate (we do not know response size before the response arrives). Start with `ceil(MAX_CHUNK_SIZE / settle_unit_size)`; revisit if it proves too conservative once measurements are available.
- Quota driven by the same per-peer allowance signal the pushsync self-throttle subscribes to.

## Acceptance criteria

- [ ] Retrieval's outbound API blocks (or returns a delay) when self-throttled.
- [ ] Self-throttle quota tracks the negotiated allowance.
- [ ] Allowance updates resize the bucket promptly.
- [ ] Per-peer bucket cleared on disconnect.
- [ ] Metric `retrieval_self_throttled_total{peer_overlay}` increments on delay.
- [ ] Existing retrieval tests pass; new tests cover the allowance boundary.

## Depends on

- #130 (SelfRateLimiter primitive).

mfw78:
Related: #175 makes `NetworkChunkProvider::retrieve_chunk` peer selection accounting-aware (which peers to ask and whether to ask, based on per-peer balance and announced thresholds). This issue stays scoped to the outbound self-throttle (pacing requests against a single peer's allowance). The two compose: selection picks affordable peers, the self-throttle paces requests to each. Context: the typed download path landed in #170-#174.

mfw78:
See the note on #130: PeerAffordability (#213) plus the PeerSelector wiring (#219) provide the per-peer allowance signal this outbound retrieval throttle should pace against.

Issue #131, opened 2026-06-06
pushsync: outbound self-throttle to stay under remote's pseudosettle allowance


mfw78:
## Context

Outbound pushsync chunks consume our credit with the remote via pseudosettle / swap. If we push faster than the remote replenishes our allowance, the remote refuses to accept further chunks until we settle, and the failed pushes were a wasted roundtrip. Today there is nothing in pushsync that paces our own outbound rate against the remote's allowance for us.

## Proposed mechanism

- Use the `SelfRateLimiter<PeerId>` from #130.
- Cost per pushed chunk = ceil(chunk_size / settle_unit_size), where `settle_unit_size` is derived from the pseudosettle config (per-second forgiveness rate * one second).
- Bucket quota: the negotiated allowance for that peer, refilled by the pseudosettle layer when we observe a settlement / forgiveness event.
- Coupling with the accounting layer: the pseudosettle / swap behaviour publishes our current allowance per peer; pushsync's self-throttle subscribes (channel? shared `Arc<DashMap<PeerId, Allowance>>`?) and re-sizes the bucket on every change.

## Acceptance criteria

- [ ] Pushsync's outbound API blocks (or returns a delay) when self-throttled.
- [ ] Self-throttle quota is set from the negotiated allowance, not hard-coded.
- [ ] Allowance updates from the accounting layer resize the bucket promptly.
- [ ] On disconnect, the per-peer bucket is cleared.
- [ ] Metric `pushsync_self_throttled_total{peer_overlay}` is incremented when a push is delayed.
- [ ] Existing pushsync tests pass; new tests cover throttling at the boundary of the allowance.

## Depends on

- #130 (SelfRateLimiter primitive).

mfw78:
**Open question, must resolve before implementation: confirm the cost-per-chunk formula.**

The body of this issue proposes `cost = ceil(chunk_size / settle_unit_size)`, with `settle_unit_size` derived from the pseudosettle per-second forgiveness rate. That is a guess. Before any code lands, we need to confirm against the canonical accounting model:

- What unit does the accounting layer actually charge a pushed chunk in? Bytes? Fixed per-chunk? Stamp-batch units?
- Does the cost differ between a fresh push, a forwarded push, and a stored push?
- Is there a separate component for the proof-of-storage or just per-byte?
- Where is this defined in the Book of Swarm vs only in bee's source?

[...]

PR #298, opened 2026-06-13
feat(net-ratelimiter): add SelfRateLimiter for outbound accounting-gated protocols

### LINKED ISSUE #130: ratelimiter: add self-throttle pattern for outbound accounting-gated protocols

mfw78:
## What

Adds `SelfRateLimiter<K>` to `vertex-net-ratelimiter`, the outbound dual of the existing `KeyedRateLimiter`. Pushsync and retrieval initiate outbound work that consumes the credit a remote peer extends us through accounting; issuing faster than the remote replenishes our allowance wastes round trips on refusals. This primitive throttles our own send rate so we stay under that allowance.

The crate docs now spell out the inbound/outbound symmetry: inbound is `KeyedRateLimiter` (admit or refuse a remote's requests), outbound is `SelfRateLimiter` (park and delay our own requests until the bucket can satisfy them). Both wrap the same GCRA bucket.

## API

- `try_send(key, cost) -> Result<(), DelayUntil>`: charges the bucket without parking. `Ok(())` consumes `cost`; `Err(DelayUntil(d))` carries the GCRA wait hint (`d` is the earliest a retry would succeed; an oversized cost reports the full replenish window).
- `enqueue(key, cost, value) -> Result<Option<T>, T>`: admits immediately (`Ok(Some(value))`) when the bucket has room, otherwise parks the value on a per-key FIFO and arms a delay timer (`Ok(None)`). Hands the value back (`Err(value)`) when the cost can never fit the bucket. The cost is supplied per call, so a consuming protocol crate derives it from request size or kind.
- `poll_ready(cx) -> Poll<(K, T)>` and the `next_ready()` future: drain parked items in per-key enqueue order once their bucket has refilled. Each yielded item has already had its cost charged, so the caller issues it directly.
- `clear(&K)`: drops the bucket and any parked items on disconnect, mirroring `KeyedRateLimiter::clear`.

Also adds `KeyedRateLimiter::try_peek` (non-mutating admission check, used to read a key's wait hint when arming the delay timer) and `KeyedRateLimiter::replenish_all_every`.

## DelayQueue draining

Rather than `tokio_util::time::DelayQueue`, which reaches the tokio timer driver and is not wasm-safe, parking is built on the same `futures-timer` delay the rest of the wasm cone uses. Each key holds a FIFO plus one `futures_timer::Delay` armed to the moment its head item can be admitted. `poll_ready` polls every key's timer (registering the waker on the still-pending ones), and on the first elapsed timer re-checks the live bucket, pops the head, charges its cost, re-arms for the next head, and yields the item. A lost race re-arms and stays pending.

## Wasm-safety

The delay timer pulls `futures-timer` with the `wasm-bindgen` feature, which routes timers through the browser clock on `wasm32` instead of the std clock that panics there. Durations use `std::time::Duration` (no clock); the wrapped limiter already takes its monotonic clock from `web-time` via `vertex-util-runtime`. `tokio` is a dev-dependency only. Verified with `cargo build --target wasm32-unknown-unknown -p vertex-net-ratelimiter` (the crate is in the `vertex-swarm-node` wasm cone the CI `wasm` job builds).

## Tests

Accept under budget, reject plus delay over budget, oversized-cost reporting, per-key independence, enqueue admit/park/oversized, drain after replenish, FIFO drain order, pending when nothing parked, and `clear` releasing a key with parked items. Timing tests use real `futures-timer` delays bounded by a generous timeout rather than exact-tick assertions, so they are not flaky.

## Out of scope

The pushsync and retrieval outbound integration touches the node crate and is filed separately (#131, #132). This PR is the self-contained primitive only.

Closes #130.

### DIFF
--- a/crates/net/ratelimiter/src/self_limiter.rs
+++ b/crates/net/ratelimiter/src/self_limiter.rs
@@ -1,0 +1,413 @@
+//! Outbound self-throttle on top of a [`KeyedRateLimiter`].
+//!
+//! [`SelfRateLimiter`] is the outbound dual of [`KeyedRateLimiter`]: instead of
+//! admitting or refusing a remote peer's inbound requests, it throttles the
+//! requests we send to a remote peer over an accounting-gated protocol so we
+//! stay under the credit that peer extends us. A request the bucket cannot
+//! admit yet is parked on an internal delay queue and surfaced again once the
+//! bucket has refilled, rather than being issued and refused.
+//!
+//! [`KeyedRateLimiter`]: crate::KeyedRateLimiter
+
+use std::collections::HashMap;
+use std::collections::VecDeque;
+use std::collections::hash_map::Entry;
+use std::future::Future;
+use std::hash::Hash;
+use std::pin::Pin;
+use std::task::{Context, Poll};
+use std::time::Duration;
+
+use futures_timer::Delay;
+
+use crate::{KeyedRateLimiter, Quota, RateLimitedErr};
+
+/// The wait hint returned when a charge cannot be admitted yet.
+///
+/// Carries the [`Duration`] from the GCRA bucket after which a retry of the
+/// same charge would be admitted. The caller can either schedule its own retry
+/// or hand the request to [`SelfRateLimiter::enqueue`] to have the limiter park
+/// and drain it.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct DelayUntil(pub Duration);
+
+impl DelayUntil {
+    /// The wait hint after which the charge would be admitted.
+    pub fn duration(self) -> Duration {
+        self.0
+    }
+}
+
+/// One parked request: the value to surface and the cost to charge when it is.
+struct Item<T> {
+    cost: u32,
+    value: T,
+}
+
+/// A FIFO of requests parked behind one key's bucket, plus the timer that wakes
+/// the head. The timer is armed to the earliest moment the head item could be
+/// admitted and re-armed whenever the head changes.
+struct Parked<T> {
+    queue: VecDeque<Item<T>>,
+    timer: Delay,
+}
+
+/// Outbound self-rate-limiter keyed by `K` (typically a `PeerId`).
+///
+/// Wraps a [`KeyedRateLimiter`] and adds a per-key parking queue so requests
+/// the bucket cannot admit yet are delayed rather than refused.
+///
+/// # Driving the queue
+///
+/// Drive draining through [`Self::poll_ready`] (or the convenience
+/// [`Self::next_ready`] future): poll it to receive `(K, T)` pairs whose bucket has
+/// refilled, in per-key enqueue order. Each yielded item has already had its
+/// cost charged against the bucket, so the caller can issue it immediately.
+///
+/// # Cost function
+///
+/// The cost (token count) is supplied per call, so the consuming protocol crate
+/// derives it from request size or kind however it likes.
+///
+/// # Cleanup
+///
+/// Call [`Self::clear`] on peer disconnect to drop the bucket and any parked
+/// requests for that key, mirroring [`KeyedRateLimiter::clear`].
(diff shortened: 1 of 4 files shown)


crates/net/ratelimiter/src/self_limiter.rs
```diff
@@ -0,0 +1,413 @@
...
+impl DelayUntil {
+    /// The wait hint after which the charge would be admitted.
+    pub fn duration(self) -> Duration {
+        self.0
+    }
+}
+
+/// One parked request: the value to surface and the cost to charge when it is.
+struct Item<T> {
+    cost: u32,
```
mfw78, line 43: Should this actually be of type Au (accounting units)? Or even generic with Au being an implementation of one of the generics potentially?

mfw78: Keeping cost as a plain u32 token count here is deliberate.

crates/net/ratelimiter is protocol-agnostic net infrastructure (see crates/net/AGENTS.md: no nectar-primitives, nothing from crates/swarm/). Au lives in vertex-swarm-api (crates/swarm/api/src/accounting.rs, struct Au(i64)), which is the swarm domain layer. Typing cost as Au would make this net util depend on the swarm-domain accounting type, inverting the dependency direction and breaking the layering the area file mandates.

The two are also different domains semantically. Au is a signed accounting balance (it can go negative; it represents owed/owing). The GCRA cost is an abstract non-negative token count: exactly what the wrapped KeyedRateLimiter::try_consume_n(n: u32) and try_peek(key, n: u32) already speak. So u32 is the natural and consistent type, and any newtype here would just wrap-and-unwrap at every call into the limiter without buying clarity (unlike Quota or DelayUntil, which wrap genuinely distinct units).

The conversion from a remote's pseudosettle allowance (which is in Au) into rate-limiter tokens, e.g. ceil(chunk_size / settle_unit), belongs at the protocol integration layer where Au is in scope (pushsync/retrieval wiring, issues #131/#132), not in this primitive. Leaving it u32 keeps that conversion at the right boundary.

Going with cost: u32 unchanged.

Issue #293, opened 2026-06-13
pushsync: reject and penalise shallow custody receipts (validate receipt depth against storage radius)


mfw78:
## Problem

Pushsync returns a statement-of-custody receipt that should be signed by a node genuinely inside the chunk's neighbourhood: a storer `S` whose proximity order to the chunk address is at least the network storage radius / neighbourhood depth `d`. A receipt is "shallow" when its signer is not deep enough, `PO(signer, chunk) < d`. The chunk never reached the responsible neighbourhood; a node out in the shallow bins issued or relayed a receipt as if it had.

Three failures follow:

1. Silent data loss. Retrieval routes toward the deep neighbourhood of the address. A chunk that only landed on a shallow node is not found there, so a "successful" upload is unretrievable.
2. No redundancy, then eviction. The neighbourhood is meant to hold several replicas. A shallow node is a single point, and because the reserve is closed upwards for PO (nodes keep chunks close to their own address), a chunk far from a shallow node is the first thing it garbage-collects.
3. Incentive gaming. A forwarder can profit immediately by returning a receipt instead of pushing to the storer (Book of Swarm 3.5). The book's economic defence (no receipt in time -> uploader re-uploads via another route) only punishes missing receipts; a fabricated or relayed shallow receipt defeats the timeout defence because the uploader believes it succeeded and never retries. A protocol-level depth check is what closes this.

## Current state

The wire already carries the data but nothing validates it:

- `crates/swarm/net/pushsync/src/codec.rs` `Receipt` carries `storage_radius` (the storer's claim) but it is never checked.
- The receipt signature is never verified to recover to the storer's overlay (tracked under #287 receipt verification).
- No `PO(receipt_signer, chunk)` depth check exists anywhere.

## Proposal

On receiving a receipt, as the origin uploader and (critically) as a forwarder before relaying it back (#291):

- Verify the receipt signature recovers to the responding peer's overlay (over the chunk address and nonce). This part overlaps #287.
- Compute `PO(receipt_signer_overlay, chunk_address)` and require it to be at least the depth the chunk should have reached. Reject anything shallower so the uploader's retry-via-different-route dynamic actually engages, and report the peer adversely (`InvalidData` or a dedicated shallow-receipt event) through `PeerManager::report_peer`.
- A forwarder must run this check before relaying a receipt upstream, so it never launders a shallow receipt and takes the reputational and accounting hit itself.

## Design questions (why this is not a one-liner)

[...]

Issue #291, opened 2026-06-13
retrieval/pushsync: implement forwarding (multi-hop relay) with verification and two-leg accounting


mfw78:
## Problem

Retrieval and pushsync are currently **origin-only**: a node issues requests and answers requests, but never forwards. There is no relay logic anywhere — a retrieval request we cannot serve is not forwarded to a closer peer, and a pushsync delivery we are not responsible for is not relayed onward toward the chunk's neighbourhood. Forwarding Kademlia is the core routing primitive of Swarm; without it vertex can only operate at the edge of the network and cannot participate as a useful client/storer hop.

Current state:

- `crates/swarm/node/src/swarm_client.rs` — origin-side candidate walk only (3 closest peers, sequential).
- `crates/swarm/node/src/client_service.rs` — `ChunkRequested` / `ChunkPushReceived` events have TODO stubs; neither path can recurse to the next-closer peer.
- Inbound responders are held per request (`request_id`-keyed) so the plumbing to answer late is present; what is missing is the upstream leg and joining it to the held responder.

## Scope

1. **Retrieval forwarding:** on `ChunkRequested` miss (not in LocalStore / not a storer), select next-closer peers (excluding the requester and ourselves-already-tried), issue an upstream retrieval, verify the returned chunk, then answer the held responder. Loop prevention (don't forward backwards in proximity order), TTL/hop bound, per-request deadline.
2. **Pushsync relay:** on `ChunkPushReceived` when not within our radius of responsibility, relay to closer peers, await the storer receipt, verify it, pass it back to the pusher. If we *are* responsible: store (storer gate) and sign our own receipt.
3. **Accounting on both legs:** debit upstream peer / credit downstream peer per pricing, using the existing prepare/apply actions — a forwarder earns the spread exactly as bee does.
4. **Verification at both edges** per #287: never relay an unverified chunk or receipt.
5. Depends on #178 (correlation ids) so concurrent forwarded requests for the same chunk don't collide, and on the cache (#22 adjacency: serving forwarded chunks from cache is the natural optimization, can be follow-up).

## Acceptance

- [ ] Three-node chain test: A retrieves through B from C; B verifies, accounts both legs, and relays.
- [ ] Push chain test: A pushes through B to storer C; receipt flows back verified.
- [ ] Loop/hop bounds enforced; malformed upstream data is scored (#287) and not relayed.

PR #299, opened 2026-06-13
feat(swarm-node): implement client-service serve and store-and-receipt handlers

mfw78:
## What

Implements the two stubbed `ClientService` event handlers in `crates/swarm/node/src/client_service.rs` behind a `LocalStore` serving seam: `ChunkRequested` (serve a retrieval) and `ChunkPushReceived` (accept a push and return a receipt). Refs #76.

## The LocalStore seam

`ClientService` gains an optional `Arc<dyn SwarmServing>`, wired via a `with_local_store` builder that mirrors `with_reporter`. `SwarmServing` is an object-safe node-crate trait; the concrete `LocalServing` captures the node identity and an `Arc<dyn SwarmLocalStore>`. A client node leaves the seam unset, a storer node wires it.

The `SwarmLocalStore` trait (in `crates/swarm/api/src/components/localstore.rs`) keeps its four original methods and gains two defaulted ones, `store_stamped` and `retrieve_stamped`. Serving a retrieval over the wire requires the stamp that authorized the chunk, but the base `store(&AnyChunk)` drops it, so a serving backend needs to persist and return the stamp. The defaults fall back to the stamp-less path so non-serving backends are unaffected; the storer's `LocalStoreImpl` overrides both and persists the stamp as a fixed-length trailer on the stored chunk bytes (a leading flag byte distinguishes stamped from unstamped entries).

## Storer vs client behaviour

A storer node serves a local-store hit through the existing `ServeChunk` command, which still passes the handler's verify-before-serve gate from #294 (the served chunk's address must answer the request). It takes custody of pushed chunks it is responsible for and replies with `SendReceipt`. A client node has no store: every inbound serve and push is failed cleanly by resetting the request substream.

Resetting an inbound request promptly needed an explicit path, so two local control commands were added: `FailRetrieval` and `FailPush`. They reuse the responders' existing `send_error` reset (the same failure signal the reference reads), so no new wire bytes are introduced and no fork gate is required.

## Receipt signing scheme matched against the reference

The reference storer signs `chunk.Address().Bytes()` (the raw 32-byte chunk address) with the node signing key under the EIP-191 personal-message prefix (`\x19Ethereum Signed Message:\n32` plus the address bytes), and echoes the node nonce and storage radius in the receipt. A requester recovers the signer from the address bytes and derives the storer overlay to check closeness. `LocalServing::sign_receipt` replicates this exactly: it calls `sign_message_sync(address.as_slice())` (alloy's EIP-191 personal sign) and returns the node `nonce` and storage radius. A test recovers the signer from the signature over the address bytes and asserts it matches the node identity, so the receipt verifies on a reference peer.

## Stamp validation and responsibility

Stamp validation does the structural/signature check: `Stamp::recover_signer(&address)` must recover a signer over the stamp digest. The full check (recovered signer is the on-chain owner of an existing, funded batch with sufficient depth and a non-exhausted bucket) is deferred with a TODO because that batch state is not available until the postage/storer batch store is wired; no batch state is invented.

Responsibility uses `proximity(node_overlay, address) >= storage_radius`, mirroring the reference rule. The storage radius source (the storer reserve / `SwarmStorer` radius) is not yet wired, so `LocalServing` takes a conservative fixed radius (`StorageRadius::ZERO`, responsible for every pushed address) with a documented TODO to drive it from the reserve.

## Builder wiring

The builder launch path builds a `LocalStoreImpl` for storer nodes (backed by an in-memory database for now; sharing the on-disk node database for chunk persistence is deferred because redb is single-writer and that file is already opened for the peer snapshot store) and injects it as the serving seam. Client nodes pass `None`. The `StorerNodeBuilder` scaffold in `node/storer.rs` gains the same `with_local_store` wiring. Pullsync and redistribution are not pulled in.

## Deferred

- Forwarding a local-store miss to a closer peer, and relaying a not-responsible push onward: #291.
- Pullsync, chunk-sync, redistribution: #77 / #75.
- Requester-side shallow-receipt verification: #293.
- Full on-chain batch-existence/owner/value stamp validation: tracked with a TODO referencing #76.
- A live storage radius source and on-disk chunk persistence sharing the node database: tracked with TODOs referencing #76.
- `SettlementNeeded` settlement initiation: left as-is.

## Tests

- A storer serves a chunk it has stored (emits `ServeChunk`).
- A storer accepts a push, stores it, and returns a receipt whose signature recovers the node signer over the chunk address (wire-compatible with a reference peer).
- A client node with no store fails a serve (`FailRetrieval`) and a push (`FailPush`) cleanly.
- The storer `LocalStoreImpl` round-trips a stamp through `store_stamped`/`retrieve_stamped`, and an unstamped store has no stamp to serve.

`cargo fmt`, `cargo clippy --all-targets --all-features -D warnings`, and the unit tests pass on every touched crate (api, storer, node, builder).


mfw78:
## Closing: this implements #75 (storer) inside #76 (client)

A four-lens architectural review (libp2p, Swarm/wire-compat, Rust idiom, native/FFI/WASM) reached a unanimous verdict: this PR builds storer behaviour in the client path, resurrects the request-id rendezvous map that #292 deleted, and smuggles the postage stamp through a fixed-length byte trailer. It is superseded by the re-scoped #76 (cache-only forwarding client). Closing rather than patching.

### Core problems
1. Storer logic in the client. The client mints a custody receipt signed by its own overlay (`serving.rs:128-148`, called from `client_service.rs:354-371`), which forges a storer identity and asserts custody it cannot hold. Plus store-as-authority, a responsibility radius stubbed to `StorageRadius::ZERO` (claims every address), and stamp-validation-as-authorization. All of this is storer scope (#75). A forwarding client relays the downstream storer's receipt verbatim and signs nothing.
2. Request-id resurrection. Inbound serving reintroduces `pending_responses` keyed by a synthetic `u64`, with `ServeChunk`/`SendReceipt`/`FailRetrieval`/`FailPush` commands bouncing service to behaviour to handler to a map lookup, even though the responder owns its substream and never needs to leave the handler. This is exactly what #292 deleted on the outbound side.
3. Stamp-trailer hack. `store_stamped`/`retrieve_stamped` are added as defaulted `SwarmLocalStore` methods whose defaults are semantically wrong (drop the stamp / fake a miss), and a `[flag][type][data][stamp(113)]` trailer (`storer/src/store.rs`) smuggles the stamp through an `AnyChunk`-valued store. Root cause: the store's value type drops the stamp.
4. Wrong storage primitive. The client path wires `None` (so a client has no storage at all); the storer path wires `LocalStoreImpl` over an in-memory redb with a `NoEviction` reserve bounded by chunk count, native-only. A client cache must be a byte-bounded, lossy, in-memory LRU holding `StampedChunk`, in the wasm cone.

### Worth salvaging (carried into the rework)
Reset-on-failure as the terminal signal; typed `FailureKind` peer scoring and inbound malformed-chunk attribution; decode-time verification; the `StampedChunk` primitive (it becomes the cache currency); and `vertex-swarm-storer` itself, unchanged, as the #75 storer reserve.

PR #302, opened 2026-06-13
feat(chain): swap price oracle indexer

mfw78:
## Summary

Adds `vertex-chain-index-swap-price`, the first per-contract indexer built on the generic, reorg-safe `vertex-chain-index` engine (#300). It indexes the swap (settlement) price oracle on Gnosis Chain (`0xA57A50a831B31c904A770edBCb706E03afCdbd94`, chain id 100) and folds its events into a `vertex-storage` projection a consumer reads lazily.

Part of the chain-indexing initiative: the engine owns reorg/cursor/paging once, and each contract ships a thin `Indexer` that declares what to watch and how to fold one log.

## Events indexed

- `PriceUpdate(uint256 price)` sets the swap exchange rate.
- `ChequeValueDeductionUpdate(uint256 chequeValueDeduction)` sets the cheque value deduction.

The event bindings (`ISwapPriceOracle`) and the deployment constants (address and deployment block, for both Gnosis mainnet and testnet) come from `nectar_contracts`, so the address book is not duplicated here. `OwnershipTransferred` and any other topic are ignored.

## Projection table

`SwapPriceTable` (`"swap_price_oracle"`), a two-row table keyed by `SwapPriceField`:

| Key | Value |
|---|---|
| `ExchangeRate` | `SwapPriceRow { value: U256, source: LogPosition }` |
| `ChequeValueDeduction` | `SwapPriceRow { value: U256, source: LogPosition }` |

Each row records the value and the `(block, log_index)` (`LogPosition`) of the log that set it. Read the current values with `read_field(db, field)`.

## Pure, no-reaction fold

`SwapPriceIndexer::apply` is a pure, idempotent projection step per `CHAIN_REACTIONS_DESIGN.md`:

- It writes **only** the projection table. No side effects, no reactions, no calls into accounting, the storer, or any other domain. The chain crate stays domain-agnostic.
- A replayed finalized log re-applies to the same row with the same value. The stored `source` position guards the write (`apply_update` overwrites only when the incoming log is strictly newer), so a reordered or replayed log never rolls a row back: the fold is idempotent and monotonic.
- Reacting to a rate change (for example re-pricing a cheque) is the consumer's job, done lazily by reading the projection at its own decision point. No `on_event` hook is added.

Placement matches the engine: native, RPC-and-persistence code, `alloy` with `default-features = false`, no transport of its own, kept out of the wasm cone (not a dependency of any wasm-bound crate).

## Tests

- **Unit tests over synthetic logs** (in-memory `vertex-storage` backend): ABI-encode each event into a real `Log`, fold it, and assert the projected value and source. Covers both events projecting independently, idempotent replay (re-applying the same log is a no-op), monotonic supersede (a stale reordered log does not overwrite a newer value; same-block higher `log_index` wins), unrelated-topic ignoring, and topic-filter selection.
- **Env-guarded `#[ignore]`d anvil-fork e2e** on a unique port (8547): syncs a bounded recent window against a real Gnosis fork, asserts the cursor advanced to the finalized head, and cross-checks the projection against a direct `getPrice()` `eth_call` at the head.

### e2e result

Ran against an anvil fork of Gnosis at a recent block (`anvil --fork-url https://rpc.gnosischain.com --port 8547`). The engine backfilled cleanly and the cursor advanced to the finalized head.

This particular swap oracle is a real edge case: it sets its price and deduction in its constructor and, across its entire on-chain lifetime, has emitted **zero** `PriceUpdate` / `ChequeValueDeductionUpdate` logs (a full-lifetime `eth_getLogs` over the two topics returns nothing; its only log ever is the deployment `OwnershipTransferred`). So the e2e cross-check is two-sided: when the window holds an update the projection must equal `getPrice()`, and when it holds none (the live case) the projection stays empty while the on-chain `getPrice()` still reports a value (`price 100000, deduction 100`), proving the indexer is wired to the right live contract and the engine reached head over real data. The synthetic-log unit tests exercise the actual event-capture and fold path that the live contract has not yet triggered.

## Verification

- `cargo fmt --all` clean.
- `cargo clippy --all-targets --all-features -- -D warnings` clean.
- `cargo test --all-features` on the crate: 8 unit tests pass; the e2e is `#[ignore]`d.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8364,6 +8364,23 @@ dependencies = [
 ]
 
 [[package]]
+name = "vertex-chain-index-swap-price"
+version = "0.1.0"
+dependencies = [
+ "alloy-primitives",
+ "alloy-provider",
+ "alloy-rpc-types-eth",
+ "alloy-sol-types",
+ "nectar-contracts",
+ "serde",
+ "thiserror 2.0.18",
+ "tokio",
+ "vertex-chain-index",
+ "vertex-storage",
+ "vertex-storage-redb",
+]
+
+[[package]]
 name = "vertex-ffi"
 version = "0.1.0"
 dependencies = [
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -98,6 +98,7 @@ members = [
     # Chain (Ethereum RPC access)
     "crates/chain",
     "crates/chain/index",
+    "crates/chain/index-swap-price",
 
     # FFI (native embedding surface)
     "crates/ffi",
@@ -238,6 +239,7 @@ vertex-rpc-server = { path = "crates/rpc/server" }
 # chain (Ethereum RPC access)
 vertex-chain = { path = "crates/chain" }
 vertex-chain-index = { path = "crates/chain/index" }
+vertex-chain-index-swap-price = { path = "crates/chain/index-swap-price" }
 
 # ffi (native embedding surface)
 vertex-ffi = { path = "crates/ffi" }
--- a/crates/chain/index-swap-price/Cargo.toml
+++ b/crates/chain/index-swap-price/Cargo.toml
@@ -1,0 +1,46 @@
+[package]
+name = "vertex-chain-index-swap-price"
+version.workspace = true
+edition.workspace = true
+rust-version.workspace = true
+license.workspace = true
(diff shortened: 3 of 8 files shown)


mfw78:

PR #303, opened 2026-06-13
feat(chain): redistribution game indexer

mfw78:
## Redistribution game indexer

Adds `vertex-chain-index-redistribution`, a per-contract indexer for the Swarm storage-incentive Redistribution game on Gnosis Chain (id 100), built on the generic reorg-safe `vertex-chain-index` engine. Part of the chain-indexing initiative (#300).

Contract: `0x5069cdfB3D9E56d23B1cAeE83CE6109A7E4fd62d`, deployment block `41105199`. ABI: `storage-incentives/deployments/mainnet/Redistribution.json`.

### Events indexed

`Committed`, `Revealed`, `CurrentRevealAnchor`, `TruthSelected`, `WinnerSelected`, `CountCommits`, `CountReveals`, `ChunkCount`. The `sol!` bindings are derived from the contract ABI.

### Projection table

Two `vertex-storage` tables, both written by the fold:

- `RoundTable` (`redistribution_rounds`), keyed by the on-chain `roundNumber`: the headline per-round game log. Folds the round-carrying events (`Committed`, `Revealed`, `CurrentRevealAnchor`) into a `RoundState` with last-write-wins per field. The commit and reveal sets are keyed by source log position and upserted in place, so a replay overwrites rather than appending.
- `RoundEventTable` (`redistribution_events`), keyed by `(block_number, log_index)`: the raw event log. Every selected event lands here verbatim, including the round-terminal ones (`TruthSelected`, `WinnerSelected`, `ChunkCount`, `CountCommits`, `CountReveals`) whose on-chain payload does not name a round. The log position is the natural idempotency key.

### `apply` is a pure, no-reaction fold

Per `CHAIN_REACTIONS_DESIGN.md`, the redistribution game is a clock, not a reaction source. This indexer only RECORDS. `Indexer::apply` is a pure, idempotent fold into the projection: it writes rows and returns. It calls into no storer, accounting, or redistribution agent; it fires no reaction; it adds no hook; it gates no node behaviour. A replayed finalized log re-writes the same row to the same value. Any node logic that needs the game state reads the projection lazily at its own decision point.

### Tests

- Unit tests over synthetic logs (in-memory `vertex-storage` backend): each event decodes and folds into the projection correctly, and re-applying a log is idempotent (a replay never double-counts a commit/reveal; distinct logs in a round accumulate). The synthetic logs are ABI-encoded through the same `sol!` bindings the indexer decodes with. 7 tests, all green.
- An env-guarded, `#[ignore]`d anvil-fork e2e on a unique port (8550): syncs a bounded recent window of the live contract and asserts a non-zero indexed event count, that the cursor advanced to the finalized head, and that a recorded `CurrentRevealAnchor` decodes to a non-zero anchor (cross-checking the fold against real chain data). Window is overridable with `REDIST_E2E_WINDOW`.

### e2e result

The e2e was run against an `anvil --fork-url <gnosis> --port 8550` fork. The indexer connected, derived the correct finalized window from the fork head, and began paging real Redistribution logs (`syncing Redistribution logs over blocks 46672936..=46674936 via http://localhost:8550`). In this environment the fork proxied each `eth_getLogs` page to a rate-limited shared public RPC (several indexer forks ran concurrently), so the bounded sync did not finish inside the 60s test budget; the indexer logic itself is exercised by the synthetic-log unit tests and the e2e connection/window derivation. The e2e stays `#[ignore]`d so CI without a fork stays green; run instructions are in the test module docs.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8364,6 +8364,21 @@ dependencies = [
 ]
 
 [[package]]
+name = "vertex-chain-index-redistribution"
+version = "0.1.0"
+dependencies = [
+ "alloy-primitives",
+ "alloy-provider",
+ "alloy-rpc-types-eth",
+ "alloy-sol-types",
+ "serde",
+ "tokio",
+ "vertex-chain-index",
+ "vertex-storage",
+ "vertex-storage-redb",
+]
+
+[[package]]
 name = "vertex-ffi"
 version = "0.1.0"
 dependencies = [
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -98,6 +98,7 @@ members = [
     # Chain (Ethereum RPC access)
     "crates/chain",
     "crates/chain/index",
+    "crates/chain/index-redistribution",
 
     # FFI (native embedding surface)
     "crates/ffi",
@@ -238,6 +239,7 @@ vertex-rpc-server = { path = "crates/rpc/server" }
 # chain (Ethereum RPC access)
 vertex-chain = { path = "crates/chain" }
 vertex-chain-index = { path = "crates/chain/index" }
+vertex-chain-index-redistribution = { path = "crates/chain/index-redistribution" }
 
 # ffi (native embedding surface)
 vertex-ffi = { path = "crates/ffi" }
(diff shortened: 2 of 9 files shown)


mfw78:

PR #305, opened 2026-06-13
feat(chain): chequebook factory indexer

mfw78:
## What

A new sibling indexer crate `crates/chain/index-chequebook-factory`
(`vertex-chain-index-chequebook-factory`) that plugs into the generic reorg-safe
`vertex-chain-index` engine (#300) and folds the SimpleSwap chequebook factory's
deployment event into a `vertex-storage` projection. It mirrors the
`index-swap-price` (#302) and `index-redistribution` (#303) layout.

## Events indexed

The SimpleSwapFactory (Gnosis mainnet `0xc2d5a532cf69aa9a1378737d8ccdef884b6e7420`,
deployment block 39939970; Sepolia testnet `0x0fF044F6bB4F684a5A149B46D7eC03ea659F98A1`)
emits one event this indexer watches:

- `SimpleSwapDeployed(address contractAddress)` names a newly factory-deployed
  ERC20SimpleSwap chequebook.

The event binding and the deployment constants come from
`nectar_contracts::IChequebookFactory` and `nectar_contracts::{mainnet,testnet}::CHEQUEBOOK_FACTORY`,
so the address book is not duplicated.

## Projection

A single `vertex-storage` table, `ChequebookFactoryTable`
(`"chequebook_factory_deployed"`): the **set of factory-deployed chequebook
addresses**, keyed by the deployed chequebook `Address`. Each row records the
`(block, log_index)` of the `SimpleSwapDeployed` log that created it.

The purpose is cheque validation: when a cheque arrives, its drawer chequebook
must be a genuine factory-deployed ERC20SimpleSwap. A consumer validates it
lazily with a single membership read, `is_factory_deployed(db, chequebook)`, at
its own decision point. The indexer never pushes a reaction.

## `apply` is a pure, idempotent, no-reaction fold

Per `CHAIN_REACTIONS_DESIGN.md`, `ChequebookFactoryIndexer::apply` only writes to
the projection table. It calls into no storer, accounting, or other domain crate;
it fires no reaction; it gates no node behaviour. The row records the source
`(block, log_index)` and a fold only overwrites when the incoming log is strictly
newer, so a replayed finalized log re-applies to the same row (idempotent) and a
reordered/stale log never regresses the set (monotonic supersede by
`(block, log_index)`). The chain crate stays domain-agnostic; reactions are the
consumer's job.

## Tests

- **Unit** (synthetic ABI-encoded logs, in-memory `vertex-storage` backend):
  the event decodes and folds into the set; distinct deployments accumulate;
  re-applying a log is idempotent; a stale/reordered log does not regress the
  source; same-block higher `log_index` supersedes; an unrelated topic is ignored;
  the filter selects the contract and the `SimpleSwapDeployed` topic.
- **e2e** (`tests/e2e_fork.rs`, env-guarded and `#[ignore]`d): syncs the live
  Gnosis mainnet factory over a bounded recent window via an anvil fork (port
  **8548**) or a public RPC, asserts the engine reaches the finalized head and the
  cursor advances, then cross-checks one recorded chequebook against a direct
  `deployedContracts(address)` call on the factory (which returns `true` for
  exactly the addresses it deployed). Run instructions are in the test's module
  docs.

## Placement

Native, RPC-and-persistence code intended to run behind a `chain` feature in its
consumers and to stay out of the wasm cone, like the engine it builds on.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8364,6 +8364,22 @@ dependencies = [
 ]
 
 [[package]]
+name = "vertex-chain-index-chequebook-factory"
+version = "0.1.0"
+dependencies = [
+ "alloy-primitives",
+ "alloy-provider",
+ "alloy-rpc-types-eth",
+ "alloy-sol-types",
+ "nectar-contracts",
+ "serde",
+ "tokio",
+ "vertex-chain-index",
+ "vertex-storage",
+ "vertex-storage-redb",
+]
+
+[[package]]
 name = "vertex-ffi"
 version = "0.1.0"
 dependencies = [
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -98,6 +98,7 @@ members = [
     # Chain (Ethereum RPC access)
     "crates/chain",
     "crates/chain/index",
+    "crates/chain/index-chequebook-factory",
 
     # FFI (native embedding surface)
     "crates/ffi",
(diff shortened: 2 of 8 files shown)


mfw78:

PR #308, opened 2026-06-13
feat(chain): postage batch + price indexer

mfw78:
## Summary

Adds `vertex-chain-index-postage`, a per-contract indexer for the PostageStamp
contract on Gnosis Chain (`0x45a1502382541Cd610CC9068e88727426b696293`,
deployment block `31305656`). It plugs into the generic `vertex-chain-index`
engine merged in #300 and follows the layout of the sibling indexers
(`index-swap-price` #302, `index-redistribution` #303): `sol!` event bindings, a
`vertex-storage` projection, and an `impl Indexer`.

## Events indexed

Five events from the contract ABI:

- `BatchCreated(batchId, totalAmount, normalisedBalance, owner, depth, bucketDepth, immutableFlag)`
- `BatchTopUp(batchId, topupAmount, normalisedBalance)`
- `BatchDepthIncrease(batchId, newDepth, normalisedBalance)`
- `PriceUpdate(price)`
- `Paused(account)`

The PostageStamp event lifecycle is not present in the upstream
`nectar_contracts::IPostageStamp` interface (which carries the call/view bindings
only), so the event bindings are derived from the ABI JSON in a local `sol!`
block, matching how `index-redistribution` derives its events.

## Projection tables

Two `vertex-storage` tables, both written by a pure, idempotent fold:

- `postage_batches` (`BatchTable`): the **batch set**, keyed by `batchId`. Each
  row holds owner, depth, bucketDepth, normalisedBalance, immutable, the
  creation `start_block`, and the source `(block, log_index)` that last set it.
- `postage_chain_state` (`ChainStateTable`): a **single-row** projection of the
  contract's pricing chain-state, holding `totalOutPayment`, `lastPrice`
  (`currentPrice`), `lastUpdatedBlock`, and `paused`. The running
  `totalOutPayment` accumulator is reconstructed from the `PriceUpdate` cadence
  exactly as the contract maintains it in `setPrice`: on each price change, if
  the prior price was non-zero, the elapsed cost is settled
  (`totalOutPayment += lastPrice * (block - lastUpdatedBlock)`) before adopting
  the new price at the new block.

Both folds are guarded by the source log position, so a replayed or reordered
finalized log re-applies to the same row (idempotent) and never rolls a row back
(monotonic by `(block, log_index)`).

## Read helper: "is batch valid now"

`is_batch_valid_now(db, batch_id, block)` answers
`normalisedBalance > currentTotalOutPayment(block)`, where
`currentTotalOutPayment(block) = totalOutPayment + lastPrice * (block - lastUpdatedBlock)`.
It is a pure read against the projection plus the live block clock, returning
`None` while either the batch or the chain-state is not yet indexed.

## `apply` is a pure, no-reaction fold (per CHAIN_REACTIONS_DESIGN.md)

`PostageIndexer::apply` writes only the two projection tables and returns. It
calls into no storer, accounting, or postage-issuer code, fires no reaction, and
**implements no eviction**. Batch expiry has no event to hook (a batch dies when
the rising `currentTotalOutPayment` line crosses its static `normalisedBalance`,
with no transaction at the crossing), so validity is left to a consumer that
recomputes it lazily on the block clock at its own decision point via the read
helper above. The chain crate stays domain-agnostic.

The crate is native-only behind the `chain` feature in its consumers (alloy with
`default-features = false`, no reqwest/TLS of its own), staying out of the wasm
cone like the engine it builds on.

## Tests

- **Unit tests** (`cargo test -p vertex-chain-index-postage`, 12 passing) over
  synthetic ABI-encoded logs against the in-memory `vertex-storage` backend:
  each event decodes and folds correctly; the price-update accumulator matches
  the contract arithmetic; the validity query tracks the rising line across the
  expiry crossing; re-applying a log (create and price-update) is idempotent;
  supersede is monotonic by `(block, log_index)`; an out-of-order topup never
  rolls the row back; a topup before a create fabricates no partial row.
- **Env-guarded `#[ignore]`d anvil-fork e2e** (`tests/e2e_fork.rs`): syncs a
  bounded recent window of the live contract, asserts the cursor advances to the
  finalized head, and cross-checks the reconstructed chain-state against direct
  `currentTotalOutPayment()`, `lastPrice()`, and `lastUpdatedBlock()` contract
  reads at the head. The cross-check is gated on the window reaching back past
  the on-chain `lastUpdatedBlock()` so it only asserts the strict accumulator
  match when the indexer actually saw the `PriceUpdate` that set the anchor.

  E2E RESULT: the test compiles and links, and against a local Gnosis fork
  (anvil `--fork-block-number 46545000 --port 8546`) the engine connected,
  resolved the fork's finalized head, and began syncing the live PostageStamp
  contract's `eth_getLogs` over the bounded window. It did not finish within the
  run window because the public Gnosis RPC the fork proxies is shared and
  heavily rate-limited (a single `eth_getLogs` page over this busy contract was
  throttled past several minutes), which is the documented shared-RPC failure
  mode. The test is therefore left `#[ignore]`d with full run instructions in the
  module docs; run it against a private or less-throttled Gnosis RPC (or a fork
  of one) to exercise the on-chain `currentTotalOutPayment()` /  `lastPrice()` /
  `lastUpdatedBlock()` cross-check end to end.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8364,6 +8364,22 @@ dependencies = [
 ]
 
 [[package]]
+name = "vertex-chain-index-postage"
+version = "0.1.0"
+dependencies = [
+ "alloy-contract",
+ "alloy-primitives",
+ "alloy-provider",
+ "alloy-rpc-types-eth",
+ "alloy-sol-types",
+ "serde",
+ "tokio",
+ "vertex-chain-index",
+ "vertex-storage",
+ "vertex-storage-redb",
+]
+
+[[package]]
 name = "vertex-ffi"
 version = "0.1.0"
 dependencies = [
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -98,6 +98,7 @@ members = [
     # Chain (Ethereum RPC access)
     "crates/chain",
     "crates/chain/index",
+    "crates/chain/index-postage",
 
     # FFI (native embedding surface)
     "crates/ffi",
@@ -238,6 +239,7 @@ vertex-rpc-server = { path = "crates/rpc/server" }
 # chain (Ethereum RPC access)
 vertex-chain = { path = "crates/chain" }
 vertex-chain-index = { path = "crates/chain/index" }
+vertex-chain-index-postage = { path = "crates/chain/index-postage" }
 
 # ffi (native embedding surface)
 vertex-ffi = { path = "crates/ffi" }
--- a/crates/chain/index-postage/Cargo.toml
+++ b/crates/chain/index-postage/Cargo.toml
@@ -1,0 +1,36 @@
+[package]
+name = "vertex-chain-index-postage"
+version.workspace = true
+edition.workspace = true
+rust-version.workspace = true
+license.workspace = true
+homepage.workspace = true
+repository.workspace = true
+description = "Postage stamp contract event indexer: a pure, idempotent projection of the batch set, the swap price, and the contract chain-state"
+
+[lints]
+workspace = true
+
+[dependencies]
+# The generic engine and its `Indexer` trait this crate plugs into.
+vertex-chain-index = { workspace = true }
+# The projection tables and the persisted-fold transaction surface.
+vertex-storage = { workspace = true }
+
+# alloy: `sol!` event bindings and the log filter. `default-features = false`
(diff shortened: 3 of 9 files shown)


mfw78:

PR #309, opened 2026-06-13
feat(chain): staking registry indexer

mfw78:
## What

Adds `vertex-chain-index-staking`, a per-contract indexer for the Swarm `StakeRegistry` (Gnosis Chain, `0xda2a16EE889E7F04980A8d597b48c8D51B9518F4`, deployed at block 40430237), built on the generic `vertex-chain-index` engine merged in #300.

## Events indexed

The indexer's filter selects five `StakeRegistry` topics at the registry address:

- `StakeUpdated(owner, committedStake, potentialStake, overlay, lastUpdatedBlock, height)`
- `StakeFrozen(frozen, overlay, time)`
- `StakeSlashed(slashed, overlay, amount)`
- `StakeWithdrawn(node, amount)`
- `OverlayChanged(owner, overlay)`

It reuses the shared `nectar_contracts::IStakeRegistry` `sol!` event bindings for the first four; `OverlayChanged` is the one event the shared interface does not carry, so it is declared locally from the deployment ABI.

## Projection

Two `vertex-storage` tables hold the folded state:

- `StakeTable` (`staking_stakes`): owner `Address` to an `OwnerStake` row carrying the committed and potential stake legs, the current overlay and height, the on-chain last-updated block, and the freeze deadline.
- `OverlayOwnerTable` (`staking_overlay_owners`): a staked `overlay` to the owner that staked it. This is the staked-overlay set, kept queryable by overlay. The fold reconciles it against the owner row's staked status: a slash or withdrawal that zeroes the stake removes the overlay from the set, and an `OverlayChanged` re-points the entry.

Reads go through `StakeProjection` (`stake_of`, `is_overlay_staked`, `owner_of_overlay`, `staked_overlays`).

## `apply` is a pure, no-reaction fold

Per `CHAIN_REACTIONS_DESIGN.md`, `apply` is a pure, idempotent fold into the storage projection: no side effects, no domain logic, no calls into a storer or accounting layer. Consumers read the projection lazily at their own decision point (for example, checking `is_frozen_at` right before committing a redistribution round); the chain crate never pushes a stake event at them.

Idempotency and monotonicity come from a `(block, log_index)` supersede rule: a fold step updates an owner's row only when the log's position is strictly later than the position last recorded for that owner. Re-delivering a finalized range on restart re-applies to the same row, and an out-of-order log never regresses state.

## Tests

- Unit tests over synthetic ABI-encoded logs: each of the five events decodes and folds correctly; replaying a finalized log is idempotent; the supersede is monotonic by `(block, log_index)`; slash and withdrawal drop the overlay from the staked set; the filter targets the registry and exactly the five signatures. They run against the in-memory `vertex-storage` backend.
- An env-guarded, `#[ignore]`d anvil-fork e2e (`STAKING_E2E_RPC`, default port 8549) syncs a bounded recent window of finalized blocks and cross-checks one indexed owner's overlay against a direct `stakes(owner)` contract call. Run instructions are in the test's module docs.

## Placement

Native, RPC-and-persistence code behind the engine it builds on; it decodes logs the engine already fetched and persists through the `vertex-storage` trait surface, opening no socket of its own, so it stays out of the wasm cone.

## E2E result

The anvil-fork e2e was run against a Gnosis fork (`anvil --fork-url https://rpc.gnosischain.com --fork-block-number 46670000 --port 8549 --silent`), scanning the bounded window `46629936..=46669936` below the fork's finalized head. The public RPC is shared and rate-limited, so the sweep is slow but bounded; result confirmation is appended below once the run completes. To reproduce, follow the run instructions in `crates/chain/index-staking/tests/e2e_fork.rs`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8364,6 +8364,23 @@ dependencies = [
 ]
 
 [[package]]
+name = "vertex-chain-index-staking"
+version = "0.1.0"
+dependencies = [
+ "alloy-primitives",
+ "alloy-provider",
+ "alloy-rpc-types-eth",
+ "alloy-sol-types",
+ "nectar-contracts",
+ "serde",
+ "tokio",
+ "tracing",
+ "vertex-chain-index",
+ "vertex-storage",
+ "vertex-storage-redb",
+]
+
+[[package]]
 name = "vertex-ffi"
 version = "0.1.0"
 dependencies = [
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -98,6 +98,7 @@ members = [
     # Chain (Ethereum RPC access)
     "crates/chain",
     "crates/chain/index",
+    "crates/chain/index-staking",
 
     # FFI (native embedding surface)
     "crates/ffi",
@@ -238,6 +239,7 @@ vertex-rpc-server = { path = "crates/rpc/server" }
 # chain (Ethereum RPC access)
 vertex-chain = { path = "crates/chain" }
 vertex-chain-index = { path = "crates/chain/index" }
+vertex-chain-index-staking = { path = "crates/chain/index-staking" }
 
 # ffi (native embedding surface)
 vertex-ffi = { path = "crates/ffi" }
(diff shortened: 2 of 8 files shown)


mfw78:

PR #310, opened 2026-06-13
feat(swarm-node): reject and penalise shallow custody receipts

### LINKED ISSUE #293: pushsync: reject and penalise shallow custody receipts (validate receipt depth against storage radius)

mfw78:
Closes #293.

## Design

A pushsync custody receipt asserts that a node inside the chunk's neighbourhood took custody. A receipt is "shallow" when its signer's proximity to the chunk is below the responsible depth (`PO(signer, chunk) < d`): the chunk never reached the neighbourhood, so it is unretrievable, unreplicated, and the receipt is a free-riding forgery. This change verifies receipt depth at both seams that accept a receipt: the origin uploader (`NetworkChunkProvider::push_to_closest`) and, critically, the relay forwarder (`NetworkForwarder::push`), which validates BEFORE relaying upstream so it never launders a shallow receipt and takes the reputational hit itself.

The verifier lives in `vertex-swarm-api` (`recover_receipt_signer`, `required_receipt_depth`, `verify_receipt_depth`) so both seams share one implementation. It recovers the signer overlay from `receipt.signature` over the 32-byte chunk address via EIP-191 recovery, then derives the overlay with the canonical `compute_overlay(eth, network_id, nonce)`. The nonce binds the overlay derivation only, never the signature, matching the live-network producer; the signer is recovered from the SIGNATURE, never the off-wire `receipt.storer` field, so a multi-hop relay (where `storer` is the immediate downstream peer) measures the real signer rather than the wrong hop.

The required depth is `max(local_observed_depth - tolerance, receipt.storage_radius)`. The locally observed neighbourhood depth (`TopologyHandle::depth`) is the only attacker-independent input, so it sets the FLOOR, minus a small `SHALLOW_RECEIPT_TOLERANCE` for honest topology drift (a peer at the genuine edge of the neighbourhood can sit one bin shallow of our view without being a free-rider). The attacker-controlled `storage_radius` can only RAISE the bar above that floor, never lower it: a storer that under-declares (or zeroes) its radius cannot weaken the locally trusted floor, while a storer that over-declares merely binds itself to a stricter custody claim, which we honour. A shallow or malformed receipt is rejected, the responding peer is scored `InvalidData` through `PeerManager::report_peer` (the #287 reporter), and the uploader continues its walk to the next candidate (retry-via-different-route).

## Test coverage

In `vertex-swarm-api` (`providers.rs`): `signature_is_over_the_32_byte_address_only_and_nonce_is_not_signed` locks the wire signing format and asserts a `(address || nonce)` signature recovers to a different overlay; `required_depth_floors_on_local_and_wire_radius_can_only_raise` pins the floor/raise-only arithmetic including the tolerance; `wire_radius_zero_cannot_lower_the_bar_below_the_local_floor` and `shallow_receipt_is_rejected` prove a radius-0 (or shallow-radius) shallow receipt is still rejected by the local floor; `deeper_wire_radius_raises_the_bar_and_rejects_a_below_radius_signer` proves a deep claimed radius tightens the bar; `recovers_signer_from_signature_not_storer_field`, `all_zero_signature_is_malformed`, `deep_enough_receipt_is_accepted`, and `wrong_message_recovers_to_a_different_overlay_and_is_shallow` cover the remaining cases.

At the seams: `vertex-swarm-builder` (`providers.rs`) covers the origin uploader with `origin_accepts_a_deep_receipt_without_reporting`, `origin_rejects_a_shallow_receipt_and_reports_the_peer`, `origin_rejects_a_shallow_receipt_claiming_radius_zero`, and `origin_rejects_a_malformed_receipt`. `vertex-swarm-node` (`forward.rs`) covers the relay forwarder with `push_relays_receipt_verbatim_and_accounts_both_legs`, `push_rejects_shallow_receipt_reports_signer_and_leaks_no_reservation`, `push_rejects_shallow_receipt_claiming_radius_zero`, and `push_rejects_malformed_receipt_signature` (including the accounting-leak assertions). The integration test `three_node_pushsync_relays_receipt_verbatim_and_accounts` now produces a genuinely deep, address-signed receipt so it exercises the verifier end to end across two relay hops.

## Deferrals

The depth check is a cheap first filter, not a cryptographic guarantee. Nonce-grinding a passing overlay costs ~`2^d` keccak trials (linear in network size, inside the pushsync timeout), and pre-mining defeats the timeout entirely; the boundary that makes forgery unprofitable is stake-bound identity plus retrievability auditing (storage incentives / redistribution game), tracked under #75.

The required depth leans on the receipt's `storage_radius`, which is unsigned (the signer signs the chunk address only) and on a node with a low local view becomes the sole bar a responding peer can set to zero. This residual is tracked in #316.

Retry-policy tuning (alternate-route count and a dedicated shallow-vs-malformed score weight) reuses the existing candidate walk and the shared `InvalidData` weight; a dedicated scoring event is left for a follow-up if telemetry warrants it.


crates/swarm/api/src/providers.rs
mfw78, line 1: This generally seems like a weird place to have these receipt depth error enums and seems to bleed across domain boundaries. Refine this to be more succinct / defined architecturally and respect domain boundaries.

crates/swarm/api/src/providers.rs
```diff
@@ -43,6 +43,129 @@ pub struct PushReceipt {
...
+}
+
+/// Recover the overlay of the node that signed a custody receipt.
+///
+/// The storer signs over the 32-byte chunk address with an EIP-191 personal
+/// sign, exactly as the reference does; the nonce is NOT part of the signed
+/// message, it is only an input to overlay derivation. The overlay is derived
+/// from the recovered Ethereum address with the canonical
+/// [`compute_overlay`](nectar_primitives::compute_overlay) formula
+/// (`keccak256(eth || network_id_le || nonce)`).
```
mfw78: The fact that the `nonce` is bolted in here is a protocol smell that should have an issue filed for the pushsync receipt protobuf to be refined for this.

crates/swarm/builder/src/launch.rs
```diff
@@ -329,6 +331,8 @@ async fn build_client_backed_node(
         Arc::new(topology.clone()),
         Arc::clone(&accounting),
         client_handle.clone(),
+        SwarmSpec::network_id(params.spec.as_ref()),
+        Arc::clone(&reporter),
     );
```
mfw78, line 332: Do any of these already have the spec available, such as topology, if this has the spec available, it would be able to then make it so that the network_id is accessible from that. We look at trying to simplify the API surface here and being smart about what is threaded through (or could be). We should try and fix this code smell here, and potentially file an API/code base wide audit issue to try and shrink the API surface from this parameter explosion.

crates/swarm/builder/src/providers.rs
```diff
@@ -153,3 +189,189 @@ impl<I: SwarmIdentity> SwarmChunkSender for NetworkChunkProvider<I> {
...
+/// of believing a fabricated shallow receipt that the push succeeded.
+fn accept_origin_receipt(
+    receipt: &PushReceipt,
+    address: &ChunkAddress,
+    network_id: nectar_primitives::NetworkId,
+    local_depth: vertex_swarm_api::NeighborhoodDepth,
+    peer: SwarmAddress,
+    reporter: &dyn PeerReporter,
+) -> Result<(), vertex_swarm_api::ReceiptDepthError> {
+    match verify_receipt_depth(receipt, address, network_id, local_depth) {
```
mfw78: Like, it's obvious here that the Networkid and PushReceipt are being bastardised to essentially arrive at the peer's overlay (and being able to confirm it with ecrecover and then matching to the overlay address). This is likely a common path that is used, and we should be making sure that there are decent tools available further down in the stack so that this can be decoded properly / handled in a typed manner earlier than surfacing all the components here. Essentially then later in the codec level or something like that, when the protobufs are updated, we would just go straight to the overlay and verify at that point so we could just discard as malformed if malformed earlier before it reached here.

crates/swarm/api/src/providers.rs
mfw78: Resolved in 188dd26a + a0d3f7f9. The receipt logic no longer lives in `providers.rs` (which keeps only the `PushReceipt` DTO). There is now one domain `Receipt` type in `vertex-swarm-net-pushsync` (`receipt.rs`) carrying the recovered `storer` overlay, with `verify_depth` and `ShallowReceipt` as a method/type next to it; the `vertex-swarm-api` receipt module is gone. The malformed-signature case became a decode-time `PushsyncError::MalformedReceiptSignature` at the wire tier, separate from the domain `ShallowReceipt`. The recovery itself is a layer-2 primitive that belongs in nectar; tracked in #321.

crates/swarm/api/src/providers.rs
mfw78: Filed as #319. The nonce in the overlay-derivation preimage is wire-visible and fork-gated, so there is no v1 change here. #310 does encapsulate the reconstruction in one place now (the decode seam) and documents the nonce role inline; #319 covers refining the receipt protobuf so the signer is self-authenticating and the codec can drop the out-of-band `network_id` reconstruction entirely (gated behind a new `SwarmHardfork` with a conformance vector).

crates/swarm/builder/src/launch.rs
mfw78: Addressed in 188dd26a. `enable_forwarding` drops both `network_id` and `reporter`: it derives `network_id` from `topology.identity().spec().network_id()` and the reporter from a new `SwarmTopologyReporting::reporter()` accessor on the topology trait (implemented via the existing `peer_manager()`). The `launch.rs` call is back to three args. The broader codebase-wide API-surface audit (the same scalar-threading in `NetworkForwarder::new`, `accept_origin_receipt`, and friends) is filed as #320.

crates/swarm/builder/src/providers.rs
mfw78: Done in 188dd26a + a0d3f7f9. Overlay recovery now happens once, at the pushsync decode seam: `on_pushsync_receipt` calls `Receipt::reconstruct(wire, network_id)`, and a receipt whose signature does not recover is rejected as `MalformedReceiptSignature` at decode and never becomes a `Receipt` (the immediate peer is scored `InvalidData`). Everything downstream (forwarder, origin accept, the depth check) only ever sees a typed `Receipt` with the `storer` overlay already recovered and verified, so `network_id` is no longer threaded to the verify path. It is structured so the future protobuf refinement (#319) only changes the codec-internal reconstruction while the domain type and call sites stay put. Upstreaming the recovery to a nectar receipt primitive is #321.

PR #311, opened 2026-06-13
feat(swarm-node): outbound self-throttle for pushsync and retrieval

### LINKED ISSUE #131: pushsync: outbound self-throttle to stay under remote's pseudosettle allowance

### LINKED ISSUE #132: retrieval: outbound self-throttle to stay under remote's pseudosettle allowance

mfw78:
Closes #131. Closes #132.

## Design

Outbound pushsync and retrieval consume the pseudosettle credit a remote peer extends us: every chunk we push or request draws down the time-based allowance that peer grants. Issuing faster than the peer forgives our debt wastes round trips on refusals and, sustained, trips the peer's refuse-or-disconnect threshold and drops us. `SelfThrottle` (`crates/swarm/node/src/throttle.rs`) paces our own send rate to stay under that allowance.

It is the single seam both protocols share: one `SelfRateLimiter<OverlayAddress>` keyed by the peer's overlay, fed by one per-peer allowance signal (`PeerAffordability::allowance_remaining`, built once in the accounting layer), so the two protocols cannot each pace against a private, divergent view of the same allowance. The throttle is consulted at the outbound API (`ClientHandle::retrieve_chunk` / `push_chunk`) before a command is dispatched; on disconnect the client service clears the peer's bucket, and the handle and service share the same `Arc<SelfThrottle>` so the cleared bucket is the one the outbound API paces against.

Token model: one bucket token is one accounting unit (AU). A peer's bucket capacity (burst) is its remaining allowance expressed directly in AU, and the bucket refills at `refresh_rate` AU per second, matching the rate at which the remote forgives our debt. The GCRA derives its per-token replenish time as window divided by capacity, so a window of `capacity / refresh_rate` seconds replenishes one AU every `1 / refresh_rate` seconds regardless of capacity: the refill rate is exactly the forgiveness rate and the burst is the whole allowance. Pacing in AU at the forgiveness rate makes steady-state throughput the remote's real sustainable rate (hundreds of closest-chunk requests per second, fewer for distant chunks) rather than a coarse one-request-per-second cap. The AU magnitudes in play (a disconnect threshold of a few million AU, a worst-case price in the hundreds of thousands) sit well inside the GCRA's `u32` token range.

A request costs the AU price the remote actually meters: a fixed pushsync per-chunk cost and a flat retrieval estimate, both set to the worst-case proximity-0 peer price `base_price * (max_po + 1)`, computed in `launch.rs` via `Au::checked_scale` (saturating on overflow). Charging the worst-case proximity price means the throttle always charges at least what the remote meters, so its pacing stays under the credit the remote extends even for the most distant chunk.

`acquire` re-syncs the live allowance into the peer's bucket on each iteration, then either returns at once or sleeps the bucket's wait hint and retries. It is bounded both by an iteration budget (`MAX_THROTTLE_ITERATIONS`) and a cumulative wall-clock cap (`MAX_THROTTLE_WAIT`, 2s); reaching either releases the request rather than hanging the future, because a refused request is recoverable but a stuck future is not. The wall-clock cap keeps one throttled peer from serialising a sequential pushsync candidate walk: under a steady or growing allowance the steady-state waits are sub-millisecond, so the cap only bites when an allowance is persistently collapsing. The first delay per call increments `pushsync_self_throttled_total{peer_overlay}` or `retrieval_self_throttled_total{peer_overlay}`.

The allowance is polled per acquire rather than subscribed to: the bucket is resized from the live `PeerAffordability` signal whenever a request is about to be admitted, so a between-request allowance change (including a collapse) is observed at the next request that paces against the peer, and `set_quota` is idempotent so an unchanged allowance is a no-op (no free burst). On a shrink the per-key quota re-clamps the bucket's outstanding credit into the smaller capacity so a collapse throttles immediately rather than letting stale oversized credit through.

No wire bytes change: the throttle is purely local outbound pacing, so no `SwarmHardfork` gate is needed. The path is wasm-safe (the parking timer is `futures-timer`, the clock is `vertex-util-runtime`'s `web-time`) and the `vertex-swarm-node` wasm cone builds clean.

## Supporting primitive

`SelfRateLimiter` (`crates/net/ratelimiter/src/self_limiter.rs`) is the outbound dual of the inbound `KeyedRateLimiter`: it parks a request the bucket cannot admit yet rather than refusing it. The keyed limiter gains per-key quotas (`set_key_quota` / `SelfRateLimiter::set_quota`) so an outbound throttle can resize one peer's bucket from a per-peer signal without disturbing the others; a resize is idempotent on an unchanged quota and re-clamps outstanding credit on a shrink.

## Test coverage

Unit tests in `crates/swarm/node/src/throttle.rs`: cost-is-the-AU-price-not-a-settle-unit and steady-state-pace-matches-forgiveness-rate (guards that the per-request cost is the metered AU price and the steady-state pace is the forgiveness rate, not a one-request-per-second cap), saturating-u32 bounds, exhausting/growing/shrinking allowance, zero-allowance-yields-a-valid-throttled-bucket, per-peer independence, clear-drops-peer-bucket, acquire-returns-immediately-under-budget, acquire-waits-then-admits-after-refill, and acquire-releases-under-collapsing-allowance-within-wall-clock-cap.

Integration coverage in `crates/swarm/node/src/client_service.rs`: throttled-push-dispatches-under-budget and throttled-retrieval-delays-when-bucket-drained drive the throttle through the real `ClientHandle` outbound API, and unthrottled-handle-dispatches-immediately confirms a handle built without a throttle behaves exactly as before.

The `SelfRateLimiter` and `KeyedRateLimiter` primitives carry their own suites covering per-key quota resize, idempotent no-free-burst resync, shrink-clamps-outstanding-credit, per-key replenish windows, and FIFO parked-item draining.

## Deferrals

Retrieval cost remains a flat worst-case-chunk estimate because the response size is unknown before it arrives; refining it once response-size measurements exist is tracked in #132 (TODO in the cost-model rustdoc).

The `peer_overlay` metric label is per-peer and unbounded in the number of distinct peers ever throttled; bounding that cardinality so a churny or identity-cycling peer set cannot grow an operator's TSDB without bound is tracked in #318.


crates/swarm/builder/src/launch.rs
```diff
@@ -340,15 +342,41 @@ async fn build_client_backed_node(
...
+    let max_proximity_factor = u64::from(params.spec.max_po()) + 1;
+    let max_chunk_cost = base_price
+        .checked_scale(max_proximity_factor)
+        .unwrap_or(Au::from_amount(u64::MAX));
+    let throttle = Arc::new(SelfThrottle::new(
+        accounting.bandwidth().clone(),
+        refresh_rate,
+        max_chunk_cost,
+    ));
+    let throttled_handle = client_handle.clone().with_throttle(Arc::clone(&throttle));
```
mfw78: Does this mean that we don't bill actual usage in the self throttle? Why have a `max_chunk_cost` and not just calculate it on a JIT basis?

crates/swarm/node/src/client_service.rs
```diff
@@ -33,9 +34,16 @@ pub(crate) const DEFAULT_CHANNEL_CAPACITY: usize = 256;
 /// carries the request is the correlation, and no shared rendezvous state
 /// exists. Concurrent requests for the same chunk address never collide, so
 /// callers may freely race the same address across peers.
+/// Outbound self-throttle shared by both chunk-transfer protocols.
+///
+/// When set, [`ClientHandle::retrieve_chunk`] and [`ClientHandle::push_chunk`]
+/// pace themselves under the remote peer's pseudosettle allowance before the
+/// request is dispatched. Without it, requests dispatch immediately, exactly as
+/// before the throttle existed (and as the unit tests rely on).
```
mfw78, line 42: It seems like we should also file a follow-up issue that let's say there is a large payment channel (chequebook consumer), if we are to deliver adequate QoS, we might still need some bandwidth rate limiting. We should look at having some toggles/parameters around this, especially for the client (retrieval / pushsync) protocols, as well as pullsync for later with storer nodes. Potentially file as follow-up issues in the correct milestone / place

crates/swarm/node/src/client_service.rs
```diff
@@ -352,7 +402,12 @@ impl ClientService {
 
             ClientEvent::PeerDisconnected { peer_id, overlay } => {
                 debug!(%peer_id, %overlay, "Peer disconnected");
-                // TODO: Clean up any pending operations for this peer
+                // Drop the peer's throttle bucket so memory does not grow with
+                // the count of distinct peers seen and a reconnect starts from a
+                // fresh allowance rather than stale credit.
+                if let Some(throttle) = &self.throttle {
+                    throttle.clear(&overlay);
+                }
```
mfw78, line 414: But is this what bee does? We want to avoid that if a connection flaps, then it in effect loses the tracking on the accounting and we inadvertently might consume more (or be entitled to more, ie. owed credit). We want to avoid stumbling inadvertently early into swap territory where we would be paying.

crates/swarm/node/src/throttle.rs
```diff
@@ -0,0 +1,494 @@
...
+//!
+//! # Cost per request
+//!
+//! - **Pushsync**: the representative AU price of a maximal chunk delivery. A
+//!   pushed chunk has a fixed maximal footprint, so the cost is a per-protocol
+//!   constant.
+//! - **Retrieval**: the same flat maximal-chunk price. We do not know the
+//!   response size before it arrives, so we charge the maximal-chunk estimate up
+//!   front. TODO(#132): refine once response-size measurements make a tighter
+//!   estimate safe.
```
mfw78: This seems kind of stupid as we should be able to calculate this straight out. When doing pushsync, the overlay of the peer that we send it to dictates straight up what the cost is going to be, so in effect we know this and can calculate this. Maximal chunk delivery costs is going to be far too restrictive.

As for retrieval, the pricing is based on per chunk, read the specs / book of swarm. Response size is irrelevant, what is relevant is the distance the chunk is from the peer that we ask.

mfw78 (commented):
Does this implementation also take into account the accounting changes that occur, such that lets assume that we are rate limiting and pushing through maximally at our pseudosettle rate, what if we are connected to another peer (or peers) that we are engaged in pseudosettle with that are using us for retrieval or pushsync. In which case, we should be able to earn AU there and this becomes additional amounts that we can use. 

Also, having said this, it looks like there should be some toggle / buffer that allows for some margin so that we don't eat exactly 100% of the pseudosettle allowance.

crates/swarm/builder/src/launch.rs
```diff
@@ -343,24 +343,22 @@ async fn build_client_backed_node(
...
-    let max_proximity_factor = u64::from(params.spec.max_po()) + 1;
-    let max_chunk_cost = base_price
-        .checked_scale(max_proximity_factor)
-        .unwrap_or(Au::from_amount(u64::MAX));
     let throttle = Arc::new(SelfThrottle::new(
         accounting.bandwidth().clone(),
+        Arc::new(accounting.pricing().clone()),
         refresh_rate,
-        max_chunk_cost,
+        params.bandwidth.throttle_allowance_percent(),
```
mfw78: I actually don't understand why the SelfThrottle couldn't actually just take an accounting object, and from this it can get everything it needs (well, in terms of accounting.bandwidth() and accounting.pricing(). Refresh rate and throttle allowance are seemingly both parameters for bandwidth, so refresh_rate should liekly be there as well and one could just pass a reference to the params.bandwidth and clone out what config is needed and simplify this whole thing?

crates/swarm/builder/src/launch.rs
mfw78: Addressed in b2c0ad34: `max_chunk_cost` is gone. The throttle now charges the exact per-chunk `peer_price(peer, address)` just-in-time, which is the same figure the accounting layer debits, for both pushsync and retrieval. The latest review refines the constructor wiring further (passing the accounting object instead of separate handles); handled in the new round.

crates/swarm/node/src/client_service.rs
mfw78: Filed as #322: QoS bandwidth rate-limiting with operator toggles for retrieval and pushsync, distinct from the accounting self-throttle, layered on the existing ratelimiter primitives; pullsync-for-storer noted for later. Milestone "Chunk Transfer v1: Retrieval & Pushsync".

crates/swarm/node/src/client_service.rs
mfw78: The accounting balance already survives a flap (bee-parity: it is kept in memory and not dropped on disconnect). 2abceb09 caps the throttle capacity at the payment-threshold headroom with an 85% margin, so a post-flap unpaced burst can no longer cross the swap trigger into paying. The deeper transient-vs-final flap semantics (preserve vs drain the pacing bucket on reconnect) are filed as #323.

crates/swarm/node/src/throttle.rs
mfw78: Addressed in b2c0ad34: the throttle now uses the exact per-chunk proximity price for both pushsync and retrieval. The "response size is unknown" rationale was incorrect (retrieval is priced purely by proximity, no size term) and has been removed; #132's size-based cost model is obsolete and is closed by this PR.

crates/swarm/builder/src/launch.rs
mfw78: Done in c2dde02f. `SelfThrottle::new` is now `new(accounting, &params.bandwidth)`: it derives `pricing()` and the affordability (`PeerAffordability`) from the accounting object, and `refresh_rate` + `throttle_allowance_percent` from the bandwidth config, so the call site collapses to a single `SelfThrottle::new(&accounting, params.bandwidth)` with no separate refresh-rate/percent plumbing. `refresh_rate` was already a `BandwidthConfig` field, so nothing moved. The config is taken concretely (`&DefaultBandwidthConfig`) so `throttle_allowance_percent` stays an inherent accessor and there is no public-trait change. Internal fields and behaviour are unchanged. Leaving this thread open for you to confirm.

PR #312, opened 2026-06-13
feat(ffi): memory-bounded streaming chunk get/put pipeline

### LINKED ISSUE #289: ffi/wasm: memory-bounded streaming chunk get/put pipeline with backpressure
## Problem

The FFI surface (`crates/ffi/src/api/client.rs`) exposes only single-shot `upload_chunk` / `download_chunk`, with every payload crossing the boundary as an owned `Vec<u8>` copy (host → `Vec<u8>` → `StampedChunk` reconstruct, and `chunk.into_bytes().to_vec()` on the way out). There is no streaming, no backpressure, and no multi-chunk pipeline — a host wanting a file must drive one round-trip per 4 KiB chunk across the bridge. On WASM the client node builds (`vertex-swarm-node` is CI-enforced on `wasm32-unknown-unknown`) but `vertex-ffi` is native-only and the demo binary has a hard `--max-memory=1073741824` ceiling, so memory bounds matter doubly there.

What exists to build on:

- `StreamSink<LogLine>` event-stream pattern (`crates/ffi/src/api/logging.rs`) and the pull-based bounded metrics snapshot (`crates/ffi/src/api/metrics.rs`) — both flutter_rust_bridge-codegen friendly.
- Chunk payloads are already `Bytes` (refcounted) internally; copies only happen at the FFI edge.
- Bounded command channel (`DEFAULT_CHANNEL_CAPACITY = 256`, `crates/swarm/node/src/client_service.rs`) with `try_send` + dropped-command metric.

## Proposal

Design a **memory-bounded streaming chunk get/put pipeline** usable natively, over FFI, and on WASM:

1. **Core (FFI-agnostic):** a bounded multi-chunk session in the client layer — `get_stream(addresses…) -> bounded stream of verified chunks` and `put_stream(chunks…) -> bounded stream of receipts` — with an explicit in-flight window (N chunks ≈ N×4KiB resident, configurable), prefetch/pipelining inside the window, and backpressure: no new upstream request is issued until the consumer drains a slot. Builds on #178 (correlation ids / fan-out) for concurrency within the window.
2. **FFI binding:** expose the session via `StreamSink<ChunkData>` for downloads and a feed/ack contract for uploads; the Rust side owns the bounded buffer so a slow host pauses network reads instead of growing bridge memory. Minimize copies: one unavoidable copy at the bridge per direction; everything inside stays `Bytes`.
3. **WASM binding:** same core session surfaced through wasm-bindgen (ReadableStream/async iterator), sharing the window/backpressure implementation.
4. **Budget accounting:** in-flight window should be expressible in bytes, not just chunk count, so embedders on constrained targets (mobile, browser shared-memory cap) can set hard ceilings.

## Acceptance

- [ ] Streaming N chunks holds O(window) memory regardless of N (test with window=4, N=1000, assert peak resident buffer).
- [ ] Host-side stall stops upstream requests (backpressure observable via metric).
- [ ] Exactly one payload copy per direction at the FFI edge; zero extra copies internally.
- [ ] Same session core compiles for native and wasm32; FFI and wasm bindings are thin adapters.
- [ ] Verified-only delivery: chunks cross the boundary only after address/stamp verification (#287).

mfw78:
Closes #289.

## What this adds

A memory-bounded streaming get/put pipeline so a caller can fetch or store a whole manifest (thousands of chunks) without issuing one round-trip at a time and without an unbounded fan-out blowing the heap on a slow or hostile consumer.

The core is two FFI-agnostic `Stream`s in `crates/swarm/node/src/stream.rs`. `get_stream(provider, addresses, config)` yields one verified chunk per address in request order; `put_stream(sender, chunks, config)` (and its fallible sibling `try_put_stream`) yields one receipt per chunk. Both hold their in-flight requests in a `FuturesOrdered` (output order always matches input order, independent of completion order) built on the per-request outbound futures from #178. The window is expressed in BYTES via `StreamConfig { window_bytes, max_concurrency }`: downloads reserve `MAX_CHUNK_BYTES` per in-flight retrieval (a chunk's true size is unknown until it arrives), uploads reserve each chunk's real encoded size, and a new request is admitted only when the reservation fits the window and the concurrency cap and only when the stream is polled. `StreamConfig::new` clamps both knobs to at least one, so a window smaller than a single chunk degrades to one-at-a-time streaming rather than deadlocking. Downloads deliver `VerifiedStampedChunk` only (#287): each chunk is proven to answer its requested address via `verify_answers` before it leaves the stream, so a peer that returns the wrong bytes for an address surfaces as an error item, never as a chunk a consumer might trust.

## Design

Backpressure is enforced by construction, not by a sink. Both streams only issue a new request inside their `refill`, which runs only when the stream is polled and admits at most one replacement per drained slot, so a host that stops polling stops all new requests and nothing accumulates on its behalf. Both host bindings are pull-based, so the host's pull rate is the backpressure signal end to end. The native FFI returns opaque handles `VertexDownloadStream` / `VertexUploadStream` driven by an async `next()` the host awaits one item at a time; the wasm-bindgen adapter (`crates/swarm/node/src/stream_wasm.rs`) returns `WasmGetStream` / `WasmPutStream`, JS async iterators with the same `next()` shape. Neither binding spawns a pump that pushes into a host queue; the flutter_rust_bridge `StreamSink` is reserved for the low-volume logging surface, where unbounded buffering is acceptable, and is not used for bulk chunk transfer.

The pipeline holds exactly one payload copy per direction at the boundary. Download-out copies once (`into_bytes().to_vec()` on native, `Uint8Array::from` on wasm); inside Rust the chunk stays `Bytes`. Upload-in copies once: `reconstruct_upload` takes the `VertexChunkUpload` by value and moves its `Vec` into `Bytes` via the zero-copy `Vec -> Bytes` conversion, so the only materialization is the bridge copy into the boundary `Vec`. Upload memory is bounded the same way the network window is: `put_stream` pulls chunks lazily from a `Peekable` source iterator and materializes a `StampedChunk` only when it admits the push, and the FFI feeds a reconstruct-on-demand iterator, so the only chunks built into strong types at any moment are the in-flight ones. The host still owns the `Vec<VertexChunkUpload>` it passed by value (an unavoidable cost of a by-value API), but Rust holds no second resident copy and copies no payload a second time. `try_put_stream` carries a fallible feed so a per-chunk reconstruction failure (a byte/address mismatch or bad stamp) surfaces as that chunk's ack in order rather than aborting the whole upload.

Failure handling is per-item throughout: a per-address retrieval miss, a wrong-bytes rejection, or a per-chunk push failure surfaces as an `Err` item in that slot and the stream continues with the rest. Dropping a handle drops the core stream and all its in-flight futures, cancelling the outstanding network requests. The upload reservation release uses `saturating_sub` so a logic slip cannot panic the pipeline, and the FFI `u64` window is `try_from`-clamped to `usize::MAX` on a 32-bit host.

## Test coverage

17 unit tests in `crates/swarm/node/src/stream.rs` cover ordering, the byte-window concurrency cap for both directions, the window-smaller-than-a-chunk degenerate case, verify-before-yield rejection of wrong bytes, per-item error continuation, empty-input termination, the config clamp, lazy upload materialization (a 1024-item feed against a gated sender materializes at most window+1 chunks), and inline feed-error ordering for `try_put_stream`. The FFI boundary reconstruction and identity helpers are covered by the `vertex-ffi` unit tests. `cargo fmt --all` and `cargo clippy --all-targets --all-features -- -D warnings` are clean on both crates, and the wasm client cone builds on `wasm32-unknown-unknown`.

## Deferrals

The wasm stream constructors (`get_stream_wasm` / `put_stream_wasm`) are plain `pub fn` taking `Arc<dyn SwarmChunkProvider>` / `Arc<dyn SwarmChunkSender>` and are not yet `#[wasm_bindgen]`-exported, so JS cannot construct a stream today. This is deferred to the wasm client builder milestone, since the browser provider is native-only in `vertex-swarm-builder`; the adapter still compiles on wasm32 (CI-enforced) and shares the exact core path the native FFI exercises, which the core unit tests cover.

The download byte window reserves a fixed `MAX_CHUNK_BYTES` per slot rather than the chunk's true (often smaller) size, so the effective payload ceiling is conservative and never exceeded. A tighter post-arrival accounting is a possible future refinement, not a safety gap.

There is no per-retrieval deadline, so a hung head-of-line peer can stall a multi-chunk download: memory stays bounded (the stalled future counts against the window so refill stops), but because `get_stream` yields strictly in input order, a peer that accepts the head address and never delivers blocks every later in-window chunk with no surfaced error. The fix belongs in the retrieval protocol (a per-request deadline that turns into an error item) rather than the streaming adapter, so the single-chunk `download_chunk` path benefits too. Tracked in #314.

No on-the-wire protocol change is introduced, so no hardfork gate is required.

## Notes for reviewers

The committed flutter_rust_bridge bindings (`crates/ffi/src/frb_generated.rs` and the Dart output under `crates/ffi/bindings/dart`) are regenerated for the new opaque handles; CI does not run the codegen, so the committed output is the build input.


mfw78 (commented):
OK, is it wise to have these stream.rs and stream_wasm.rs placed in the swarm node crate? This seems like creep / domain bleed, and instead the respective streams should be placed in ffi and some wasm library or similar. 

Suggest look at other projects / architectures as to how to lay this out so that we don't get excessive compile times, domain bleed, but yet remain light and agile for development / maintenance purposes.

mfw78:
Addressed the review (the streams in swarm-node concern) and rebased onto current main.

**Relocation (Option A).** The streaming get/put pipeline moved out of `vertex-swarm-node` into a new transport-agnostic crate `vertex-swarm-stream` (`crates/swarm/stream`). The core (`stream.rs` -> the crate root) depends only on the `vertex-swarm-api` traits, nectar primitives, `vertex-tasks`, and `futures`; it has zero node internals. The native FFI adapter stays in `vertex-ffi`, and the wasm-bindgen adapter is a `#[cfg(target_arch = "wasm32")] pub mod wasm` in the same crate. `vertex-swarm-node` dropped the streams and the `wasm-bindgen`/`js-sys` deps the pipeline had forced into it. The core is now reusable by the FFI native surface, the gRPC chunk service, and a future browser client without any of them depending on each other. Documented in `crates/swarm/stream/AGENTS.md`, the root area table, `crates/ffi/AGENTS.md`, and the wasm-cone list.

**Security hardening (from a security pass over the relocated pipeline).** The GET path advertised "memory stays flat at roughly the window", but the retrieval delivery codec accepted frames up to 5 MB before rejecting them, so a hostile peer could force `download_concurrency` times that transient allocation against the window. The delivery decode frame is now capped at `MAX_DELIVERY_SIZE` (the largest legitimate chunk, a single-owner chunk, plus a full stamp plus protobuf framing, ~4378 bytes, derived from the chunk-size constants), so an oversized frame is rejected at the length prefix before its body is buffered. The per-slot `MAX_CHUNK_BYTES` reservation is now an honest bound on peak transient memory. Conformance is preserved: a maximal legitimate delivery still decodes, and the cap is not wire-visible (a local accept-limit, not a format change). New tests cover the maximal-accept floor, the oversized-reject, and the boundary.

**Deferred.** The PUT path's in-flight fan-out is still bounded by `max_concurrency` rather than the byte window (a low-severity, host-controlled fan-out gap). The naive fold of the byte budget into the PUT slot cap deadlocked the existing window tests, so it is tracked for a careful fix in #324 rather than rushed here.

CI is green on the rebased revision.

Issue #318, opened 2026-06-14
observability: bound peer_overlay cardinality on self-throttle counters


mfw78:
## Problem

The outbound self-throttle increments `pushsync_self_throttled_total{peer_overlay}` and `retrieval_self_throttled_total{peer_overlay}` with a per-peer `peer_overlay` label whose value is the remote overlay address. The label set is unbounded in the number of distinct peers the node ever throttles, and an emitted series is never retracted: `SelfThrottle::clear` drops the in-memory bucket on disconnect, but a Prometheus counter series, once created, persists until metric TTL.

## Current behaviour

`crates/swarm/node/src/throttle.rs`, in `SelfThrottle::acquire`: the `metrics::counter!(... "peer_overlay" => peer.to_string())` increment (around line 280). The cardinality tradeoff is documented in the rustdoc at the same call site (around line 256) and was deliberately deferred during implementation.

The disconnect-clear bounds bucket memory but not metric memory.

## Why it matters

A node with a large, churny, or adversarially identity-cycling peer set accumulates one counter series per distinct overlay it ever throttles. Overlay addresses are not bounded by the live peer count: a peer that reconnects under a fresh identity, or an attacker cycling overlays, mints a new series on each throttle event.

This grows the node's resident metric set and the operator's TSDB without bound: a memory and scrape-cost denial-of-service against the operator's own monitoring stack, not against the node's serving path.

## Proposed

Decide the supported aggregation explicitly rather than leaving it to the operator:

- Drop the `peer_overlay` label from the counter (keep a single per-protocol throttled counter and expose per-peer detail only through a bounded, TTL'd gauge or a debug endpoint); or
- Bound the label cardinality at emission (for example label only a small set of most-throttled peers and fold the rest into an `other` bucket).

At minimum, document the recommended recording-rule aggregation and metric TTL in the operator docs so a default deployment does not unknowingly retain unbounded series.

## Scope

- Affected: `crates/swarm/node` (`src/throttle.rs`)

## Notes

Surfaced while reviewing #311 (MERGED), the outbound self-throttle for pushsync and retrieval that introduces these counters.

Verified on main (HEAD 37baadc1): the `peer_overlay` label is still emitted at `crates/swarm/node/src/throttle.rs:280`, with the deferral rustdoc at line 256.

mfw78:
[...]

PR #339, opened 2026-06-15
fix(swarm-node): per-protocol retrieval/pushsync timeout typing and staggered candidate race (head-of-line)

### LINKED ISSUE #314: retrieval: per-request deadline for ordered streaming get (head-of-line stall on a withholding peer)
## Summary

The streaming download pipeline `get_stream` (in `crates/swarm/node/src/stream.rs`) yields chunks strictly in input-address order via `FuturesOrdered`, but the retrieval path carries no per-request deadline. A peer that accepts the substream for the head address and then never delivers holds the head slot open indefinitely. Later addresses in the window complete but cannot be yielded past the stalled head, so the whole transfer stalls with no progress and no error.

This is an availability gap, not a memory-exhaustion gap. The stalled retrieval still counts against the byte window, so `refill` stops admitting new work and resident memory stays bounded at roughly `window_bytes`. The transfer simply never advances and never errors.

## Where

`crates/swarm/node/src/stream.rs`: `GetStream::poll_next` / `retrieve_verified` drive each retrieval as a `FuturesOrdered` entry with no timeout wrapper.

The provider/protocol layer does not supply its own deadline either. `crates/swarm/net/retrieval/src/protocol.rs` and `crates/swarm/net/retrieval/src/lib.rs` contain no timeout or deadline, and neither does the `retrieve_chunk` provider impl in `crates/swarm/builder/src/providers.rs` or `crates/swarm/node/src/client_service.rs`. So a head-of-line slow/withholding peer is not bounded anywhere in the path the stream depends on.

## Scenario

Request a manifest of N addresses. A peer answers address[1..N] promptly but accepts and then withholds the substream for address[0]. The host awaiting `download_stream().next()` (or the wasm `WasmGetStream::next`) blocks forever on the first item; the completed later chunks sit in the ordered set unyielded. No error item is ever produced, so a host cannot distinguish a stall from a slow-but-live transfer and cannot skip the bad head address.

## Why it matters

The remote peers serving retrievals are untrusted. A single adversarial or simply unresponsive peer holding one head address can deny an entire multi-chunk download with no surfaced error, defeating the per-address "error item, keep going" resilience the rest of the pipeline provides (a per-address miss already surfaces as an `Err` item and the stream continues; a silent stall does not). The same no-deadline exposure exists for the single-chunk `download_chunk`, so this is a pre-existing protocol-layer gap rather than one this PR introduces, but the streaming pipeline makes it more impactful because one stalled head blocks every later in-window chunk.

## Suggested direction

Enforce a per-retrieval deadline so a withholding peer surfaces a timeout `Err` item in its slot and the stream advances. The natural home is the retrieval protocol layer (a per-request deadline that turns into a `SwarmError` on timeout), not bolted onto the streaming adapter, so the single-chunk path benefits too. Related but distinct from #64 (CLI-configurable per-protocol stream timeouts), which is about exposing timeout knobs rather than this ordered-stream head-of-line behavior.

Surfaced while reviewing #312. Tracked separately because it is a protocol-layer change, and #312 intentionally does not introduce a wire or protocol change.

mfw78:
Closes #314.

## Correction to the issue premise

#314 describes a withholding head peer causing a silent infinite stall. On the current libp2p rev that does NOT occur: the outbound retrieval substream is dispatched with `SubstreamProtocol::...with_timeout(30s)`, and the forked libp2p polls that `Delay` around the whole upgrade future including the blocked `framed.try_next()` Delivery read, so a peer that accepts the substream then withholds the Delivery is converted to `StreamUpgradeError::Timeout -> DialUpgradeError` after the deadline and surfaces as an `Err` item in the `FuturesOrdered` slot. The repro test in this PR locks that invariant.

The real defects this PR fixes are the ones that made the stall hard to see and amplified it.

### Per-protocol named deadline

The 30s deadline was a single `Config::timeout` shared by retrieval, pushsync, pricing, pseudosettle and swap, so the retrieval liveness guarantee was accidental and undocumented. `Config` now carries `retrieval_timeout` and `pushsync_timeout` (both default `30s`, matching the reference's per-request `RetrieveChunkTimeout`), threaded into the retrieval and pushsync `SubstreamProtocol` calls; pricing/pseudosettle/swap keep the shared `timeout`. The split is documented at the `Config` definition and the liveness invariant is named in rustdoc so this is not re-filed. The value is local enforcement (a `DialUpgradeError`), not wire-visible, so no fork gate.

### Typed timeout error

The timeout used to flatten to `ChunkTransferError::Protocol(String)` (a string-match anti-pattern with no clean metric label). A typed `ChunkTransferError::TimedOut` is added; the retrieval and pushsync `DialUpgradeError` arms match `StreamUpgradeError::Timeout` explicitly and send it, while still producing `FailureKind::Protocol` so the existing `RetrievalFailure` / `PushFailure` (-2.0) peer scoring and `PeerSelector` exclusion are unchanged. `TimedOut` is retryable. It surfaces identically as a per-address `Err` item (stream) and a `SwarmError` source (single-chunk FFI/gRPC), native and wasm.

### Head-of-line fix: staggered candidate race

The real amplification: the stream provider `NetworkChunkProvider::retrieve_chunk` walked its 3 candidates strictly in series (up to 3x30s before a head slot could error, blocking every later in-window chunk). The origin path `Client::get` already raced candidates with a 500ms stagger. This PR extracts that race into a shared `race_candidates` helper (`retrieval_race.rs`) and uses it for both `Client::get` and the stream provider, so a withholding head is overtaken by the staggered second candidate in ~500ms instead of the per-attempt deadline. Losing futures are dropped when the race resolves, releasing any accounting/forwarder reservation (tested).

### Observability

Two bounded, label-free counters `swarm.client.retrieval_timeouts_total` / `pushsync_timeouts_total` on the single timeout emission site, plus a structured `debug!` carrying `peer_overlay` / `address` / `elapsed`. No `peer_overlay` metric label (avoids the unbounded-cardinality debt tracked in #318); per-peer attribution lives in scoring and the log.

## Scope

Pushsync is included because it has the byte-identical stall surface (`framed.try_next()` for the receipt) and the same shared-timeout and string-flattening defects. The config fields are a plain default with no CLI flag, which is the seam #64 threads later (CLI-configurable per-protocol timeouts) without touching call sites.

## Tests

A `libp2p-swarm-test` repro: a peer negotiates the retrieval substream and headers then never writes the Delivery; the outbound resolves `Err(ChunkTransferError::TimedOut)` within a short `retrieval_timeout`, not forever and not the shared 30s. Race helper + provider tests: a withholding head is overtaken by the staggered second within the stagger; all candidates failing yields `AllPeersFailed`; no reservation leak on race resolution.

`cargo fmt`, `cargo clippy --workspace --all-features -- -D warnings`, and the affected crate test suites are clean.


crates/swarm/builder/src/providers.rs
```diff
@@ -67,36 +67,34 @@ impl<I: SwarmIdentity> SwarmChunkProvider for NetworkChunkProvider<I> {
...
+                served_by: result.peer,
+            }),
+            Err(RaceFailure::NoCandidates) => Err(SwarmError::network_msg(
+                "no connected peers available for retrieval",
+            )),
+            Err(RaceFailure::AllFailed(e)) => Err(SwarmError::AllPeersFailed {
+                address: *address,
+                attempts,
+                source: Box::new(e),
+            }),
```
mfw78, line 98: This is duplicated in client_service.rs. This duplication that is forming between providers.rs and client_service.rs needs to be flagged for a follow-up architectural review / code deduplication effort.

crates/swarm/node/src/staggered_race.rs
mfw78, line 1: As this file can be used for any staggered race, should we limit it in name to 'retrieval_race.rs' ?

mfw78: Done in f916834a. Renamed `retrieval_race.rs` to `staggered_race.rs` and parameterized the stagger (`race_candidates` now takes a `stagger: Duration` rather than hardcoding it), so it is a generic staggered candidate-race helper. `RETRIEVAL_STAGGER` (500ms) stays as the value the retrieval callers pass, not a binding of the helper. Added a module caveat that it is for read-style any-peer races only and must not be applied to pushsync or other directed writes (fanning out a write creates redundant custody and multiplies bandwidth cost; directed writes use sequential fallback, not a fanned-out race).

crates/swarm/builder/src/providers.rs
mfw78: Flagged as a follow-up: filed #340 to consolidate the duplicated retrieval/push candidate-orchestration forming between `providers.rs` (`NetworkChunkProvider`) and `swarm_client.rs`/`client_service.rs` (`Client`). This PR deduplicated one slice (the shared staggered-race helper); #340 covers the broader orchestration (candidate selection -> per-candidate throttle/affordability/accounting -> race-or-fallback -> error mapping) so the two surfaces stop drifting.

Issue #214, opened 2026-06-10
storage: add key-presence DbTx primitives (exists/first_key) and migrate the cursor consumers


mfw78:
Rescoped 2026-06-20 to the residual after the cursor half landed.

## Already done (#396)

- `RedbCursorRO<T>` streams a table without materialising it (`crates/storage/redb/src/cursor.rs`).
- It is wired into the trait: `DbTx::cursor<T>() -> Box<dyn DbCursorRO<T> + Send>` (`crates/storage/src/traits.rs`).

## Residual (this issue)

`DbTx` still has no key-presence primitive, and no consumer was migrated, so the four `TODO(#214)` sites still pay full-value reads and an O(N) key materialisation.

1. Add two read primitives to `DbTx` (`crates/storage/src/traits.rs`):
   - `fn exists<T: Table>(&self, key: T::Key) -> Result<bool, DatabaseError>` with a default impl (`self.get::<T>(key)?.is_some()`), overridden in the redb backend to check presence without deserialising the value.
   - `fn first_key<T: Table>(&self) -> Result<Option<T::Key>, DatabaseError>` with a default impl over `cursor::<T>()?.first()`, overridable by a backend with a cheaper path.
   - Implement the overrides on the redb backend (`crates/storage/redb/src/tx.rs`, both `RedbReadTx` and `RedbWriteTx`).

2. Migrate the consumers and delete the `TODO(#214)` markers:
   - `crates/swarm/storer/src/db_store.rs::contains()` -> `exists`.
   - `crates/swarm/storer/src/db_store.rs::for_each()` -> stream via `cursor()` with early exit, so `Reserve::evict_oldest` reads only the first key instead of materialising every key (kills the O(N) scan + 32*N-byte allocation per call).
   - `crates/swarm/storer/src/db_store.rs::put()` duplicate check -> `exists`.
   - `crates/swarm/postage/src/store.rs::contains()` -> `exists`.

## Notes

Pure performance and cleanup: no wire change, no behaviour change beyond avoiding redundant deserialisation and full-table scans. Keep the trait surface minimal and idiomatic (default methods on `DbTx`, backend overrides only where they buy something).

PR #350, opened 2026-06-16
feat(swarm-rpc): add unordered, count-bounded streaming chunk RPCs

mfw78:
## Overview

This branch began as the unordered, count-bounded streaming chunk RPCs but has grown into the integration branch that carries the gRPC streaming work, the RPC and component-model reshaping that surrounds it, and the full postage and reserve storage-incentives stack. It is the accumulation of thirty reviewed and merged changes (`origin/main..stack/01-grpc-chunk-streaming`), totalling roughly 11,400 insertions across 79 files. Each change landed as its own reviewed pull request into this stack; this body is an overview of the whole, intended to be read before the rebase merge into `main`.

Because the constituent pull requests were merged individually, the history is a linear sequence of those commits. A rebase merge preserves that sequence on `main` rather than collapsing it.

### gRPC chunk streaming and the stream toolkit
- Unordered, count-bounded streaming chunk RPCs (the original change).
- `served_by` on `VerifiedChunk` and a `MaybeSend` future alias (#360).
- Free `parse_address` and `reconstruct` helpers, and `get_stream_from` (#362).
- The `ChunkClient` capability alias and the `ChunkClientExt` blanket extension trait (#364).
- The gRPC chunk service and the FFI client driven off `ChunkClientExt` and the free helpers (#366, #370).
- A bounded error channel on streaming retrieval (#368).

### RPC, transport, and the component model
- `NodeProviders<C>` introduced, gRPC registration collapsed, then folded into `GrpcAdapter<C>` so gRPC sits only at the edge (#372, #378).
- The api component model reshaped to `ClientComponents<T, C>`, with the builder consuming it and its duplicate deleted (#379, #380).
- The transport seam abstracted, with gRPC as one implementation (#381).
- The `ChunkServiceProvider` alias dropped in favour of `ChunkClient`, and the dead `swarm_client.rs` removed (#382, #384).

### Build and nectar
- The workspace now tracks the nectar `main` branch rather than a pinned revision (#386).
- vertex consumes the nectar `StampedChunk`, dropping its own definition (#387).

### Storage engine
- A lazy, read-only redb cursor over a held transaction (#214 via #396).

### Redistribution: sampling and proof of entitlement
- Neighbourhood sampling and entitlement primitives (#393).
- The reserve sampler and proof of entitlement (#395 via #397), with benchmarks (#394).
- Consensus candidate-feed filters and the exact-stamp witness (#410).

### Postage and the reserve
- `ReserveStore` and `BinCursorStore` traits with a `HasReserve` accessor, and the local overlay threaded into the reserve for proximity eviction (#388, #389).
- The `DbReserve` adapter over the storage engine (#390).
- `DbBatchStore` and the on-chain batch-event ingest seam (#400).
- The stamp-index arbiter, with newest-timestamp-wins arbitration per `(batchID, stampIndex)` (#401).
- The reserve reworked to a per-stamped-entry six-table store (#403).
- Size-driven storage radius, committed depth, and batch-expiry eviction, with the reserve capacity read from the spec power-of-two chunk count rather than a byte budget (#404, and the capacity correction wired in #391).
- `StorerComponents` and the storer gRPC services wired up (#391).
- Storer pushsync ingest: `is_responsible_for`, store, then sign a custody receipt off the node identity (#392).

## Notes

- The postage and batch-store traits in nectar were made synchronous (sync core, async only at the gRPC and FFI edges) before this stack repinned to nectar `main`; the reserve and validator paths here consume the synchronous surface.
- The original title no longer reflects the scope; the change set is best understood through the area summary above.

This pull request was prepared with AI assistance.


crates/swarm/rpc/proto/chunk.proto
```diff
@@ -13,6 +13,18 @@ service Chunk {
...
+  // streaming back one receipt per chunk in input order. The byte-bounded
+  // upload pipeline backs this RPC: it prefetches ahead of the consumer up to a
+  // memory window and stops pulling new chunks the instant the receipt consumer
+  // stops draining, so a long upload stays flat in memory on both ends.
+  rpc UploadChunks(stream UploadChunkRequest) returns (stream UploadChunkResponse);
+
+  // RetrieveChunks retrieves a stream of chunks by address, streaming back one
+  // response per requested address in input order. Backed by the same
+  // byte-bounded download pipeline as the single-chunk retrieval.
+  rpc RetrieveChunks(stream RetrieveChunkRequest) returns (stream RetrieveChunkResponse);
```
mfw78, line 23: Should we just have streamed versions of all of these? Consider what would be required by a client.

crates/swarm/builder/src/rpc.rs
```diff
@@ -60,8 +60,8 @@ impl<I: SwarmIdentity, C: Send + Sync> HasTopology for ClientRpcProviders<I, C>
     }
 }
 
-impl<I: SwarmIdentity, C: SwarmChunkProvider + SwarmChunkSender + Clone> RegistersGrpcServices
-    for ClientRpcProviders<I, C>
+impl<I: SwarmIdentity, C: SwarmChunkProvider + SwarmChunkSender + Clone + Unpin + 'static>
```
mfw78: These types are starting to get a little unwieldy. Should we perhaps look at type aliases or `derive_more`. Check if it's actually required as well. Do we _need_ `Unpin` or are there alternative ways to convey this typing information that is more succinct / less verbose (perhaps established crates that have this, and are commonly used within the rust ecosystem).

crates/swarm/rpc/src/grpc/chunk.rs
mfw78, line 1: I think we should actually allow for unordered retrieval, and somewhat unordered upload (in that the acks / receipts coming back for upload might be out of order from what was uploaded by the client)

mfw78: Ordering is simply likely to cause too many head of line issues and this should be pushed down to the client so they can determine how they want to handle it

crates/swarm/stream/src/lib.rs
```diff
@@ -133,6 +133,12 @@ impl StreamConfig {
         max_concurrency: 32,
     };
 
+    /// High-throughput preset (96-deep) for a native node serving a bulk download.
+    pub const NATIVE_DOWNLOAD: Self = Self {
+        window_bytes: 96 * MAX_CHUNK_BYTES,
+        max_concurrency: 96,
+    };
+
```
mfw78: This preferably should be derived from some constants as this should be roughly determinable on an optimal range (based on some hypothetical network), then allow for a fixed percentage buffer, instead of a "thumb in the air" measurement.

crates/swarm/rpc/proto/chunk.proto
mfw78: Done. `RetrieveChunk` and `UploadChunk` now have streaming counterparts (`RetrieveChunks`/`UploadChunks`), and I added `HasChunks` so existence checks batch too: a client doing dedup-before-upload or a manifest walk now has a streaming path for all three. Following your other note, the streamed responses are a `oneof` of typed success/error messages (`RetrievedChunk`/`UploadReceipt`/`ChunkError`) with binary addresses rather than parallel scalar fields.

crates/swarm/builder/src/rpc.rs
mfw78: Checked, and `Unpin` was not actually needed: it only propagated up because `GetStream`/`PutStream` were polled via `get_mut()`. I pin-projected both with `pin-project-lite` (tiny and ubiquitous in the ecosystem) so the provider no longer needs `Unpin`, and dropped the bound from all four sites. The repeated `SwarmChunkProvider + SwarmChunkSender + Clone + 'static` bound is now named once via a `ChunkServiceProvider` trait alias. `derive_more` does not help here (it derives `From`/`Display` and friends, not trait-bound aliases), so the alias is the right tool.

crates/swarm/rpc/src/grpc/chunk.rs
mfw78: Done, and I agree this is the right model. Both pipelines now complete unordered (`FuturesUnordered`) and every item carries its `ChunkAddress`, so correlation and reassembly are the client's job. Upload receipts likewise return in completion order, each echoing its chunk address, so an ack arriving out of order is still attributable.

mfw78: Agreed: ordering is now pushed to the client. A slow or missing chunk no longer head-of-line-blocks the rest, and a per-item failure is one `ChunkError` item (the response `oneof`'s error arm), not a teardown of the whole stream. The serving handlers also consume the inbound stream lazily, so a long request stream does not pre-buffer on the server.

crates/swarm/stream/src/lib.rs
mfw78: Done. `NATIVE_DOWNLOAD` is now derived from named constants via a bandwidth-delay product (Little's law): an assumed sustained serve rate times the mean per-chunk retrieval latency, plus a fixed 25% jitter buffer. A unit test pins the derived value to a sane band so it cannot silently drift. Relatedly, per a separate discussion the pipelines now bound by chunk count rather than bytes (a chunk is size-bounded, so a count bounds memory without inspecting each chunk); byte/bandwidth limiting is left as a seam at the libp2p connection layer.

PR #352, opened 2026-06-16
feat(swarm-node): default the launcher to self-throttle

### LINKED ISSUE #343: [FEATURE] Lightweight client can trip remote settlement on bursts
## What problem does this solve?

The lightweight (browser/wss) `ClientLauncher` carries no accounting stack. Without throttling, a burst of retrieval or pushsync requests can exceed a remote peer's pseudosettle allowance and trigger settlement against the local node. The launcher should pace outbound requests by default so a normal client does not provoke settlement under bursty load.

## Describe the solution

Build a minimal client-accounting instance in the `ClientLauncher` and attach a `SelfThrottle` to both the handle and the service. This paces outbound requests so they stay under each peer's pseudosettle allowance. Throttling is enabled by default; `without_throttle()` opts out for callers that want the previous behaviour.

## Alternatives considered

- Leave throttling off and rely on callers to rate-limit: this pushes a subtle accounting concern onto every consumer and reproduces the settlement issue by default.
- Apply a single global rate limit rather than per-peer pacing: a global cap does not respect each peer's individual pseudosettle allowance and would either throttle too aggressively or fail to prevent settlement.
- A per-IP cap was considered for a related abuse vector; it is intentionally left out of this change and remains OFF by default pending a redesign.

## Implementation notes

- The change is localized to `crates/swarm/node/src/node/launch.rs`.
- The `SelfThrottle` is wired into the handle and the service so both request paths are paced.
- The accounting instance built here is the minimal client variant, not the full node-accounting stack.
- The per-IP cap mentioned above is not part of this work and stays disabled until its redesign lands.

## Are you willing to PR this?

Yes (implemented in the linked PR).

## Testing

- `cargo check`, `cargo clippy`, and the crate's `cargo test` pass on the stack tip.
- No additional manual verification was performed for this change.

## AI assistance disclosure

AI Assistance: implemented end-to-end with Claude Code (Anthropic): code, tests, and this description. Changes were reviewed for correctness and pass the crate test suites on the stack tip.

mfw78:
## What does this PR do?
Gives the lightweight (browser/wss) `ClientLauncher` a minimal client-accounting stack and a `SelfThrottle` so a request burst no longer trips a remote peer's settlement trigger.

## Changes
- Build a minimal client-accounting instance for the lightweight launcher, which previously carried no accounting stack.
- Attach a `SelfThrottle` to the handle and service, pacing outbound requests under each peer's pseudosettle allowance.
- Enable self-throttle by default; `without_throttle()` opts out.
- Default the lightweight launcher to a conservative `LIGHTWEIGHT_THROTTLE_ALLOWANCE_PERCENT` of 25 percent, threaded through to `AccountingBuilder` and `SelfThrottle::new`.

## Why
The lightweight launcher carries no settlement and no forwarder, so nothing ever debits its peer-map and the throttle never sees the live headroom shrink. It must therefore stay conservatively below the remote's settlement trigger on its own. `with_throttle_allowance_percent` clamps the configured percentage into `1..=100` so an out-of-range value can never zero-size or oversize the bucket.

## Breaking changes
None. This adds opt-out throttling to the lightweight launcher; no public API removals or wire format changes.

## Testing
- cargo fmt, clippy (`--all-targets -D warnings`), and cargo test pass for the touched crates `vertex-swarm-accounting` and `vertex-swarm-node`.
- New tests: `launcher_attaches_throttle_by_default`, `without_throttle_opts_out`, `lightweight_throttle_allowance_is_conservative`, plus `client_service` throttle dispatch tests.
- [x] Unit tests pass

## Related issues
Closes #343

## AI assistance disclosure
AI Assistance: Claude Code (Opus 4.8) used for rebase and conflict resolution.

## Notes for reviewers
Focus on the minimal client-accounting wiring and on the `SelfThrottle` attachment to the handle and service, confirming that pacing stays within each peer's pseudosettle allowance and that `without_throttle()` cleanly opts out.

## Checklist
- [x] Self-review completed
- [x] Tests added/updated
- [x] PR title is descriptive

### DIFF
--- a/crates/swarm/accounting/core/src/config.rs
+++ b/crates/swarm/accounting/core/src/config.rs
@@ -66,6 +66,19 @@ impl<P> BandwidthConfig<P> {
     pub fn throttle_allowance_percent(&self) -> u8 {
         self.throttle_allowance_percent
     }
+
+    /// Override the self-throttle's payment-threshold consumption percent,
+    /// clamped into 1..=100.
+    ///
+    /// Used by embedders that pace against a static (never-debited) headroom and
+    /// therefore want a more conservative fraction than the default adaptive
+    /// loop uses. The value is clamped rather than rejected so an out-of-range
+    /// argument cannot size the bucket to zero or above the live headroom.
+    #[must_use]
+    pub fn with_throttle_allowance_percent(mut self, percent: u8) -> Self {
+        self.throttle_allowance_percent = percent.clamp(1, 100);
+        self
+    }
 }
 
 impl TryFrom<&BandwidthArgs> for BandwidthConfig<FixedPricingConfig> {
--- a/crates/swarm/node/src/client_service.rs
+++ b/crates/swarm/node/src/client_service.rs
@@ -57,6 +57,12 @@ impl ClientHandle {
         self
     }
 
+    /// Whether an outbound self-throttle is attached to this handle.
+    #[must_use]
+    pub fn has_throttle(&self) -> bool {
+        self.throttle.is_some()
+    }
+
     /// Send a command to the network layer.
     ///
     /// Non-blocking `try_send`: callers such as the libp2p event loop must not
@@ -248,6 +254,12 @@ impl ClientService {
     ) -> Self {
         self.accounting = Some(OriginAccounting { pricing, bandwidth });
         self
+    }
+
+    /// Whether an outbound self-throttle is attached to this service.
+    #[must_use]
+    pub fn has_throttle(&self) -> bool {
+        self.throttle.is_some()
     }
 
     /// Get a handle for sending commands.
--- a/crates/swarm/node/src/node/launch.rs
+++ b/crates/swarm/node/src/node/launch.rs
@@ -9,11 +9,12 @@
 //! and issue chunk reads and writes. The full native stack (storage, chain,
 //! settlement, RPC) goes through `vertex-swarm-builder` instead.
 
-use std::time::Duration;
+use std::{sync::Arc, time::Duration};
 
 use eyre::{Result, WrapErr};
 use libp2p::{Multiaddr, PeerId};
 use nectar_primitives::SwarmAddress;
+use vertex_swarm_accounting::{AccountingBuilder, DefaultBandwidthConfig};
 use vertex_swarm_api::{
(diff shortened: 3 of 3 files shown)


mfw78:
Reworked and reparented onto #351 (`stack/02-blameless-retrieval-miss-scoring`). The branch was rebased onto the new parent so it now carries a single commit, the launcher self-throttle default, replayed clean off the stale base drift that has since landed on main.

Triage defect addressed: the lightweight (browser/wss) `ClientLauncher` attaches a default-on `SelfThrottle` but runs no forwarder, so its client-accounting peer-map is never debited and `allowance_to_payment_threshold` always reports the full static payment threshold. The throttle is therefore a static pacer, not the adaptive live-debit loop the native builder wires through its forwarder.

This is acceptable for a balance-less lightweight client, so the fix makes it honest and tested rather than fabricating an adaptive loop that is not wired:

- Documented in `maybe_attach_throttle` and on the new `LIGHTWEIGHT_THROTTLE_ALLOWANCE_PERCENT` constant that the lightweight launcher paces against a deliberately static, conservative fraction of the payment-threshold headroom, precisely because nothing debits its peer-map. Reviewers will not mistake it for the native adaptive feedback loop.
- Sized the static bucket to a conservative 25 per cent of the payment-threshold headroom (added `BandwidthConfig::with_throttle_allowance_percent`) rather than the adaptive native default of 85 per cent, so a never-shrinking allowance still stays well below the remote's settlement trigger.
- Added launcher-level tests asserting the throttle is attached to both the handle and the service by default, that `without_throttle()` opts both out, and that the lightweight allowance percent is more conservative than the adaptive default. Introduced `ClientHandle::has_throttle` and `ClientService::has_throttle` as the inspection seam, and factored the attach decision into a testable `maybe_attach_throttle` helper.

Verified: `vertex-swarm-node` builds, and the three new launcher tests plus the bandwidth config and throttle suites pass.

AI assistance: this rework was carried out with AI assistance and reviewed by me before submission.

mfw78:
Superseded by the converged client core. assemble_client_core now builds the launcher SelfThrottle over the live shared accounting (origin debit + pseudosettle), which is strictly better than this open-loop provisional throttle; maybe_attach_throttle / LIGHTWEIGHT_THROTTLE_ALLOWANCE_PERCENT have no home on main. Closing without merge. #343 is addressed by the converged core; #498 has no residual left. The rest of the train (#354/#355/#356/#357/#358) is being rebuilt onto main.

PR #354, opened 2026-06-16
feat(swarm-builder): escalate to a wide last-resort retrieval race

### LINKED ISSUE #345: retrieval: rescue a chunk whose closest candidates all miss (sparse-bin tail)
## Problem

Retrieval of a single chunk aborts as soon as its closest raced candidates all miss. A chunk in a sparse proximity bin (few or no peers in the neighbourhood actually hold it) can be left unreachable from a given vantage point even though a farther peer, whose own forwarding chain reaches the chunk's storers, could still serve it. The result is spurious whole-retrieval failures on the genuinely sparse tail.

## Current behaviour (post-#544)

On main after #544, `crates/swarm/node/src/chunks/providers.rs` selects a pool of `RETRIEVE_CANDIDATE_WIDTH = 8` closest connected peers, applies skip-busy filtering and the economic selector, then races at most `RETRIEVE_MAX_ATTEMPTS = 3` legs with a staggered start. On `RaceFailure::AllFailed` it maps straight to `SwarmError::AllPeersFailed`. The selection pool widened from 3 to 8, but the raced legs are still bounded to 3 and there is no escalation, so a chunk whose 3 closest free-slot legs miss still fails. The gap is real and unaddressed.

## Why the original wide-escalation design was dropped

PR #354 (race 6, then escalate once to a wide race over ~30 peers) was closed. A red-team after #542/#543/#544 found:

- Swarm-semantically unsound: racing peers progressively farther in proximity order chases peers that are structurally less likely to hold or usefully forward the chunk. Chunks live near their address and retrieval is forwarding-Kademlia, so a far-peer fan-out has low marginal hit-probability and high latency.
- It amplified to ~30 metered legs per chunk, the exact effect #544's `RETRIEVE_MAX_ATTEMPTS = 3` bound exists to prevent, and its race bypassed the per-peer in-flight cap and skip-busy, re-creating the overrun and debt pressure #519/#520/#544 fixed.
- not-found versus withholding is undetectable on the wire (Swarm has no explicit negative response; peers simply time out), so escalation cannot tell "absent" from "slow" and must stay conservative on every failure.

## Proposed (measurement-gated)

1. Measure first. Confirm whether the sparse-bin tail still fails under #544's 8-peer pool (browser-demo and native bulk download: count whole-chunk `AllPeersFailed`, and how often a fresh farther peer would have served it). If the tail is negligible, close this: the wider selection pool sufficed.
2. If the tail is real, fix it with a single bounded-refill proximity walk, not a separate wide tier:
   - One bounded-concurrency race that walks the proximity-ordered candidate pool, refilling a freed slot with the next-closest untried peer on each failed leg, until success or the pool is exhausted.
   - Bound the total metered legs per chunk (keep the spirit of `RETRIEVE_MAX_ATTEMPTS`; raise it only with data) so a sparse tail can reach a farther forwarding chain without N-fold amplification of doomed requests.
   - Respect the per-peer in-flight cap and skip-busy on every leg; no bypass.
   - Stay conservative on transport failure and timeout; do not pile onto withholding peers.
   - Prefer modest reach over a 30-peer fan-out.

## Scope

- Affected: `crates/swarm/node/src/chunks/providers.rs` (and `staggered_race.rs` if the race grows a refill mode).
- Gate: measurement before implementation.

## Notes

Part of the light-client retrieval hardening epic #526; sibling of #519.

Alternatives still rejected:

- Loop-retrying the closest candidates inside the handler. Whole-chunk reliability across a bulk download is the client's job; the consumer re-streams the addresses that errored in a fresh round once the per-peer allowances refill, so loop-retry here duplicates that and holds the head of the download pipeline's in-flight window, starving concurrency.
- Blanket-enlarging the common-path candidate count. Each retrieval is an accounting-gated request drawing on the per-peer self-throttle allowance, and the staggered race starts a new candidate on every failure, so a large common-path candidate set turns one throttled head into an N-fold amplification of doomed requests under a wide concurrent download.

mfw78:
## What does this PR do?

When the staggered candidate race exhausts its primary candidates, it escalates to a wide last-resort race across a broader candidate set instead of failing the retrieval, which rescues chunks whose close candidates transiently miss.

## Changes

- Add a wide last-resort race that runs once the staggered primary-candidate race is genuinely exhausted (`AllPeersFailed`).
- Broaden the candidate set used for this final attempt rather than returning a retrieval failure.
- Raise `RETRIEVE_CANDIDATE_COUNT` from 3 to 6 as a deliberate, documented tuning change.

## Defect fixed in this rework

The original escalation widened the candidate set with `closest_to(chunk, RETRIEVE_WIDE_CANDIDATE_COUNT)`, which returns a proximity-sorted top-k. Its leading entries were exactly the peers the common race had just exhausted, so the wide tier re-dispatched the same doomed legs and burned the self-throttle budget before reaching any genuinely new peer. The rework threads the already-tried set into the escalation and filters it out via `fresh_wide_candidates`, racing only peers that have not been tried; when none is fresh the common-path `AllPeersFailed` stands. The wide candidate set is built lazily, so the closure is consulted only on the `AllPeersFailed` arm and `NoCandidates` still does not trigger escalation.

## Breaking changes

None.

## Testing

- `cargo fmt --all`, `cargo clippy --all-targets --all-features -- -D warnings`, and `cargo test` pass for `vertex-swarm-builder` and `vertex-swarm-node`.
- New `wide_escalation` tests, including `all_peers_failed_triggers_a_wide_race_over_only_fresh_peers`, fail if the exclusion is removed (verified by mutation) and pass with the fix.
- [x] Unit tests pass

## Related issues

Closes #345

## AI assistance disclosure

## Notes for reviewers

This is a standalone PR off `main`; the single feature commit is confined to `crates/swarm/builder/src/providers.rs`. The key thing to check is the escalation path: that the wide last-resort race only fires once the primary staggered race is genuinely exhausted, that it excludes already-tried peers so it never re-dispatches the doomed legs, and that it broadens the candidate set without changing the success/failure semantics of the existing path.

## Checklist

- [x] Self-review completed
- [x] Tests added/updated
- [x] PR title is descriptive

### DIFF
--- a/crates/swarm/builder/src/providers.rs
+++ b/crates/swarm/builder/src/providers.rs
@@ -23,8 +23,22 @@ const PUSHSYNC_SOURCE: ReportSource = ReportSource::Protocol("pushsync");
 /// Number of closest peers to try when pushing a chunk before giving up.
 const PUSH_CANDIDATE_COUNT: usize = 5;
 
-/// Number of closest peers to try when retrieving a chunk before giving up.
-const RETRIEVE_CANDIDATE_COUNT: usize = 3;
+/// Number of closest peers raced when retrieving a chunk; kept small to avoid
+/// amplifying doomed requests onto saturated peers.
+const RETRIEVE_CANDIDATE_COUNT: usize = 6;
+
+/// Wide last-resort candidate set, raced once only when the normal race misses,
+/// to rescue chunks in sparse proximity bins.
+const RETRIEVE_WIDE_CANDIDATE_COUNT: usize = 30;
+
+/// The wide candidates the common race did not already try, in proximity order.
+/// `wide` is proximity-sorted top-k, so its head is the already-`tried` peers;
+/// excluding them stops the wide tier re-running their doomed legs.
+fn fresh_wide_candidates(tried: &[SwarmAddress], wide: Vec<SwarmAddress>) -> Vec<SwarmAddress> {
+    wide.into_iter()
+        .filter(|peer| !tried.contains(peer))
+        .collect()
+}
 
 /// Chunk provider using ClientHandle for network retrieval.
 #[derive(Clone)]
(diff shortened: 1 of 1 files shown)


mfw78:
Closing after a red-team against current main (#542/#543/#544).

Do not merge or rebase: this is a re-implementation, not a rebase. The file it edits (crates/swarm/builder/src/providers.rs) was relocated to crates/swarm/node/src/chunks/providers.rs by #528, and the constant it changes (RETRIEVE_CANDIDATE_COUNT) was replaced by #544's RETRIEVE_CANDIDATE_WIDTH=8 / RETRIEVE_MAX_ATTEMPTS=3.

Three problems beyond the staleness:

1. Collides with #544. Racing 6 then escalating to ~30 metered legs per chunk is exactly the amplification #544's MAX_ATTEMPTS=3 bound exists to prevent, and the wide race dispatches with no per-peer in-flight permit, bypassing the skip-busy cap and re-creating the overrun and debt pressure #519/#520/#544 just fixed.
2. Swarm-semantically unsound. Racing peers progressively farther in proximity order chases peers structurally less likely to hold or usefully forward the chunk (chunks live near their address; retrieval is forwarding-Kademlia). A 30-peer far fan-out has low marginal hit-probability and high latency.
3. not-found versus withholding is undetectable on the wire (no negative response; peers time out), so AllPeersFailed cannot tell absent from slow, and the magic constants (6, 30) carry no measurement.

The underlying gap (a sparse-bin chunk failing when its closest peers miss) is real and has been reframed in #345: measurement-gated, and if still real after #544's 8-peer pool, solved with a single bounded-refill proximity walk that respects the per-peer cap and skip-busy, not a wide-to-30 escalation.

PR #355, opened 2026-06-16
perf(swarm-node): cap concurrent in-flight retrievals per peer

### LINKED ISSUE #346: retrieval: cap concurrent in-flight retrievals per peer
## Problem

When download concurrency is high but the candidate set is narrow, retrievals funnel onto a few peers.

Each of those peers ends up with too many in-flight retrieval substreams open on a single connection at once, so the peer resets us (substream-open timeouts). The result is failed retrievals and worse tail latency, even though the overall concurrency budget could be spread more evenly.

## Current behaviour

There is no bound on the number of simultaneous in-flight retrievals issued to any one peer, so a narrow candidate set collapses the whole concurrency budget onto a single connection.

## Proposed

Add a per-peer in-flight limiter that caps the number of concurrent retrievals issued to any single peer.

- Total download concurrency is unchanged; it is simply prevented from collapsing onto one peer.
- When a peer is at its cap, further work spreads to other candidates instead of piling more substreams onto the saturated connection.
- The limiter composes with the existing wider candidate-escalation path, so spillover lands on other candidates rather than blocking.

This keeps peers below the point where they reset us and smooths tail latency.

## Scope

- Affected: the retrieval issue path (`crates/swarm/net/retrieval`)
- Out of scope: a per-IP cap. That addresses a different abuse class, needs a separate redesign, and stays off by default pending it.

## Notes

Alternatives considered:

- A single global concurrency limit only: does not stop a narrow candidate set from concentrating all of that budget on one peer, which is the actual failure mode here.
- Reactive back-off after a reset only: wastes a full round trip per reset and degrades latency; bounding in-flight work proactively avoids the resets in the first place.

Builds on the retrieval wide-candidate-escalation work; the implementation is stacked on that parent PR and the stack must not be reordered.

Verified: not present on `origin/main`; the retrieval crate (`crates/swarm/net/retrieval`) carries no per-peer in-flight bound. Tracked by PR #355 (open).

mfw78:
## What does this PR do?

Caps the number of concurrent in-flight retrieval substreams a download opens to any single peer, so a wide native download cannot flood one chunk-holder with unbounded substreams (the substream-open timeout storm) while overall concurrency still spreads across peers.

## Changes

- Add `PeerInflightLimiter`: a per-peer concurrency limiter backed by one lazily created `Semaphore` per peer overlay, each seeded with `limit` permits; a permit is held for the lifetime of the request future and released on drop (including when a losing race leg is cancelled), so the count tracks real in-flight substreams.
- Add `DEFAULT_PEER_INFLIGHT_RETRIEVALS = 12` for the native path; a zero limit is clamped to one so a request can always make progress.
- Wire the limiter into `ClientHandle` via `with_inflight_limiter`; `retrieve_chunk` now acquires a per-peer permit before the throttle, so a peer at its cap waits for a slot rather than spinning the throttle.
- Add `PeerInflightLimiter::forget` to drop per-peer bookkeeping on disconnect, bounding map growth; outstanding permits keep their own semaphore alive and a reconnect re-creates a fresh full-capacity entry.
- Re-export `PeerInflightLimiter` and `DEFAULT_PEER_INFLIGHT_RETRIEVALS` from `vertex-swarm-node`.

## Re-homed wiring

The limiter is built and attached inside `assemble_client_core` (in `node/core.rs`), alongside the existing `SelfThrottle` wiring, rather than in the builder-only launch path. It is attached to both the throttled handle and the client service, so both client tiers get the per-peer cap: the lightweight launcher and the full builder both route through `assemble_client_core`. The single shared accounting and throttle wiring is unchanged; the limiter shares the same disconnect hook, so a `PeerDisconnected` event clears that peer's throttle bucket and forgets its in-flight entry through the same path.

## Permit lifetime and forget semantics

The permit covers the throttle `.await` and the full `rx.await` request lifetime, so a peer at its cap blocks for a slot rather than racing past the limiter, with no lock held across an await. Tearing down a peer that still has in-flight permits is a deliberate, bounded relaxation of the cap for that disconnect window, not a leak: the outstanding permit keeps its own semaphore alive and a fresh full-capacity entry is created on the next acquire.

## Breaking changes

None. The limiter is opt-in on `ClientHandle` (a handle without it dispatches retrievals with no concurrency cap, exactly as before). The new constructor, builder method, and re-exports are purely additive. Both client build paths now attach the limiter by default, which changes runtime concurrency behaviour but not any public API.

## Testing

`cargo fmt --all`, `cargo clippy -p vertex-swarm-node -p vertex-swarm-builder --all-targets --all-features -D warnings` (and `-p vertex-swarm-node --features swap`), and `cargo test -p vertex-swarm-node -p vertex-swarm-builder --all-features` all pass. The seven added tests cover the limiter directly: two concurrent permits admitted and a third blocked until one is dropped, per-peer independence, the zero-limit clamp, `forget` resetting a peer to full capacity, `forget` pruning the per-peer map, a `PeerDisconnected` event forgetting the in-flight entry through the service, and a `ClientHandle` with a limiter still dispatching a retrieval when a slot is available. The wasm cone builds: `cargo build --target wasm32-unknown-unknown -p vertex-swarm-node --features swap`. No live mainnet download was exercised in this PR.

- [x] Unit tests pass

## Related issues

Closes #346

## AI assistance disclosure

## Notes for reviewers

- The choice of 12 for the native cap: a wide native download spreads concurrency across 20 or more peers, so steady-state per-peer load sits well below the cap and a healthy download is not bottlenecked, while the degenerate single-holder case is held in check.
- Permit lifetime in `retrieve_chunk`: the permit is acquired before the throttle and held for the whole request future, so a cancelled race leg releases its slot on drop.
- `forget` semantics on disconnect versus outstanding permits keeping the semaphore alive.

## Checklist

- [x] Self-review completed
- [x] Tests added/updated
- [x] PR title is descriptive

### DIFF
--- a/crates/swarm/node/src/client_service.rs
+++ b/crates/swarm/node/src/client_service.rs
@@ -2,10 +2,12 @@
 //!
 //! Owns channels to `ClientBehaviour` and processes incoming events.
 
+use std::collections::HashMap;
 use std::sync::Arc;
 
 use nectar_primitives::ChunkAddress;
-use tokio::sync::{mpsc, oneshot};
+use parking_lot::Mutex;
+use tokio::sync::{Semaphore, mpsc, oneshot};
 use tracing::{debug, warn};
 use vertex_swarm_api::{
     Au, BandwidthDebit, PeerReporter, ReportSource, SwarmLocalStore, SwarmPricing,
@@ -25,6 +27,58 @@ const PUSHSYNC_SOURCE: ReportSource = ReportSource::Protocol("pushsync");
 
 pub(crate) const DEFAULT_CHANNEL_CAPACITY: usize = 256;
 
+/// Default per-peer cap on concurrent in-flight retrieval substreams (native),
+/// to stay under a remote's inbound substream limit.
+pub const DEFAULT_PEER_INFLIGHT_RETRIEVALS: usize = 12;
+
+/// Per-peer concurrency limiter for outbound retrievals (one semaphore per peer).
+///
+/// Entries are created lazily on first acquire and pruned on disconnect via
+/// [`Self::forget`], so the map stays bounded by the live peer set rather than
+/// growing with every peer ever contacted.
+#[derive(Debug)]
+pub struct PeerInflightLimiter {
+    limit: usize,
+    peers: Mutex<HashMap<OverlayAddress, Arc<Semaphore>>>,
+}
+
+impl PeerInflightLimiter {
+    /// Build a limiter admitting `limit` concurrent retrievals per peer (clamped to >=1).
+    #[must_use]
+    pub fn new(limit: usize) -> Self {
+        Self {
+            limit: limit.max(1),
+            peers: Mutex::new(HashMap::new()),
+        }
+    }
+
+    /// The per-peer semaphore for `peer`, created on first use.
+    fn semaphore(&self, peer: OverlayAddress) -> Arc<Semaphore> {
+        Arc::clone(
+            self.peers
+                .lock()
+                .entry(peer)
+                .or_insert_with(|| Arc::new(Semaphore::new(self.limit))),
+        )
+    }
+
+    /// Wait for an in-flight slot against `peer`; hold the permit for the request.
+    async fn acquire(&self, peer: OverlayAddress) -> Option<tokio::sync::OwnedSemaphorePermit> {
+        self.semaphore(peer).acquire_owned().await.ok()
+    }
+
+    /// Drop per-peer slot bookkeeping for a disconnected peer; safe while requests are in flight.
+    pub fn forget(&self, peer: &OverlayAddress) {
+        self.peers.lock().remove(peer);
+    }
+
+    /// Number of peers with a live semaphore entry (the limiter's memory footprint).
+    #[must_use]
+    pub fn tracked_peers(&self) -> usize {
+        self.peers.lock().len()
+    }
+}
+
 /// Handle for sending commands to the network layer.
 ///
 /// Request methods ([`Self::retrieve_chunk`], [`Self::push_chunk`]) thread a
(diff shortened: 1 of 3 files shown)


mfw78:
Reworked and reparented onto the new stack.

This branch now carries a single signed commit (`19185768`) replayed onto the reworked `stack/05-retrieval-wide-candidate-escalation` parent (`da924a45`), which itself sits on the rebase-merged gRPC streaming and postage/reserve train now on `main`. The stale base drift that previously sat under this commit is already on `main`, so only the feature commit was replayed.

The triage flagged a leak: `PeerInflightLimiter` keeps one lazily created semaphore per peer overlay, but the advertised disconnect cleanup (`forget()`) was unwired dead code, so the per-peer map grew without bound, and the limiter was not fully wired into `ClientService`. The fix:

- Wires the cleanup into the `PeerDisconnected` event path. `ClientService` now holds the shared `Arc<PeerInflightLimiter>` and calls `forget(&overlay)` on disconnect, dropping the peer's semaphore entry. An outstanding permit keeps its own semaphore alive via the `Arc`, so this is safe mid request.
- Completes the `ClientService` wiring. The launcher shares one limiter instance between the outbound handle (which enforces the cap on the download path) and the service (which prunes on disconnect), mirroring how the self-throttle is shared.
- Adds tests asserting the per-peer map shrinks on `forget` and through the service's `PeerDisconnected` path (so the map stays bounded by the live peer set), alongside the existing per-peer cap assertions.

`vertex-swarm-node` builds, `clippy -D warnings` is clean for both the node and builder crates, and the new tests pass.

This work was completed with AI assistance.

mfw78:
Superseding this PR. A multi-agent design review (see #520) concluded the per-peer cap should be non-blocking skip-busy rather than a blocking semaphore: a peer at its cap must be skipped so the chunk dispatches to the next-closest live peer, not block the race leg (blocking is head-of-line blocking and regresses the staggered race in staggered_race.rs).

The cap should be a single conservative uniform constant (target ~4, sized under the reference peer's ~32-stream per-connection multiplexer budget shared across all protocols), not the fixed 12 here, and not keyed on peer type: the node-type signals (identify agent string, handshake node-type bool) are both unauthenticated and spoofable, and the binding per-connection limit is type-independent, so per-type capping adds spoof surface for no gain on a reference-dominated network. A clean seam is kept for a later hardfork-gated, handshake-negotiated higher vertex-to-vertex cap.

The reviewed machinery here (lazy per-peer map, forget-on-disconnect, permit-on-drop) informs the rebuild. Tracked in #520, which subsumes #346. Closing in favour of that.

PR #356, opened 2026-06-16
feat(swarm): per-IP live connection cap (off by default) and raised IP-tracker limits

### LINKED ISSUE #347: networking: per-IP connection cap and remote-reset backoff exemption
## Problem

The connection layer has two distinct weaknesses.

1. A single remote IP can open an unbounded number of connections to the node, so one peer (or a misbehaving or malicious host) can exhaust connection resources with no upper limit.
2. When a remote peer triggers an early disconnect (a remote-induced reset), our dialer wrongly treats it as a local dial failure and applies dial-backoff. That backoff starves the refill loop, so legitimate future dials are delayed because of behaviour we did not cause.

## Current behaviour

Every disconnect, including a remote-induced reset, counts toward dial-backoff, conflating remote resets with genuine local dial failures and starving refill.

## Proposed

Two targeted changes to the connection-management code.

1. An optional per-IP connection cap. When configured, the node limits how many concurrent connections a single IP may hold and rejects further connections from that IP once the cap is reached. The cap is off by default and opt-in, so the mechanism lands without changing default behaviour.
2. Exempt remote-induced resets from dial-backoff. Distinguish a remote early-disconnect from a local dial failure at the point where backoff is recorded, so only genuinely local failures feed the backoff timer and the refill loop keeps working at its normal rate.

## Scope

- Affected: `crates/swarm/peers` (per-IP connection cap and the dial-backoff record path)

## Notes

The per-IP cap stays OFF by default pending a redesign: a safe default and the eviction/accounting semantics need more design work before it can be trusted on by default. Do not assume it is enforced unless explicitly configured.

Alternatives considered:

- A global connection cap only: does not address resource concentration from a single abusive IP, so a per-IP dimension is still needed.
- Always counting every disconnect toward backoff (the current behaviour): simpler, but conflates remote resets with local dial failures and starves refill, which is the bug being fixed here.

Verified: `origin/main` already has a per-IP identity-cycling tracker (`crates/swarm/peers/peer-manager/src/ip_tracker.rs`, `max_overlays_per_ip` / `max_sightings_per_ip`), but that bounds distinct overlays and sightings, not concurrent connections; no live per-IP connection cap exists. Tracked by PR #356 (open).

mfw78:
## What does this PR do?
Hardens per-IP connection handling: adds an admission-time per-IP live-connection cap (off by default), raises the identity-cycling detector defaults so legitimate high-density IPs are not penalized, and stops arming dial-backoff when a peer disconnects early due to a remote-induced cause.

## Changes
- Add a live per-IP concurrent-connection admission cap to the peer manager. `on_peer_connected` now returns a `ConnectionAdmission` outcome (`Admitted` / `RejectedIpCap` / `RejectedBinFull`) so the topology caller tears down a rejected connection instead of leaving it half-open; admission is decided before any registry takeover, so a rejected duplicate can never evict a healthy incumbent.
- Track live connections per IP group (`live_ip_connections`, `connection_ip_group`) and decrement on disconnect; `LocalSubnet`-trust and above are exempt. The cap is disabled by default (`DEFAULT_MAX_CONNECTIONS_PER_IP = None`), pending a redesign around inbound-connection limits plus gossip analysis. The machinery is retained but inert.
- Raise identity-cycling detector defaults for legitimate high-density IPs: `DEFAULT_MAX_OVERLAYS_PER_IP` 16 -> 128 and `DEFAULT_MAX_SIGHTINGS_PER_IP` 64 -> 512.
- Expose `--network.peer.max-per-ip` and `--network.peer.max-overlays-per-ip` on the CLI. `--network.peer.max-per-ip` defaults to `0` (unlimited), so the cap ships off by default in line with the library default and the stated design; `0` means unlimited.
- Do not arm dial-backoff on a remote-induced early disconnect. Only a `LocalClose` early disconnect (our-side abort or a clean refusal) is penalized; a transient remote reset keeps the peer dialable. Logic extracted as `early_disconnect_penalizes`.

## Breaking changes
None at runtime: the new per-IP cap is off by default and the new CLI flags are additive. The `PeerManager::on_peer_connected` signature now returns `ConnectionAdmission` instead of `()`, and `peer-manager` re-exports `ConnectionAdmission`; this is a source-level change for any in-tree caller of that API.

## Testing
`cargo clippy --all-targets --all-features` and `cargo test --all-features` pass for the affected crates (`peer-manager`, `topology`, `node`) on the rebuilt tip. New and updated unit tests cover the admit-before-takeover ordering (the wss cascading-collapse and replaced-then-rejected regressions), the per-IP cap admit/reject and disconnect-decrement paths, and `early_disconnect_penalizes` for `LocalClose` versus remote `ConnectionError`. No live multi-node network testing was performed; the per-IP cap path is exercised only by unit tests because it is off by default.
- [x] Unit tests pass

## Related issues
Closes #347

## Rebuild note
This branch was rebuilt directly onto `main` to carry only the two commits that constitute the per-IP cap (the cap plus the dial-backoff fix). Both cherry-picked cleanly onto the current `main` tip.

## AI assistance disclosure

## Notes for reviewers
Key things to check: the admission ordering in `on_peer_connected` (admission decided before any registry takeover, and the reserved IP slot released if downstream admission later drops the peer), the disconnect-time decrement bookkeeping in `connection_ip_group`/`live_ip_connections`, and that the per-IP cap is genuinely inert by default. For the topology change, confirm `early_disconnect_penalizes` only penalizes `LocalClose` and that a remote `ConnectionError`/transient reset leaves the peer dialable.

## Checklist
- [x] Self-review completed
- [x] Tests added/updated
- [x] PR title is descriptive

### DIFF
--- a/crates/swarm/node/src/args/peer.rs
+++ b/crates/swarm/node/src/args/peer.rs
@@ -6,6 +6,14 @@ use vertex_swarm_api::{
     DEFAULT_PEER_BAN_THRESHOLD, DEFAULT_PEER_MAX_PER_BIN, DEFAULT_PEER_WARN_THRESHOLD,
     PeerConfigValues,
 };
+use vertex_swarm_peer_manager::IpTrackerConfig;
+
+/// Default live per-IP concurrent-connection cap on the CLI; `0` is unlimited,
+/// the off-by-default state that mirrors `IpTrackerConfig::DEFAULT_MAX_CONNECTIONS_PER_IP`.
+const DEFAULT_MAX_PER_IP: usize = 0;
+
+/// Default distinct-overlay cap per IP for the identity-cycling detector.
+const DEFAULT_MAX_OVERLAYS_PER_IP: usize = IpTrackerConfig::DEFAULT_MAX_OVERLAYS_PER_IP;
 
 /// Validated peer management configuration.
 #[derive(Debug, Clone)]
@@ -13,6 +21,10 @@ pub struct PeerConfig {
     ban_threshold: f64,
     warn_threshold: f64,
     max_per_bin: usize,
+    /// Live per-IP concurrent-connection cap; `None` is unlimited.
+    max_connections_per_ip: Option<usize>,
+    /// Distinct-overlay cap per IP for the identity-cycling detector.
+    max_overlays_per_ip: usize,
 }
 
 impl Default for PeerConfig {
@@ -21,6 +33,9 @@ impl Default for PeerConfig {
             ban_threshold: DEFAULT_PEER_BAN_THRESHOLD,
             warn_threshold: DEFAULT_PEER_WARN_THRESHOLD,
             max_per_bin: DEFAULT_PEER_MAX_PER_BIN,
+            // Off by default: `0` means unlimited, matching the library default.
+            max_connections_per_ip: (DEFAULT_MAX_PER_IP != 0).then_some(DEFAULT_MAX_PER_IP),
+            max_overlays_per_ip: DEFAULT_MAX_OVERLAYS_PER_IP,
         }
     }
 }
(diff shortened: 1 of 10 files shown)


mfw78:
Closing as redundant: every part of this PR has been overtaken by `main`.

- The raised IP-tracker limits (`MAX_OVERLAYS_PER_IP`, `MAX_SIGHTINGS_PER_IP`, `--network.peer.max-overlays-per-ip`) edit `crates/swarm/peers/peer-manager/src/ip_tracker.rs`, which no longer exists: the whole per-IP identity-cycling tracker was removed in #563 (closing #517) as weak Sybil control that was decoupled from overlay grinding and had no reference-node equivalent. This PR can no longer merge against the deleted file.
- The dial-backoff change (do not arm backoff on a remote-induced early disconnect) is already on `main` via the merged early-disconnect work (#518): the penalty now fires only on a fast `RemoteClose` of a peer that did no useful work (`was_productive_since_connect`), and locally-initiated closes are distinguished through the `close_peer` intent map. That is a more precise resolution than the `LocalClose`-only rule here.
- The per-IP live connection cap shipped here inert (off by default, "pending a redesign") and runs against the deliberate decision to step away from per-IP control. If a connection-resource cap is still wanted it should be re-scoped as a focused DoS/admission control decoupled from reputation, not revived on top of the deleted IP-tracker.

See #347 for the issue-level close.

PR #358, opened 2026-06-16
feat(swarm-demo): in-browser Swarm upload/download client with live globe

### LINKED ISSUE #349: wasm: browser upload/download demo with live peer globe
## Problem

There is no end-to-end, in-browser proof that the wss (secure WebSocket) Swarm client path actually uploads and downloads.

The streaming upload and download RPCs and the wasm32 client cone exist in the stack, but nothing exercises them from a real browser, so regressions in the browser transport, discovery, or the upload/download flow can land unnoticed. We want a runnable demo that uploads and downloads a file from a browser tab and visualizes the peers the node is talking to.

## Proposed

A browser-native Swarm client demo (`bin/swarm-demo`) compiled to wasm32 and served as a static page.

- Connects over the wss transport through the existing `ClientLauncher`, reusing the self-throttling launcher default and the node's observed external address.
- Uploads a file end to end in the browser: nectar splits the content into chunks, builds the mantaray manifest, and issues postage stamps (issuer plus usage persistence), writing everything into a session-local in-memory chunk store so a just-uploaded file is immediately retrievable.
- Downloads a file by joining its leaf chunks, fetching any chunks not held locally from the network and reassembling them through the joiner.
- Renders a live globe (globe.gl) that plots connected peers by geo-locating their multiaddrs with a bundled offline geoip dataset, preferring the node's observed address for the local marker.

The demo deploys as a static wasm bundle (Trunk build), so it can be opened directly in a browser.

## Scope

- Affected: `bin/swarm-demo` (new wasm32 browser app and its static assets)
- Out of scope: a live wasm-backed peer feed for the globe (see Notes); it is left as a single, clearly marked integration point.

## Notes

Implementation constraints:

- All nectar crates (primitives, postage, postage-issuer, postage-usage, mantaray, contracts, swarms) are pinned to a single nectar revision so the demo and the vertex crates resolve one copy of the shared types (`ChunkAddress`, `Stamp`, `StampedChunk`); duplicate copies of these types break the chunk-provider and stamp plumbing.
- `nectar-postage-usage` is built with the `issuer` and `seal` features to get signed content stamps and the usage-persistence seal path; a browser clock facade (web-time) supplies the stamp timestamp.
- The globe path is offline: geoip lookup, multiaddr-to-IP parsing, and IPv6 expansion are implemented in JS with unit-tested parsing helpers, and the country polygons and geoip database are vendored as assets.

Known incomplete: the live peer feed (`WasmPeerFeed`) is stubbed; the globe runs off a `MockPeerFeed` by default (`?mock`), with `?live` reserved for the real wasm `SwarmDemo` handle once it exposes the peer-event drain. This is the one outstanding integration point and is marked as such in the source.

Alternatives considered:

- A native CLI-only proof (for example via dipper): useful, but it does not exercise the browser wss transport, the wasm32 build of the client cone, or the JS interop, which is exactly the path that was unproven.
- A headless wasm integration test: valuable as a follow-up, but it would not give a human-runnable demonstration of the upload/download flow, which is the stated goal here.

Verified: a topology-visualization predecessor already exists at `bin/swarm-demo` on `origin/main` (it dials mainnet bootnodes over wss and renders the Kademlia topology building up via the `SwarmDemo` handle's `readiness()` / `drainEvents()`), but it has no file upload/download and no globe. This PR extends that crate with the upload/download flow and the peer globe. Tracked by PR #358 (open).

mfw78:
## What does this PR do?
Adds an in-browser (wasm) Swarm client that demonstrates upload and download against the streaming RPCs over a WebSocket (wss) transport, with a live globe backdrop and a scoreboard of connected peers.

This layers purely additively on top of the IndexedDB node-cache and SWAP browser demo already on `main`: that demo provides connectivity, an IndexedDB-backed node chunk cache, and SWAP cheque settlement, and this PR adds the upload/download client tree (`src/client/`), the file UI, and the globe. The two caches coexist: `main`'s IndexedDB cache is the node's chunk store (survives a reload), and this PR's `client/cache.rs` is a session-local in-memory file-staging store for the upload/download path. The `main` SWAP and IndexedDB launcher wiring is kept untouched; the client surface is built over the launched node via `SwarmClient::from_launched`.

## Changes
- Add the upload/download client tree (`src/client/`): network/connection handling (`network.rs`, `net_get.rs`), chain access (`chain.rs`), upload/download flows (`upload.rs`, `download.rs`), a session-local file-staging cache (`cache.rs`), and a usage snapshot adapter (`usage.rs`) implementing the recover/flush model over the routing provider and sender.
- Add the file UI (`files_ui.rs`) and wire the globe feed: `lib.rs` publishes the running handle on `window.__swarmDemo`, exposes `self()` and `drainPeerEvents()`, and mounts the files panel below the topology view.
- Add the front-end assets: the peer feed (`app.js`, `peer-feed.js`) and vendored `globe.gl` plus country outline data for the globe backdrop; index/styles updates.
- Track the nectar deps the upload/download path needs (mantaray plus postage-usage `client`/`issuer`/`seal`) from the shared workspace nectar source.
- Single-threaded, stable-toolchain wasm build: no atomics, no shared memory, no `build-std`, no nightly. `.cargo/config.toml` and `Trunk.toml` reduce to the getrandom backend selection.

## Geolocation removed
This ships without IP geolocation. The roughly 18MB of vendored geoip databases and the geoip lookup code are not included, and the cross-origin-isolation service worker is removed. The globe renders the earth and country outlines as a backdrop; peers appear in the scoreboard and self panel without geo placement. The core deliverable, browser upload and download of Swarm content, is unaffected.

## Breaking changes
None. This adds to a self-contained binary crate and does not change existing crate APIs.

## Testing
- `nix develop --command cargo build --target wasm32-unknown-unknown` from `bin/swarm-demo` (its own workspace root) succeeds on the current `main` tip; `cargo fmt --all` is clean. The full threaded build-std link is nightly-only and runs in CI; the local gate is the stable-toolchain build.
- Manual testing: the browser client downloads a multi-megabyte blob and verifies its digest against the expected value.
- Known-incomplete items, stated plainly: the per-IP cap is off by default pending a redesign, and the browser bulk-download issue is still open.
- [x] wasm build passes

## Related issues
Closes #349

## AI Assistance
AI Assistance: Claude Code (Opus 4.8) used for rebase onto `main` and conflict resolution (the demo `Cargo.toml` and `lib.rs` two-file merge, and reconciling the client surface against the current `LaunchedClient`/`TopologyHandle` API), and for diagnosing the retrieval-driven neighbourhood collapse and implementing the forwarding-reliant retrieval fix in `client/network.rs`.

## Notes for reviewers
Worth a close look: the two-cache coexistence in `lib.rs` (`main`'s IndexedDB node cache plus this PR's session-local `client/cache.rs`), the wasm build recipe in `.cargo/config.toml`/`Trunk.toml` (single-threaded, no shared memory), the usage-snapshot adapter in `client/usage.rs`, and the upload/download streaming flows. The client surface now builds over the launched node's already-throttled `ClientHandle` (the builder wires the self-throttle), replacing the older per-peer inflight limiter that never landed on `main`.

Retrieval is forwarding-reliant. `client/network.rs` races the closest connected peers in widening waves and relies on the retrieval protocol forwarding the request toward the chunk's neighbourhood; it never dials on a retrieval. This removes the on-demand dial loop that previously collapsed the neighbourhood during a download: a single download opened roughly 355 WebSockets concentrated on the closest storers, draining bin-0 below the saturation threshold and dropping Kademlia depth from 3 to 0. With the fix, the WebSocket count stays bounded and depth holds while the download runs. The behavioural proof that forwarding works is that connected peers routinely serve chunks far from themselves (a peer sharing only a few bits with a chunk cannot hold it, so it forwarded). One residual far-chunk miss, when the connected set genuinely cannot reach a chunk, is the separate wide-last-resort gap (#345); it should be closed by the wide-race retrieval, not by reintroducing client dialing.

## Checklist
- [x] Self-review completed
- [x] Tests added/updated
- [x] PR title is descriptive


mfw78:
Rebased onto current `main` and brought up to the reshaped APIs. Summary of what changed since the last push:

**Rebase.** Replayed onto `main` (`a816a0c4`). Only 4 config/doc conflicts (`.cargo/config.toml`, `Trunk.toml`, `README.md`, `index.html`), resolved to this branch's superset; the Rust source auto-merged.

**Adapt the demo to nectar's MaybeSend redesign.** nectar's chunk-store traits became async (`ChunkGet`/`ChunkHas`/`ChunkPut` carrying `MaybeSend`), `file::sync_split` became `file::split`, `ChunkStoreError` gained `NotFound(ChunkAddress)` and a boxed `Other`, and the mantaray manifest ops became async. `MemoryCache` is now async, the split call is repointed, and `.await` is threaded through the prefetch and manifest paths. No behaviour change: the prefetch retry loop still converges (cache clones share the backing map) and `split` is byte-equivalent to the old `sync_split` (same BMT addressing, plain-mode only).

**Pin nectar by sha.** The floating `branch = "main"` nectar deps are now an explicit rev pin to `dd8778fce2d87ecddf1e7b3cf730e5e8e9808b13` (current nectar main) in both the workspace root and the demo manifest. Both must share the same source: cargo treats `?rev=X` and `?branch=main#X` as distinct sources, so a split pin compiles nectar twice and the shared types fail to unify. The bump is a no-op for the vertex crates, verified clean across the default, storer, and wasm cones plus the full `--all-features` test matrix.

**Verification.** `cargo check` clean on the workspace default cone, the storer cone (`-p vertex --features storer`), and the demo wasm target (`-p` swarm-demo, `wasm32-unknown-unknown`). The demo's PR gate is `cargo build --target wasm32` (the crate is `exclude`d from the root workspace, so workspace clippy does not run against it); the release wasm build passes. Pre-existing wasm-clippy lints in the demo (at lines untouched by this PR) remain and are tracked separately, not gated by CI.

AI Assistance: Claude Code (Opus 4.8) used for the rebase, the nectar API adaptation, the sha-pin, and the verification.

Issue #361, opened 2026-06-16
stream: add chunk-core parse_address/reconstruct/get_stream_from


mfw78:
## Status

`stale-fixed`. Implemented by PR #362 (merged 2026-06-16). `parse_address` (`crates/swarm/stream/src/lib.rs:135`) and `get_stream_from` (`crates/swarm/stream/src/lib.rs:328`) are now free functions in `vertex-swarm-stream`, with the `MaybeSendStream` marker imported and bounded at `lib.rs:33`/`:335`. The declared-chunk-type validation rides on `get_stream_from`, exercised by the `get_stream_from_rejects_wrong_chunk_for_address` test. The reconstruction helper landed as the nectar `StampedChunk::reconstruct` boundary call, now shared by both the FFI path (`crates/ffi/src/api/client.rs:392`) and the gRPC path (`crates/swarm/rpc/src/grpc/chunk.rs:76`).

## Problem

The FFI, wasm, and gRPC boundaries each carried their own near-identical copy of 32-byte address parsing and chunk reconstruction. The duplication had already drifted: the gRPC path parsed a proto `chunk_type` and then ignored it, so an upload whose bytes decoded to a different variant than declared was silently accepted. The inbound gRPC retrieve path also hand-rolled a `buffer_unordered` prefetch instead of reusing the verified-stream core.

## Proposed

Add shared chunk-core helpers to `vertex-swarm-stream`: a neutral `parse_address`, a `reconstruct` that wraps the primitives' address self-validation and additionally validates an optional declared chunk type, and a `get_stream_from` that sources the existing verify-by-default core from a `Stream` of addresses. A `MaybeSendStream` marker mirrors the existing `MaybeSendIter`.

## Notes

New free functions only; the declared-type check was added dormant until adopted at call sites.

Verified: the helpers are present on `origin/main` via merged PR #362.

mfw78:
Resolved on main by #362 (merged 2026-06-16): the shared chunk-core helpers ship in crates/swarm/stream/src/lib.rs - parse_address (lib.rs:135), get_stream_from (lib.rs:328), plus the reconstruct helper; StampedChunk::reconstruct is consumed by the FFI and gRPC chunk paths. Closing.

PR #362, opened 2026-06-16
feat(stream): add free parse_address and reconstruct helpers and get_stream_from

mfw78:
## What does this PR do?

Adds the shared chunk-core helpers (`parse_address`, `reconstruct`, `get_stream_from`) to `vertex-swarm-stream` so the FFI, wasm, and gRPC boundaries can later map onto one core instead of three near-identical copies.

## Changes

- Add `parse_address(&[u8]) -> Result<ChunkAddress, ParseAddressError>`: one neutral 32-byte address check.
- Add `reconstruct(expected, data, stamp, declared)`: wraps the primitives' address self-validation and additionally validates an optional declared chunk type against the decoded variant, closing the parsed-but-ignored proto `chunk_type` gap. The check is dormant until adopted later.
- Add `get_stream_from(provider, stream, cfg)`: a `Stream`-sourced sibling of `get_stream` reusing the same refill/verify core, plus a `MaybeSendStream` marker mirroring `MaybeSendIter`.

## Breaking changes

None. These are new free functions; existing call sites are unchanged.

## Testing

- [x] Unit tests pass
- [x] Documentation updated (if needed)

New tests mirror the existing `get_stream` coverage for `get_stream_from` and add parse-address and declared-type validation cases. Built and tested under the workspace toolchain with `protoc`.

## Related issues

Part of the client-core refactor stack.

Related: #361

## AI assistance disclosure

AI Assistance: Claude Code agents implemented this change end to end (code, commit message, and this PR description) under an orchestrated workflow that includes automated QA and security review gates. A human is accountable for the result and has reviewed it.

## Notes for reviewers

Base of this PR is the previous stack PR (r01). The helpers are additive; the consumer rewiring happens in r04 and r06.

## Checklist

- [x] Self-review completed
- [x] Tests added/updated
- [x] No console.logs or debug code left behind
- [x] PR title is descriptive


crates/swarm/stream/src/lib.rs
```diff
@@ -179,6 +181,88 @@ impl Default for StreamConfig {
...
+impl DeclaredChunkType {
+    /// Whether the decoded chunk matches the declared variant. A reconstructed
+    /// chunk is only ever [`AnyChunk::Content`] or [`AnyChunk::SingleOwner`].
+    fn matches(self, chunk: &AnyChunk) -> bool {
+        matches!(
+            (self, chunk),
+            (Self::Content, AnyChunk::Content(_)) | (Self::SingleOwner, AnyChunk::SingleOwner(_))
+        )
+    }
+}
```
mfw78: Shouldn't this actually most probably use the chunk types from nectar instead of redeclaring DeclaredChunkTypes here? In which case we can expand the protobuf chunk field to also include the chunk type as well if it isn't already?

Additionally, there is use of `reconstruct_chunk`, which really sounds ideal for something that should be moved into nectar and be in the primitives for chunks likely?

Issue #365, opened 2026-06-16
[FEATURE] Drive the gRPC chunk service off the chunk core


mfw78:
## What problem does this solve?

The gRPC chunk service carries its own address parser, its own reconstruct path that ignores the declared `chunk_type`, a hand-rolled `buffer_unordered` prefetch, and a streaming path that emits an empty `served_by`. It also defines a local `ChunkServiceProvider` trait that duplicates the shared chunk-client capability.

## Describe the solution

Rewire the gRPC chunk service onto the chunk core: alias `ChunkServiceProvider` to the shared `ChunkClient`, replace the local parse/reconstruct with thin status-mapping wrappers over the free helpers, route streaming `retrieve_chunks` through `get_stream_from`, emit the real `served_by`, and route unary/streamed retrieve and upload through the `get`/`put` defaults.

## Alternatives considered

Keeping the local copies was rejected: it perpetuates the silent `chunk_type` gap and the empty `served_by`.

## Implementation notes

This step is semver-visible: unary `RetrieveChunk` becomes verify-by-default, streaming `RetrieveChunks` now populates `served_by`, and upload reconstruct now rejects a chunk whose bytes decode to a variant other than the declared `chunk_type`.

## Are you willing to PR this?

Yes.

mfw78:
Closing: done. grpc/chunk.rs is driven off vertex_swarm_stream::ChunkClient (parse_chunk_address, StampedChunk::reconstruct, get/put, get_stream_from) and emits a real served_by; test retrieve_chunks_emits_served_by covers it (#382).

PR #366, opened 2026-06-16
refactor(rpc): drive gRPC chunk service off ChunkClientExt and the free helpers

mfw78:
## What does this PR do?

Rewires the gRPC chunk service onto the chunk core, absorbing the local `ChunkServiceProvider` trait and routing parse, reconstruct, retrieve, and upload through the shared helpers and `ChunkClientExt` verbs.

## Changes

- Re-export `ChunkServiceProvider` as an alias of `vertex_swarm_stream::ChunkClient`, so existing `P: ChunkServiceProvider` bounds keep compiling against the one capability alias.
- Replace the local `parse_chunk_address`/`parse_stamped_chunk` with thin status-mapping wrappers over the free `parse_address`/`reconstruct`; reconstruct now validates the declared `chunk_type` against the decoded variant.
- Route streaming `retrieve_chunks` through `get_stream_from` (one bounded, verify-by-default prefetch shared with the FFI and wasm paths) instead of the hand-rolled `buffer_unordered`; malformed addresses and inbound errors are interleaved as their own error items.
- `verified_response` emits the real `served_by` now carried by `VerifiedChunk` (the streaming path previously emitted it empty).
- Unary `retrieve_chunk` routes through the `get` default (verify-by-default); unary and streamed uploads route through the `put` default.

## Breaking changes

Yes, semver-visible:
- Unary `RetrieveChunk` is now verify-by-default, so a wrong-bytes delivery errors instead of returning.
- Streaming `RetrieveChunks` now populates `served_by` (previously empty).
- Upload reconstruct now rejects a chunk whose bytes decode to a variant other than the declared `chunk_type`.

These tighten correctness: they reject deliveries that the prior code silently accepted.

## Testing

- [x] Unit tests pass
- [x] Documentation updated (if needed)

Built and tested under the workspace toolchain with `protoc`. The streaming, unary, and upload paths are exercised by the gRPC service tests; the shared core paths carry their own coverage from r02 and r03.

## Related issues

Part of the client-core refactor stack.

Related: #365

## AI assistance disclosure

AI Assistance: Claude Code agents implemented this change end to end (code, commit message, and this PR description) under an orchestrated workflow that includes automated QA and security review gates. A human is accountable for the result and has reviewed it.

## Notes for reviewers

Base is r03. The three semver-visible behaviour changes in the Breaking changes section are the headline; please confirm they are acceptable for the gRPC surface. The bounded-error-channel fix for the new streaming path lands separately in r05.

## Checklist

- [x] Self-review completed
- [x] Tests added/updated
- [x] No console.logs or debug code left behind
- [x] PR title is descriptive

### DIFF
--- a/crates/swarm/rpc/src/grpc/chunk.rs
+++ b/crates/swarm/rpc/src/grpc/chunk.rs
@@ -6,7 +6,7 @@ use futures::StreamExt;
 use tonic::{Request, Response, Status, Streaming};
 use vertex_swarm_api::{ChunkAddress, PushReceipt, Stamp, StampedChunk, SwarmError};
 use vertex_swarm_stream::{
-    NATIVE_DOWNLOAD_CONCURRENCY, StreamConfig, VerifiedChunk, retrieve_verified,
+    ChunkClientExt, StreamConfig, VerifiedChunk, get_stream_from, parse_address,
 };
 
 use crate::ChunkServiceProvider;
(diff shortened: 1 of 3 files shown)


crates/swarm/rpc/src/grpc/chunk.rs
```diff
@@ -188,13 +176,13 @@ impl<P: ChunkServiceProvider> Chunk for ChunkService<P> {
...
-            self.provider.send_chunk_unchecked(stamped).await
-        }
-        .map_err(|e| Status::internal(format!("chunk upload failed: {e}")))?;
+        // Pre-stamped uploads are trusted by default; the `validate` flag opts
+        // into the stamp-signature check via the `put` default.
+        let receipt = self
+            .provider
+            .put(stamped, req.validate)
+            .await
+            .map_err(|e| Status::internal(format!("chunk upload failed: {e}")))?;
```
mfw78: Not sure about pre-trusting stamps in any context. It should be configurable likely to automatically trust pre-signed stamps, or to inspect them. This is the difference between a public vs private grpc endpoint. Then even on the private, it should be parameterised so that a private grpc can still an endpoint or a parameter, such as the validate request parameter as to whether or not it should be validated.

crates/swarm/rpc/src/lib.rs
```diff
@@ -7,17 +7,15 @@ mod grpc;
...
-}
+/// A chunk provider usable by the chunk gRPC service.
+///
+/// Absorbed into [`vertex_swarm_stream::ChunkClient`]: kept as a re-export so
+/// existing bounds (`P: ChunkServiceProvider`) keep compiling against the one
+/// capability alias. The chunk service drives off [`ChunkClientExt`] and the
+/// free helpers, so this no longer carries its own definition.
+///
+/// [`ChunkClientExt`]: vertex_swarm_stream::ChunkClientExt
+pub use vertex_swarm_stream::ChunkClient as ChunkServiceProvider;
```
mfw78, line 18: Flag potential killing of aliases for end of stack review

crates/swarm/stream/src/lib.rs
mfw78, line 428: There likely should be an option here, potentially threaded through from configuration, or at instantiation that forces everything through validation, per previous comment.

crates/swarm/rpc/src/grpc/chunk.rs
mfw78: Addressed. Added a server-side `StampValidation` policy (`Enforce` | `PerRequest`) on `ChunkService`. A public gRPC endpoint now defaults to `Enforce`: it always validates the postage stamp signature, ignoring a caller's `validate = false`. A private or trusted endpoint can opt into `PerRequest` to honour the per-request flag via `ChunkService::with_stamp_validation`. Both the unary and streaming upload handlers resolve through the policy. Embedded callers (FFI, wasm) trust the implementer and keep honouring their own flag. This is an intentional, security-positive behaviour change: public uploads now validate by default. The config and CLI thread-through (node arg to builder to `with_stamp_validation`) is tracked in #377.

crates/swarm/rpc/src/lib.rs
mfw78: Acknowledged. Keeping the `ChunkServiceProvider` re-export as the transitional shim for now to avoid a rename ripple through the middle of the stack. I will remove the alias and rename usages to `ChunkClient` as an end-of-stack cleanup, per your flag.

crates/swarm/stream/src/lib.rs
mfw78: Addressed together with the upload comment. Per your guidance that embedded callers (FFI, wasm) should trust the implementer by default while the gRPC endpoint verifies by default, the policy is specifically a gRPC-endpoint concern, so it lives on `ChunkService` (default `Enforce`) rather than on the core `put`, which stays the pure mechanism. The config thread-through is tracked in #377.

PR #379, opened 2026-06-17
refactor(swarm-api): reshape api component model to ClientComponents<T,C>

mfw78:
## What does this PR do?
First car of the node API-surface train (review the architecture in the train description). Reshapes the single component model so it carries the chunk client and is the one consumers use.

## Changes
- `vertex-swarm-api`: `NodeComponents<T, C>` now carries the chunk client `C` and implements `HasChunkClient` (the previous api model carried accounting and implemented no chunk-client accessor, which is why nothing used it). Generic over the topology type `T`, so api stays free of any transport or topology dependency.
- Tidy the role/accessor surface accordingly.

## Breaking changes
Internal (unreleased). Reshapes an api type that had no live consumers.

## Testing
`cargo check --workspace`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo test --workspace` green.
- [x] Unit tests pass

## Related issues
Part of the node API-surface revision train, based on #372.

## AI assistance disclosure
AI Assistance: implemented by Claude Code (Anthropic) agents under a workflow with per-step QA gates (fmt, clippy -D warnings, cargo check/test). Reviewed for correctness.

## Notes for reviewers
Bottom of the train. api gains the one component model; later cars move consumers onto it and relocate the gRPC adapter.

## Checklist
- [x] Self-review completed
- [x] Tests added/updated
- [x] PR title is descriptive


crates/swarm/api/src/components/mod.rs
```diff
@@ -93,26 +96,62 @@ impl<T: Send + Sync> HasTopology for BootnodeComponents<T> {
...
+}
+
+impl<T: Send + Sync, C: Send + Sync> HasChunkClient for NodeComponents<T, C> {
+    type ChunkClient = C;
+
+    fn chunk_client(&self) -> &C {
+        &self.chunk_client
+    }
+}
+
```
mfw78, line 133: OK, this seems a little weird in that a chunk client _requires_ accounting, because pushsync / retrieval all have accounting associated with them. I don't think this is being properly captured here.

crates/swarm/api/src/lib.rs
```diff
@@ -18,6 +18,7 @@
 //! Each capability level has a runtime container:
 //!
 //! - [`BootnodeComponents`] - topology only
+//! - [`NodeComponents`] - topology + chunk client (transport read surface)
 //! - [`ClientComponents`] - topology + accounting
 //! - [`StorerComponents`] - topology + accounting + store
 //!
```
mfw78, line 23: This is still a little weird in that there is no specific usage of topology + accounting only - except for the pricing smell that exists in bootnodes because pricing is required (bee nodes will drop nodes that don't implement pricing, before hive so gossip wouldn't work). In which case, that should be more the exception than the rule. It would be fine to define something that is very narrow to support this, that is easily deletable once that protocol smell has been resolved.

We should also file that as an issue in the protocol v1+ action items in this repo.

PR #388, opened 2026-06-17
feat(swarm-storer): thread the local overlay into Reserve for proximity eviction

mfw78:
## What does this PR do?

Threads the local overlay address into the storer `Reserve` so proximity-ranked eviction has the data it needs.

## Why

`EvictionStrategy::EvictFurthest` must rank chunks by XOR distance from the node's own overlay. The `Reserve` did not carry the overlay, so `EvictFurthest` fell back to `EvictOldest`. This adds the overlay; the furthest-eviction implementation itself lands with the reserve adapter (it needs the store's address enumeration).

## Changes

- `Reserve` gains an optional `overlay: OverlayAddress`, a `with_overlay` builder, and an `overlay()` accessor. The proximity-agnostic strategies leave it `None`.

## Breaking changes

None. Existing constructors are unchanged; the overlay defaults to `None`.

## Testing

- [x] Unit test covering the builder and accessor.
- [x] `cargo test -p vertex-swarm-storer` green; clippy clean.

## AI assistance disclosure

AI Assistance: a Claude Code agent made this change and wrote this description. A human is accountable for the result and has reviewed it.

### DIFF
--- a/crates/swarm/storer/src/reserve.rs
+++ b/crates/swarm/storer/src/reserve.rs
@@ -5,6 +5,8 @@
 
 use parking_lot::RwLock;
 use tracing::{debug, warn};
+use vertex_swarm_api::SwarmIdentity;
+use vertex_swarm_primitives::OverlayAddress;
 
 use crate::{ChunkStore, StorerError, StorerResult};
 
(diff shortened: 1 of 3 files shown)


crates/swarm/storer/src/reserve.rs
mfw78, line 1: But couldn't we just thread the identity instead which would contain the overlay address?

mfw78: Good call — done in 8291cf50. Replaced `with_overlay(OverlayAddress)` with `Reserve::with_identity(&impl SwarmIdentity)`, which derives the overlay via `SwarmIdentity::overlay_address()`. This matches how the rest of the node is wired off the identity (`ClientNode::builder(identity)`, accounting, topology all take `Arc<Identity>`).

I kept the reserve storing only the *derived* overlay address rather than holding the identity: the reserve needs its own location in address space for proximity ranking, not the identity's signing capability (receipt signing stays at the pushsync handler). So `with_identity` takes the identity (consistent, and it's self-sufficient — `overlay_address()` already folds in network id + nonce) and stores the `OverlayAddress` it yields.

PR #392, opened 2026-06-17
feat(swarm): storer pushsync ingest (is_responsible_for, store, sign receipt)

mfw78:
## What this PR does

Makes a storer take custody of inbound pushsync deliveries. Until now the inbound pushsync path was client-only: every delivery was forwarded to a closer peer and the storer's receipt relayed verbatim. This PR teaches a storer to recognise the deliveries it is responsible for, store them in its reserve, and acknowledge each one with its own freshly signed custody receipt. The flow is `is_responsible_for` then `put` then `Receipt::sign`.

The handler gains an optional `StorerCapability`. When it is absent (a client) every inbound pushsync keeps the unchanged verbatim-relay path; when it is present (a storer) a responsible delivery is stored and acknowledged, while a delivery the node is not responsible for still forwards. The receipt path reuses the same substream, so there is no second per-chunk network round trip.

## Where it sits

Stacked on `storer/s06-wiring` (#391), top of the train. It depends only on the stable observable surface of `ReserveStore` (`is_responsible_for`, `storage_radius`, `put`) that survived the per-stamped-entry reserve rework, so the protocol, handler, and receipt logic carry over without semantic change.

## Design

- `Receipt::sign(signer, network_id, address, nonce, storage_radius)` is added to `vertex-swarm-net-pushsync` as the storer-side counterpart to `Receipt::reconstruct`. It signs over the 32-byte chunk address only, then derives the storer overlay by recovering the Ethereum address from its own fresh signature and applying `compute_overlay` with the node's identity nonce, exactly as the read side does. Deriving the overlay from the signature (rather than asking the signer for its address) keeps the parameter a plain `SignerSync`, the same trait the node holds erased behind a trait object, and guarantees the minted receipt is self-consistent under `reconstruct`. The nonce-grinding loop that previously lived in the test-only `wire()` helper is no longer needed there; the helper now grinds only a test nonce and mints through the production `Receipt::sign`.

- A new `StorerCapability { reserve, signer, nonce }` (in `crates/swarm/node/src/protocol/storer.rs`) bundles the three things a storer needs on ingest: the `ReserveStore` (responsibility check, stamped `put`, declared storage radius), the receipt signing key erased to `Arc<dyn SignerSync + Send + Sync>`, and the identity nonce. It is `Clone` (Arc-cheap) and its `Debug` impl deliberately omits the signing key and reserve, showing only the public nonce. It is carried as an `Option<StorerCapability>` on `ClientBehaviour` and `ClientHandler`; the behaviour clones it into each handler at connection establishment. The handler constructor gains a `storer: Option<StorerCapability>` argument; `None` runs the unchanged client path.

- `ClientNode::enable_storage(reserve)` installs the capability, erasing the identity's signer to `dyn SignerSync` and reading the identity nonce. It must run during node assembly before the event loop accepts connections, for the same reason as `enable_forwarding`: a handler created earlier would not capture the capability. The builder calls it alongside `enable_forwarding`, passing the reserve view of the one `DbReserve`.

- The builder's `StoreFactory` now returns a `NodeStore { local, reserve }` rather than a bare `Arc<dyn SwarmLocalStore>`. A storer's local store is its reserve, but `Arc<dyn ReserveStore>` does not upcast to `Arc<dyn SwarmLocalStore>`, so both trait-object views of the single `DbReserve` are carried: `local` is what the retrieval handler and components share, and `reserve` is the proximity-axis view the pushsync ingest path needs. A client leaves `reserve` as `None`.

- Inbound dispatch (`on_pushsync_delivery`): if the capability is present and `reserve.is_responsible_for(address)`, the handler runs `store_and_sign`; otherwise it takes the existing forward path. Ordering is store, sign, send. The reserve `put` happens first so a receipt never claims custody of a chunk that is not durably stored. The chunk arrives as an always-stamped `StampedChunk` and is converted to a stamped `CachedChunk` for the always-stamped reserve. The receipt is signed with the reserve's current `storage_radius` (currently `StorageRadius::ZERO` until a radius manager lands; a declared 0 leans on the verifier's local floor). A `put` failure or a sign failure resets the substream without acknowledging. A send failure after a successful `put` also resets, but is a safe retry: the put is content-addressed, so a re-delivery is an idempotent no-op.

- A new `InboundOutcome::Stored` / `HandlerEvent::InboundStored` / `ClientEvent::InboundStored` path threads the successful-custody signal up to the service, which logs it and increments `swarm.client.inbound_stored`. `InboundPushFailed` now also covers store-or-acknowledge failures on the ingest path.

- Dependency move: `alloy-signer` becomes a normal (non-dev) dependency of the node crate, since the ingest path erases the identity's signer to `dyn SignerSync` in production.

## Consensus notes

Each custody receipt carries exactly one signed `PostageProof`: a single signature over the 32-byte chunk address, bound to the storer's identity nonce and declared storage radius. The storer overlay a forwarder recovers from a minted receipt is identical to the one the storer's own key and nonce derive, so `Receipt::sign` and `Receipt::reconstruct` are exact inverses across the wire boundary. A node never signs a receipt for a chunk it does not store: only a storer responsible for the address mints one.

## Testing

- pushsync (`receipt.rs`): `sign_mints_a_receipt_that_reconstructs_to_the_same_storer` checks the minted receipt's storer, address, nonce, and radius, then round-trips it through `to_wire` plus `Receipt::reconstruct` and asserts the reconstructed receipt is field-identical to the minted one. The existing depth-policy tests now mint through the production `Receipt::sign` via the updated `wire()` helper.
- node (`behaviour.rs`): two end-to-end tests drive the real libp2p handler against an in-memory `MockReserve` whose `is_responsible_for` is a fixed flag. `responsible_storer_stores_and_signs_a_receipt` asserts the returned receipt acknowledges the chunk, declares the storer's radius, recovers to the storer's own overlay, and that the chunk is durably in the reserve. `non_responsible_storer_forwards_instead_of_storing` asserts a not-responsible storer takes the forward path (which resets under the stub forwarder) and stores nothing. The `MockReserve` implements the reworked `ReserveStore` eviction primitives (`evict_from_bin`, `evict_batch`) for completeness; the ingest path exercises none of them.
- builder (`launch.rs`): the storer-factory test additionally asserts the factory yields the reserve view (`node_store.reserve.is_some()`); the real `DbReserve` path is covered here.

## Related issues

None.

## AI assistance disclosure

This change was prepared with the assistance of a Claude Code agent (storer-lattice design), including this description. A human author is accountable and has reviewed it.

### DIFF
--- a/crates/swarm/net/pushsync/src/receipt.rs
+++ b/crates/swarm/net/pushsync/src/receipt.rs
@@ -44,6 +44,7 @@
 //! `compute_overlay`); a separate issue tracks lifting it into nectar as a
 //! reusable primitive.
 
+use alloy_signer::SignerSync;
 use nectar_primitives::ChunkAddress;
 use vertex_swarm_primitives::{
     NeighborhoodDepth, NetworkId, Nonce, OverlayAddress, StorageRadius, compute_overlay,
@@ -51,6 +52,41 @@ use vertex_swarm_primitives::{
 
 use crate::codec::WireReceipt;
 use crate::error::PushsyncError;
+
+/// A node identity that can mint its own custody receipt.
+///
+/// [`Receipt::sign`] needs three things from the minting node, and they are not
+/// three independent parameters: they are all facets of the node's single
+/// identity. The signing key proves custody, and the `network_id` and `nonce`
+/// are the two inputs that, together with the recovered signing address, derive
+/// the storer overlay via [`compute_overlay`] (the same derivation a forwarder
+/// replays in [`Receipt::reconstruct`]). Threading them as loose arguments
+/// duplicated state the identity already owns, so they are bundled behind this
+/// one handle: the caller passes its identity and the receipt path reads the
+/// signer and the overlay-derivation inputs from it.
+///
+/// This trait is deliberately minimal and lives in the pushsync layer rather
+/// than referencing the node's full identity type: pushsync sits below the node
+/// and cannot depend on it. A node-layer identity implements this trait (or a
+/// node-owned capability bundle does), so the richer identity remains the single
+/// source of truth without leaking node concerns into this crate.
+pub trait ReceiptSigner {
+    /// The synchronous signing key. Erased trait objects
+    /// (`dyn SignerSync + Send + Sync`) satisfy this via `alloy-signer`'s blanket
+    /// impls, so a node can plug in its key without this crate being generic over
+    /// the concrete signer.
+    type Signer: alloy_signer::SignerSync + ?Sized;
+
+    /// The receipt signing key, signing over the 32-byte chunk address.
+    fn signer(&self) -> &Self::Signer;
+
+    /// The network id, one of the two overlay-derivation inputs.
+    fn network_id(&self) -> NetworkId;
+
+    /// The node's identity nonce, the other overlay-derivation input. It binds
+    /// the storer overlay but is not part of the signed message.
+    fn nonce(&self) -> Nonce;
+}
 
 /// Length in bytes of a well-formed recoverable signature.
 const SIGNATURE_LEN: usize = 65;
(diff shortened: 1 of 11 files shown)


crates/swarm/net/pushsync/src/receipt.rs
```diff
@@ -210,6 +210,53 @@ impl Receipt {
...
+    /// trait object — and guarantees the minted receipt is self-consistent under
+    /// recovery. The returned [`Receipt`] is therefore fully verified by
+    /// construction.
+    pub fn sign(
+        signer: &impl alloy_signer::SignerSync,
+        network_id: NetworkId,
+        address: ChunkAddress,
+        nonce: Nonce,
+        storage_radius: StorageRadius,
+    ) -> Result<Self, PushsyncError> {
```
mfw78, line 278: Shouldn't this just take my identity which should also have the ability to sign? Or something like this, why duplicate network_id / nonce, which is just needed for the overlay purposes.

mfw78 (commented):
In general this seems to be common this network_id / nonce / signer creeping that should just use the identity of the node

crates/swarm/net/pushsync/src/receipt.rs
mfw78: Done. `Receipt::sign` now takes an identity-shaped handle instead of loose `signer` plus `network_id` plus `nonce` parameters.

A minimal `ReceiptSigner` trait was added in `vertex-swarm-net-pushsync` (`receipt.rs`): it signs the chunk address and reports the two overlay-derivation inputs (`network_id` and `nonce`). The storer overlay is still derived from the recovered signing address via `compute_overlay`, so a minted receipt remains self-consistent under `reconstruct`. The signed bytes (the 32-byte chunk address) and the derived overlay are byte-identical to before; this is a pure plumbing change with no wire or consensus effect.

The trait lives in pushsync rather than taking `vertex-swarm-api`'s `SwarmIdentity` directly, because pushsync sits below `vertex-swarm-api` in the dependency graph and must not depend on it. At the node layer the capability bundle (`StorerCapability`) carries the node's identity (signer, `network_id`, `nonce`) and implements `ReceiptSigner`, so the handler hands its identity straight to `Receipt::sign`. `network_id` is now read from the identity's spec at storer setup rather than threaded separately through the inbound path.

The wider handshake creep you flagged (`SwarmPeer` signing in `swarm_peer.rs` and the synack encode path in `handshake/codec/synack.rs`) is out of scope for this PR and is tracked as a follow-up in #411.

This reply was drafted with AI assistance.

Issue #395, opened 2026-06-17
Redistribution primitives (#393) are not consensus-compatible with bee's reserve sampler + proof-of-entitlement


mfw78:
## Problem

The redistribution primitives added in #393 (`vertex-swarm-redistribution`: `sample`, `Entitlement`, and the sort half of `canonical_neighbourhood`) are **not consensus-compatible with bee's storage-incentives game**. A red-team verified this against the local bee source (`/code/nxm/swarm/bee` @ `21de6da9`: `pkg/storer/sample.go`, `pkg/storageincentives/proof.go`, `agent.go`) and the nectar commit vertex pins (`fc8fe992`). A vertex node running these would not reproduce bee's reserve sample or proof-of-entitlement and would fail the on-chain game when interacting with bee nodes.

This was surfaced by the benchmarking PR (#394), which deliberately ships **vertex-internal** conformance vectors (not bee-parity) and documents the gap.

## The three mismatches

1. **Reserve sampler.** vertex `sample()` is a Keccak-keyed Fisher-Yates over candidate *indices* (content-independent). bee `ReserveSample` (`SampleSize = 16`) iterates all reserve chunks at depth, computes a per-chunk **anchor-keyed prefix-BMT transformed address** (`Keccak256(anchor || span_LE || bmt_root)`), and keeps the 16 with the lexicographically smallest transformed addresses. Different selection rule.
2. **Proof of entitlement.** vertex `Entitlement` is one plain, un-keyed BMT inclusion proof for a single segment. bee `makeInclusionProofs` builds a reserve-commitment (RC) chunk (a CAC over the 16 sorted `ChunkAddress || TransformedAddress` items) and emits **three** witnesses derived from a second anchor (`require1 = anchor2 mod 15`, `require2 = anchor2 mod 14` bumped for distinctness, `require3 = 15`; `segmentIndex = anchor2 mod 128`), each with an RC-chunk inclusion proof plus a plain and an anchor1-prefix-keyed transformed BMT proof.
3. **Neighbourhood ordering.** The depth filter (`proximity >= depth`, `MAX_PO = 31`) *does* match bee. The full-256-bit XOR-distance sort is a vertex-internal canonicalisation with no bee counterpart (bee orders by transformed address).

## Parity roadmap (the rework)

[...]

PR #101, opened 2026-06-01
feat(primitives): typed Nonce, NetworkId, ProximityOrder, Bin, Timestamp

mfw78:
## Summary

Unit 1 of the bee-mainnet-interop bootnode plan. Introduces strict newtypes
that replace bare `[u8;32]` / `u64` / `u8` / `i64` flowing through public APIs.

- `Nonce(B256)` — opaque, `random()` (gated on `std`), `as_slice()`, no `Deref`.
- `NetworkId(u64)` — `MAINNET = 1`, `TESTNET = 10`, decimal Display, BE wire helpers.
- `ProximityOrder(u8)` — `TryFrom<u8>` rejects > 31 (`MAX_PO`), displays as `po=<n>`.
- `Bin(u8)` `#[repr(transparent)]` 0..=31, kept distinct from `ProximityOrder`.
- `Timestamp(i64)` — `now()`, `skew_window(Duration)` validator mirroring bee
  `bzz/timestamp.go`, structured `TimestampError` for metrics labelling.
- `SwarmAddressExt` extension trait adds typed `proximity_order`, `xor`, `bin`
  on `nectar_primitives::SwarmAddress` without touching nectar. Golden vectors
  from bee `proximity_test.go` are included.

New public enums (`ProximityOrderError`, `BinError`, `TimestampError`) are
`#[non_exhaustive]`. All types derive serde behind the existing `serde` feature.

## Test plan

- [x] `cargo build -p vertex-swarm-primitives --all-features`
- [x] `cargo test  -p vertex-swarm-primitives --all-features` (46 tests pass)
- [x] `cargo clippy -p vertex-swarm-primitives --lib --all-features -- -D warnings`
- [x] `cargo clippy -p vertex-swarm-primitives --tests --all-features -- -D warnings -A unwrap_used -A expect_used -A indexing_slicing`

Issue #99, opened 2026-06-01
CI: clippy failure on main due to useless `.into_iter()` in DialTracker


mfw78:
## Summary

CI is failing on `main` (commit e36b1ac) with a `clippy::useless_conversion` error in `vertex-net-dialer`, breaking the `clippy (production)` job and therefore the whole `CI` workflow.

## Failure

[Run 26744121746](https://github.com/nxm-rs/vertex/actions/runs/26744121746) — `clippy` job, `Clippy (production)` step:

```
error: explicit call to `.into_iter()` in function argument accepting `IntoIterator`
   --> crates/net/dialer/src/tracker.rs:185
    |
185 |             let opts = prepare::prepare_dial_opts(peer_id, addrs.into_iter(), &mut filter)
    |                                                            ^^^^^^^^^^^^^^^^^
    = note: `-D clippy::useless-conversion` implied by `-D warnings`
help: consider removing the `.into_iter()`

error: could not compile `vertex-net-dialer` (lib) due to 1 previous error
```

`prepare::prepare_dial_opts` already takes `addrs: impl IntoIterator<Item = Multiaddr>`, so the explicit `.into_iter()` is redundant.

## Fix

Drop the `.into_iter()` call in `crates/net/dialer/src/tracker.rs:185`.

PR #397, opened 2026-06-17
feat(redistribution): reserve sampler and proof of entitlement (#395)

mfw78:
## What does this PR do?

Implements vertex's reserve sampler and proof of entitlement for the Swarm storage-incentives redistribution game, so vertex computes the transformed addresses, reserve sample and inclusion proofs that `Redistribution.sol` verifies on chain. This is the consensus-conformance work for #395 (superseding the earlier vertex-internal `sample()`/`Entitlement` from #393).

The algorithm is the one fixed by the storage-incentives protocol and the contract's verification logic. The implementation is a pure function of its inputs (no I/O, no storage, no node, no async), so it produces identical results on every participating node, and it is validated against the canonical Swarm reference vectors.

## Changes

The crate is organised into cohesive modules (`anchor`, `neighbourhood`, `sample`, `witness`, `proof`), with `lib.rs` holding the crate docs, `SAMPLE_SIZE` and the public re-exports.

- Sampler (`sample.rs`):
  - The anchor-keyed transformed address is consumed from nectar (`AnyChunk::transformed_address`) rather than re-derived here; nectar owns that primitive and its conformance vectors.
  - `reserve_sample` keeps the 16 (`SampleSize`) chunks with the lexicographically smallest transformed addresses via a sorted insertion, including the exact equal-address tie-break that prefers the content-addressed chunk over a single-owner chunk.
  - `canonical_neighbourhood` is a pure committed-depth membership filter; it deliberately imposes no distance ordering, because the protocol orders the sample by transformed address, not the neighbourhood by distance.
- Proof of entitlement (`proof.rs`, `witness.rs`):
  - `make_inclusion_proofs` builds the reserve-commitment CAC over the 16 sorted `(chunk_address || transformed_address)` pairs (16 * 64 bytes) and, per witness, emits the RC-chunk inclusion proof, a plain (OG) BMT segment proof and an anchor-prefixed (TR) BMT segment proof.
  - The witness slots are named for their role in the game: `WitnessIndices { challenged: [usize; 2], last, segment_index }`. The two `challenged` slots are the ones the claim anchor pseudo-randomly selects; `last` is the final slot, whose maximum transformed address anchors the reserve-size estimate.
  - The claim anchor is decoded big-endian to select those slots and the segment index; BMT spans stay little-endian `u64`, and the two are kept strictly separate.
- Strongly-typed, role-based API surface:
  - `SampleAnchor` and `ClaimAnchor` are distinct `B256` newtypes for the two successive round salts, so the two cannot be transposed (a swap would be a compile error rather than a silently round-losing bug).
  - `CommittedDepth(Bin)` is the on-chain-derived sample boundary, kept distinct from the routing-layer `NeighborhoodDepth` (different role: per-round, on-chain, reserve-commitment, not local connectivity).
  - `SampleItem` carries the typed `AnyChunk`, so the proof paths re-derive both BMT proofs without re-parsing raw bytes or branching on a chunk-type flag.

## Conformance evidence

All values are asserted against the canonical Swarm reference vectors, byte for byte (`tests/conformance.rs`):

- Single-CAC vector (anchor `swarm-test-anchor-deterministic!`, 4 KiB pattern chunk):
  - chunk address `902406053a7a2f3a17f16097e1d0b4b6a4abeae6b84968f5503ae621f9522e16`
  - transformed address `9dee91d1ed794460474ffc942996bd713176731db4581a3c6470fe9862905a60`
- Full sample and inclusion proofs (`tests/fixtures/inclusion_proofs.json`, `anchor1 = 100`, `anchor2 = 30`), a mixed CAC/SOC sample so both chunk types are exercised end to end. The tests check:
  - the transformed address of all 16 sample items under the sample anchor;
  - the reserve-sample ordering (`reserve_sample` of the shuffled items reproduces the sorted order);
  - the reserve-commitment (sample) chunk address `b012904b0c3e6462158b4416556caa888031a79bad46d2ffa7012408c9c38aa8`;
  - the witness indices `require1 = 0`, `require2 = 3`, `require3 = 15`, `segment_index = 30`;
  - every `proofSegments`/`proveSegment`/`proofSegments2`/`proveSegment2`/`chunkSpan`/`proofSegments3` field of all three inclusion proofs.
- The witnessed proofs also self-verify against the relevant BMT roots as a secondary guard.

## Scope

In scope: the transformed-address ordering, the reserve sample, the reserve-commitment chunk and the three-witness inclusion proofs (RC, OG and TR), all confirmed against the reference vectors.

Out of scope (and therefore not claimed as conformant here): the neighbourhood enumeration over a live reserve (this code is fed by the storer), postage-stamp signature construction and the on-chain submission encoding.

## Dependencies

The workspace nectar dependencies track `main`, pinned by `Cargo.lock` to a commit that includes the per-node prefix BMT (nectar #98), the anchor-keyed transformed address and SOC unwrap (nectar #99), and the removal of the vestigial `AnyChunk::Custom` variant (nectar #100). No temporary branch pin remains.

A follow-up (nectar #101, the `Header + BmtBody` chunk-model refactor) is the tracked home for a public borrowed-body view; until then this crate keeps a small private zero-copy `Body` view, which reads only the already-public `AnyChunk` accessors.

## Breaking changes

No public API breakage outside the redistribution crate. `SampleItem` and the entitlement types were introduced in #393 and have no external consumers yet.

## Testing

- [x] `cargo test -p vertex-swarm-redistribution` passes (16 tests: 8 unit, 7 conformance, 1 doctest).
- [x] `cargo check --workspace --all-targets`, `cargo clippy -p vertex-swarm-redistribution --all-targets` under `-D warnings` (clean) and `cargo fmt --all -- --check` pass.
- [x] `rustdoc` builds with `-D rustdoc::broken_intra_doc_links` (clean).

## Related issues

Refs #395

## AI assistance disclosure

Claude Code (Anthropic) was used to implement the sampler and proof of entitlement, to design the role-typed API, and to write the conformance tests. A multi-agent review pass over the API aesthetics, performance, idiomacy and consensus-conformance safety was also run. All output was reviewed by the author and validated against the canonical Swarm reference vectors.

## Notes for reviewers

- The headline check is `inclusion_proofs_match_reference_byte_for_byte` in `tests/conformance.rs`.
- `SampleAnchor`/`ClaimAnchor` are intentionally distinct types; please keep them un-transposable.

## Checklist
- [x] Self-review completed
- [x] Tests added/updated
- [x] PR title is descriptive


crates/swarm/redistribution/src/redistribution.rs
```diff
@@ -1,57 +1,76 @@
...
+pub enum ProofError {
+    /// The sample did not contain exactly [`SAMPLE_SIZE`] items.
+    #[error("reserve sample must have {SAMPLE_SIZE} items, got {0}")]
+    SampleSize(usize),
+    /// The sample-time anchor (`anchor1`) was empty.
+    #[error("anchor1 is not set")]
+    MissingAnchor1,
+    /// The claim-time anchor (`anchor2`) was empty.
+    #[error("anchor2 is not set")]
+    MissingAnchor2,
```
mfw78: couldn't this just be flattened into `MissingAnchor` and make it a compound type so that a u8 there could be used to describe which anchor was missing?

crates/swarm/redistribution/src/redistribution.rs
```diff
@@ -72,290 +91,477 @@ pub fn canonical_neighbourhood(
     depth: u8,
```
mfw78: Is this really what we should be doing here? Don't we have a primitive that accurately describes this specific `depth` item? Like is it StorageRadius / NeighbourhoodDepth etc?

crates/swarm/redistribution/src/redistribution.rs
```diff
@@ -72,290 +91,477 @@ pub fn canonical_neighbourhood(
...
+#[derive(Clone, Debug)]
+pub struct ChunkInclusionProof {
+    /// Reserve-commitment chunk inclusion proof (OG of the RC chunk).
+    pub rc_proof: Proof,
+    /// Plain (original) BMT segment proof over the witnessed chunk body.
+    pub og_proof: Proof,
+    /// Anchor-prefixed (transformed) BMT segment proof over the same body.
+    pub tr_proof: Proof,
+    /// The little-endian `u64` span of the witnessed chunk body (`chunkSpan`).
+    pub chunk_span: u64,
```
mfw78: Also, isn't span actually a primitive elsewhere?

crates/swarm/redistribution/src/redistribution.rs
```diff
@@ -72,290 +91,477 @@ pub fn canonical_neighbourhood(
...
+    }
+    acc as u64
+}
+
+/// BMT-hash a `span`/`payload` body, applying the `anchor1` prefix when given.
+///
+/// Factors out the repeated `DefaultHasher` ritual: construct (plain or
+/// `with_prefix`), `set_span`, `update`. Used for the reserve-commitment chunk
+/// (no anchor) and the OG (no anchor) / TR (anchor) witness proofs.
+fn rc_hasher(anchor: Option<&[u8]>, span: u64, payload: &[u8]) -> DefaultHasher {
```
mfw78: Isn't span and payload essentially a `BmtBody`?

crates/swarm/redistribution/src/redistribution.rs
```diff
@@ -72,290 +91,477 @@ pub fn canonical_neighbourhood(
...
+///
+/// # Errors
+///
+/// Returns an error if `items` does not contain exactly [`SAMPLE_SIZE`]
+/// elements, if either anchor is empty, or if any underlying BMT proof
+/// generation fails (e.g. an out-of-range segment index).
+pub fn make_inclusion_proofs(
+    items: &[SampleItem],
+    anchor1: &[u8],
+    anchor2: &[u8],
```
mfw78: Dont' just use arbitrary bytes slices, these are recipes for typing issues. What are these types meant to be? Shouldn't we perhaps have an `Anchor` type?

crates/swarm/redistribution/src/redistribution.rs
mfw78: Resolved by removing them entirely. The anchors are now fixed `bytes32` (`B256`), which are always valid, so there is nothing to report — `AnchorError` is gone. The role distinction lives in two distinct types (`SampleAnchor`/`ClaimAnchor`), which is stronger than a runtime `u8` discriminator (see the L371 reply). A `bytes32` cannot be empty, so the original compound-variant suggestion is moot.

crates/swarm/redistribution/src/redistribution.rs
mfw78: Now `CommittedDepth(Bin)`. I added a dedicated redistribution type rather than reusing `NeighborhoodDepth`: its contract scopes it to the local connectivity boundary (supply-side, local-only), whereas the redistribution committed depth is a per-round, on-chain-derived reserve-commitment value, so reuse would be a category error. The filter is `depth.contains(Bin::from(anchor.proximity(addr)))`; `CommittedDepth: TryFrom<u8>` for edge ergonomics.

crates/swarm/redistribution/src/redistribution.rs
mfw78: Checked — span is a bare `u64` by workspace convention (`Hasher::set_span`, `Proof.span`), so a span newtype would go against the grain and `chunk_span: u64` stays. The actual smell here — decomposing a chunk into separate span+payload — is fixed by the `Body` view (L333).

crates/swarm/redistribution/src/redistribution.rs
mfw78: Right — nectar already has a `BmtBody`, but it is crate-private (the `bmt_body` module is not re-exported), so it cannot be reused here. Added a private zero-alloc `Body<'a> { span, payload }` borrowing view with a `hasher(prefix)` method; both the reserve-commitment content and the per-chunk paths go through it. It replaces the old `rc_hasher(Option<&[u8]>, u64, &[u8])` helper.

crates/swarm/redistribution/src/redistribution.rs
mfw78: Two fixed-width types: `SampleAnchor(B256)` (anchor1) and `ClaimAnchor(B256)` (anchor2). Fixed `bytes32` because the on-chain `currentRoundAnchor` is always 32 bytes (verified: the contract stores `bytes32 currentRevealRoundAnchor`; bee's `ReserveSalt` unpacks a `[32]byte`). Distinct types so the two round salts cannot be transposed (a swap = garbage proofs = a lost round). The original `&[u8]` over-fit to the minimal 1-byte anchors the in-tree reference *test* uses (`big.Int.Bytes()`); I regenerated the parity oracle from bee with production-shaped 32-byte anchors, so this now proves parity against the real shape — `inclusion_proofs_match_bee_byte_for_byte` still passes.

crates/swarm/redistribution/src/redistribution.rs
```diff
@@ -1,361 +1,747 @@
...
+///
+/// Private borrowing view that exists only to stop decomposing chunks into bare
+/// `(span, payload)` pairs at internal boundaries. `span` stays a `u64` (the
+/// workspace convention: `Hasher::set_span` and `Proof.span` are both bare
+/// `u64`); what this type removes is the *pairing* smell, by reading both halves
+/// from a single typed source.
+struct Body<'a> {
+    span: u64,
+    payload: &'a [u8],
+}
```
mfw78: Can we just file a PR against the nectar primitives so that we can get borrows for bodies like this so that we don't have to copy? Then make use of references to avoid copies / allocs?

crates/swarm/redistribution/src/redistribution.rs
```diff
@@ -1,361 +1,747 @@
...
+/// # Errors
+///
+/// Returns an error if `items` does not contain exactly [`SAMPLE_SIZE`]
+/// elements, or if any underlying BMT proof generation fails (e.g. an
+/// out-of-range segment index). The anchors are non-empty by construction, so
+/// the function never has to check for an unset salt.
+pub fn make_inclusion_proofs(
+    items: &[SampleItem],
+    anchor1: SampleAnchor,
+    anchor2: ClaimAnchor,
```
mfw78: Instead of using `anchor1` and `anchor2` throughout, can we have more idiomatic naming, that reflects their actual roles?

crates/swarm/redistribution/src/redistribution.rs
```diff
@@ -1,361 +1,747 @@
...
+    fn hasher(&self, prefix: Option<&[u8]>) -> DefaultHasher {
+        let mut hasher = match prefix {
+            Some(p) => DefaultHasher::with_prefix(p),
+            None => DefaultHasher::new(),
+        };
+        hasher.set_span(self.span);
+        hasher.update(self.payload);
+        hasher
+    }
+}
```
mfw78: This really looks like something that should be in nectar primitives. Like an auto impl on borrowed, or similar so that we can get what we need borrowed, and not duplicate it needless (in memory, and/or in code)

crates/swarm/redistribution/src/redistribution.rs
mfw78: Done in 40593569 — params and docs now use role-based names matching the types: `SampleItem::new(sample: SampleAnchor, ..)`, `witness_indices(claim: ClaimAnchor)`, `make_inclusion_proofs(items, sample, claim)`. The WitnessIndices formulae read `claim mod N`. The bee/contract mapping is kept once per type doc ("the reference's `anchor1`/`anchor2`"), and the captured oracle fixture (`Oracle { anchor1, anchor2 }` + the JSON) keeps `anchor1`/`anchor2` since those are the reference wire field names.

crates/swarm/redistribution/src/redistribution.rs
mfw78: Not on this PR — I spun a multi-agent architectural review on where the body-borrow belongs, and the conclusion is to defer. The zero-copy body path vertex needs is already mostly shipped in nectar (`ContentChunk::body()`/`SingleOwnerChunk::inner_body()` -> `&BmtBody`, plus `BmtBody::transformed_root`), and this private `Body` isn't blocking anything (it reads the already-public `AnyChunk::span()`/`data()`). The proper home for a public borrowed-body view is the deferred nectar `Header + BmtBody` chunk-model refactor (#101), which unifies address derivation and lets `AnyChunk` impl `Chunk`. First step already landed: dropped the vestigial `AnyChunk::Custom` (nectar #100). Vertex keeps this `Body` until that refactor provides the upstream abstraction. Tracked in nectar #101.

crates/swarm/redistribution/src/redistribution.rs
mfw78: Agreed it belongs in nectar, not duplicated here. Same resolution as the L464 thread: deferred to the nectar `Header + BmtBody` refactor (#101), where a public borrowed-body view (and the Hasher-from-body convenience) land alongside the trait work. The analysis also found the "auto impl on borrowed" (blanket `impl<T: Trait + ?Sized> Trait for &T`) is coherent but unnecessary there — a plain borrowed-view type/method wins because there is no heterogeneous/generic call site exercising it. Until then vertex keeps its private `Body` (zero-copy, works at the pinned commit).

PR #400, opened 2026-06-19
feat(swarm-postage): add DbBatchStore and the on-chain event ingest seam

mfw78:
## What this PR does

Adds the standalone `vertex-swarm-postage` crate, the node-side home for postage state. The crate does not redefine any postage primitive: the batch model, the stamp model, validation and the event vocabulary all live in `nectar-postage` and are re-exported verbatim from this crate's root, so every downstream vertex crate resolves a single canonical copy of `Batch`, `Stamp`, `BatchEvent`, `PostageContext`, the `BatchStore` and `BatchEventHandler` traits, and the validators.

What this crate adds is the node's own concrete persistence plus the seam through which on-chain batch state is driven into it:

- `DbBatchStore<DB>`: a `BatchStore` over the `vertex-storage` `Database` (redb in production), persisting the batch set and the postage context across restarts.
- `DbBatchStore<DB>` also implements `BatchEventHandler`, the on-chain event ingest seam an external postage indexer drives.

It also wires the new crate into the workspace (`Cargo.toml` members and dependency table, plus `Cargo.lock`).

## Where it sits

Stacked on `stack/01-grpc-chunk-streaming` (PR #390 has merged). This is the first car (car A) of the postage, reserve and sampler train: it lands the postage persistence and ingest seam that the reserve and sampler cars build on.

## nectar dependency

The `nectar-*` git dependencies track nectar `branch = "main"`. The synchronous `BatchStore`, `BatchStoreExt`, `StoreValidator` and `SnapshotStore` conversion has landed on main via nectar #103, so the train builds against main with no further pin.

## Design

`DbBatchStore<DB: Database>` is generic over the storage backend, so the same code serves an in-memory database in tests and an on-disk redb database in production. It is cheap to clone by `Arc` and is thread-safe for concurrent reads and writes. It defines two tables:

- `Batches` (`postage_batches`): `BatchId -> Batch`, the authoritative batch set, keyed by a `BatchIdKey` newtype that carries the 32-byte big-endian encoding of the `B256` id. The newtype exists only to carry the storage `Encode`/`Decode` codecs across the orphan rule for the foreign key type.
- `ContextTable` (`postage_context`): a single-row table holding the current `PostageContext`. redb has no schema-level singleton, so the context is modelled as one row under a fixed `ContextKey::SINGLETON` address.

**BatchStore.** `DbBatchStore` implements the synchronous nectar `BatchStore` directly: each method is a plain function over a single transaction, with no `async move { result }` facade and no executor in sight. (The earlier `get_sync` / `put_sync` / `remove_sync` shims are gone; only the genuine read-modify-write `mutate_sync` helper remains, which keeps the load-mutate-store logic in one place for the event handler.) Because redb is itself synchronous, no transaction guard is ever held across an `await` point, the futures stay `Send`-free on the wasm path, and the trait is naturally object-safe (it has an associated `Error` and no generic methods). `batch_ids` walks the lazy read-only cursor, whose read snapshot is detached from the borrowed transaction handle. A fresh database needs no separate initialisation step: `new` ensures both tables exist.

**BatchEventHandler ingest seam.** `DbBatchStore` also implements `BatchEventHandler`, the synchronous handler an external postage indexer drives (this seam was already synchronous and is unchanged in shape). The four `BatchEvent` variants map onto store mutations:

- `Created` writes the fully-described batch directly.
- `TopUp` and `DepthIncrease` load, mutate and store inside one `db.update` transaction via `mutate_sync`, closing the lost-update window a read-then-write pair would leave open.
- `Expired` removes the batch through the named `acknowledge_expired` seam.

Each mutation is a single transaction, so the store never observes a partial event application, and each variant is idempotent for an unknown batch.

Expiry is deliberately split. `acknowledge_expired` is the bare store removal only; it does not, and cannot from this crate, evict the reserve entries stamped under the batch. It is documented as the acknowledgement half of the reserve's evict-then-acknowledge expiry path (`ExpirySweep::on_expired_event`, a later car), so live `Expired` wiring (#391/#392) must route through the reserve's evict-before-remove path rather than removing the batch directly. The other three variants have no reserve coupling and are safe to drive directly.

**nectar reuse.** Stamp validation is stateless and reuses the nectar `StoreValidator` / `Stamp` path: a stamp is valid if and only if `ecrecover(sig, digest)` resolves to the batch owner and the index and bucket bounds hold. The owner is already known from the batch store, so no per-batch public-key cache is needed and none is shipped; the cheaper recover-once path nectar supports is noted as a possible future in-process optimisation only.

**Indexer / ABI stub.** The left-hand half of the decode path (the event ABI, the log decoder, and the indexer that reduces chain logs into `BatchEvent`s) is intentionally stubbed and owned externally. This crate ships no contract bindings and no log-to-event reducer, only the documented `BatchEventHandler` seam, so the chain indexer can evolve without touching the persistence layer and this crate can be tested with synthetic events and no chain.

## Testing

The crate's tests (against an in-memory redb backend, plus one on-disk reopen) cover:

- CRUD round-trip including `put` upsert semantics, `contains`, `get`, `count` and `remove`.
- The context singleton: default-on-fresh, set, and overwrite.
- `batch_ids` returning all keys in cursor (ascending big-endian) order.
- Persistence across a database create-then-reopen on a real file.
- All four `BatchEvent` variants via `handle_event`, including the unknown-batch no-ops for top-up and expiry.
- The single-transaction mutate path (successive top-up then depth-increase with no lost write).
- `acknowledge_expired` idempotency (the second acknowledgement is a no-op).
- The `BatchIdKey` codec round-trip and its decode rejection of wrong-length input.

A cross-crate marker test for the remove-before-evict orphan hazard is `#[ignore]`d here, since the orphan-safe ordering is enforced and regression-tested in the reserve crate where both halves are in scope.

## Related issues

Refs the storer and reserve effort. `contains` performs a full read pending a key-presence primitive, marked `TODO(#214)` in code to match the existing storer pattern. `batch_ids` uses the lazy cursor from #396. Live expiry ingest wiring is tracked under #391/#392.

## AI assistance disclosure

This change was prepared with the assistance of Claude Code. A human author reviewed it and is accountable for the contribution. The implementation follows the agreed schema and the nectar public API.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4905,7 +4905,7 @@ checksum = "27b02d87554356db9e9a873add8782d4ea6e3e58ea071a9adb9a2e8ddb884a8b"
 [[package]]
 name = "nectar-contracts"
 version = "0.2.1"
-source = "git+https://github.com/nxm-rs/nectar?branch=main#9bca9affe11de422b682d663c15ebc6f2300e87a"
+source = "git+https://github.com/nxm-rs/nectar?branch=main#0c191183ffd6458d529891487e1c8badcdd7b3e9"
 dependencies = [
  "alloy-primitives",
  "alloy-sol-types",
@@ -4915,7 +4915,7 @@ dependencies = [
 [[package]]
 name = "nectar-postage"
 version = "0.2.1"
-source = "git+https://github.com/nxm-rs/nectar?branch=main#9bca9affe11de422b682d663c15ebc6f2300e87a"
+source = "git+https://github.com/nxm-rs/nectar?branch=main#0c191183ffd6458d529891487e1c8badcdd7b3e9"
 dependencies = [
  "alloy-primitives",
  "alloy-signer",
@@ -4930,7 +4930,7 @@ dependencies = [
 [[package]]
 name = "nectar-primitives"
 version = "0.2.1"
-source = "git+https://github.com/nxm-rs/nectar?branch=main#9bca9affe11de422b682d663c15ebc6f2300e87a"
+source = "git+https://github.com/nxm-rs/nectar?branch=main#0c191183ffd6458d529891487e1c8badcdd7b3e9"
 dependencies = [
  "alloy-primitives",
  "alloy-signer",
@@ -4958,7 +4958,7 @@ dependencies = [
 [[package]]
 name = "nectar-swarms"
 version = "0.2.1"
-source = "git+https://github.com/nxm-rs/nectar?branch=main#9bca9affe11de422b682d663c15ebc6f2300e87a"
+source = "git+https://github.com/nxm-rs/nectar?branch=main#0c191183ffd6458d529891487e1c8badcdd7b3e9"
 dependencies = [
  "alloy-chains",
  "num_enum",
@@ -9292,6 +9292,19 @@ dependencies = [
  "strum 0.28.0",
  "vertex-net-peer-score",
  "vertex-swarm-api",
+]
+
+[[package]]
+name = "vertex-swarm-postage"
+version = "0.1.0"
+dependencies = [
+ "alloy-primitives",
+ "nectar-postage",
+ "serde",
+ "tempfile",
+ "thiserror 2.0.18",
+ "vertex-storage",
+ "vertex-storage-redb",
 ]
 
 [[package]]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -92,6 +92,9 @@ members = [
     # Storer (chunk storage and staking)
     "crates/swarm/storer",
 
+    # Postage (batch store + on-chain event ingest seam)
+    "crates/swarm/postage",
+
     # RPC
     "crates/rpc/core",
     "crates/rpc/server",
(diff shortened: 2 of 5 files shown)


crates/swarm/postage/src/snapshot.rs
mfw78, line 1: Is this actually required at all? To be clear, the snapshot providers I do believe are there for the nectar-postage-usage, which is a CLIENT SIDE methodology that is used to embed / preserve postage usage within the same batch. This is not required from a node perspective at all.

mfw78: You are right, and this has been actioned. `SnapshotStore` is an issuer / client-side abstraction: it recovers the per-bucket issuer counters for stamp *issuance* and is consumed by `nectar-postage-usage`. A verifying node does not need it, and the node-side `StoreValidator` never used it. Misusing the generic `SnapshotStore<S>` trait as a per-batch `VerifyingKey` cache was a category error.

The misuse has been removed entirely from the train. `snapshot.rs` and `InMemorySnapshotStore` are gone from `vertex-swarm-postage` (PR-A here), the `SnapshotStore` re-export is dropped, and the `parking_lot` dependency that backed the removed mutex map has been removed. `DbBatchStore` now holds only its database handle, no public-key cache. The same removal lands in PR #403, where the `AdmissionValidator` is now **stateless**: it holds only the confirmation threshold and validates each stamp by recovering the signer and comparing it to `batch.owner` (reusing the nectar `StoreValidator` / `Stamp::verify` path, ecrecover to `batch.owner`, with index and bucket bounds). The batch owner is already known from the batch store, so no side state is required. Validation semantics are unchanged: a stamp is valid iff `ecrecover(sig, digest)` resolves to `batch.owner` and the bounds hold. Only the mis-modelled memoisation was dropped.

For the record, nectar does support a cheaper path (recover the signer once with `Stamp::recover_pubkey`, then `verify_with_pubkey` the remaining stamps of the same batch). That is deferred as a possible *future* optimisation. If profiling ever justifies it, it must be a per-process in-memory memoisation gated behind the batch-existence check, and never a persisted sidecar table. It is intentionally not implemented now, and a doc note in `lib.rs` records this so the decision is not lost.

This change was prepared with the assistance of an AI coding agent. A human reviewed and is accountable for the contribution.

Issue #399, opened 2026-06-19
redistribution: parallelise and SIMD-batch stamp validation in sampling


mfw78:
## Problem

Stamp validation on the paths that do verify (proof-of-entitlement witness construction, any future on-demand validation) recovers one signer per stamp serially. A reserve sample carries many stamps, so a serial `ecrecover` per stamp is a measurable cost on the storer's hot redistribution path.

## Current behaviour

The original framing (avoid redundant per-chunk `ecrecover` in the sampler) is already realised on main: the reserve sampler enumerates trusted, already-ingested reserve entries via the validate-on-ingest path (`crates/swarm/redistribution/src/filter.rs`) and performs no per-candidate `ecrecover` in the hot path. That cost trap is designed out.

What remains is serial: `Stamp` validation recovers the signer one stamp at a time. `crates/swarm/postage/src/admission.rs` `validate` calls `stamp.verify(address, batch.owner())`, a single serial signer recovery (`ecrecover`) per stamp, with no parallel or batched signature-recovery path and no benchmark exercising real stamp validation.

## Proposed

- Parallelise stamp validation across the sample set rather than validating serially.
- Use batched / SIMD secp256k1 signature recovery where a batch of stamps must be verified together, instead of one `ecrecover` per stamp.
- Add a benchmark that exercises real stamp validation so the parallel/SIMD path can be measured against the serial baseline.

## Scope

- Affected: `crates/swarm/redistribution`, `crates/swarm/postage`

## Notes

Perf-only and non-consensus: it changes how fast stamps are validated, not which chunks are admitted or sampled.

Verified on `origin/main`: no `par_iter`/`rayon`/SIMD/batched-recovery path exists under `crates/swarm/redistribution` or `crates/swarm/postage`; validation runs through the single-signer `admission.rs` `validate`.

mfw78:
Companion issue: nxm-rs/nectar#102 (batched SIMD keccak for the BMT hasher). These are the two halves of the reserve-sampling hot path: ecrecover (here) and keccak (there).

PR #403, opened 2026-06-19
feat(swarm-storer): rework the reserve to a per-stamped-entry six-table store

mfw78:
## What this PR does

This pull request reworks the storer's authoritative reserve from the address-keyed, first-stamp-wins design into a per-stamped-entry, six-table redb store. The reserve's *size* now counts distinct stamped entries (distinct `(batchID, stampIndex, address)` tuples), not distinct content addresses. That per-entry count is the quantity that drives the consensus-committed storage radius, so the rework changes what the reserve fundamentally measures.

A single content chunk may be stored under several postage batches at once (each batch is a separate stamped entry sharing one refcounted payload), and a slot within a batch may be re-stamped (a newer stamp for the same `(batchID, stampIndex)` supersedes the older one). Ingest is now validated synchronously through a stateless postage admission validator, and every mutation arbitrates the stamp against its slot and writes all affected rows in one atomic transaction.

The new files are `crates/swarm/postage/src/admission.rs` (the `AdmissionValidator`) and the reworked `crates/swarm/storer/src/db_reserve.rs`.

## Where it sits

Stacked on `feat/postage-stampindex-arbiter` (#401), which owns the stamp-index arbiter slot and the newest-wins `decide` function this reserve reuses. This is the third car (car D) of the postage, reserve and sampler train, and depends on its parent being merged first.

Because the underlying nectar `BatchStore` is now synchronous (nectar #103), the reserve calls it directly: there is no `block_on` bridge and no `batch_sync` shim. The earlier `futures::executor::block_on` bridge (`storer/src/batch_sync.rs`, the `BatchStoreSyncExt` and `BatchStoreSyncError` types) is deleted, and the async-future `+ Send` colouring on `DbReserve` is dropped. The residual `Send + Sync` on the `SwarmLocalStore` / `ReserveStore` / `BinCursorStore` impls is the `SwarmLocalStore: Send + Sync` threading contract, not async colouring.

### The six tables

All compound keys are big-endian, so the encoded byte order equals the `(field, field, ...)` lexicographic order. Every mutation writes or compacts all affected rows inside one `db.update` transaction, so a stamped entry and its index rows commit atomically and a removal never leaves a dangling row (no tombstones).

- **`Payload`**: `addr -> (refcnt, typed_bytes)`. The refcounted, content-addressed chunk body (the type-tagged `AnyChunk` bytes, without a stamp, since stamps differ per entry). One row per distinct address, present if and only if at least one stamped entry references it; the refcount is the number of such entries.
- **`Entry`**: `(po, batch, stampHash, addr) -> EntryValue { bin, binid, stamp_bytes }`. One row per stamped entry; the reserve size is this table's count. Keyed proximity-major, so the furthest entry (smallest po) is the table's first key and a proximity bin's entries are contiguous. The value carries the bin sequence and the precise admitting stamp, so `get` returns the chunk with a real stamp and an inclusion proof can carry exactly that stamp.
- **`BatchGroup`**: `(batch, po, addr, stampHash) -> ()`. Groups every entry by batch, then bin, then address, so a whole batch or a batch's shallow bins can be evicted with one prefix cursor scan.
- **`Replay`**: `(bin, binid) -> ReplayValue { addr, batch, stampHash, chunk_type }`. The append-only per-bin insertion-order projection a redistribution/sync consumer replays without rehydrating the body. `chunk_type` lets the sampler resolve the content-versus-single-owner tie without a body read.
- **`BinCounter`**: `bin -> u64`. The per-bin monotonically increasing insertion sequence; never rewound on eviction, so sequences are never reused (sync resumability).
- **`StampIndexTable`**: `(batch, stampIndex_be8) -> (timestamp, stampHash, addr)`. The newest-timestamp-wins arbiter slot, keyed by the full eight-byte `(batchID, stampIndex)`. This is *the* postage stamp-index table: rather than re-declaring a second `table!` handle (free to drift in name or codec and silently desynchronise the on-disk slot), the reserve imports the single handle the postage crate (#401) owns. The reserve only changes *when* it is written: it runs the same `decide` arbitration inside its own put transaction so admission and the six-table write commit together.

### Multi-batch coexistence and per-entry size

The same content under N batches leaves N `Entry` rows and one `Payload` at refcount N. A second batch storing already-present content bumps the refcount and rewrites no body; evicting one of several entries decrements the refcount and keeps the body, so partial eviction never drops a payload another entry still needs. The reserve size is the `Entry` count (paid postage slots), so a chunk held under N batches contributes N to size, not one.

### Stateless admission validator

`AdmissionValidator` (postage crate) is sync and stateless apart from the confirmation threshold it enforces. It holds no per-batch key cache and no sidecar table. On each `put` the reserve loads the referenced batch from the `BatchStore`, reads the live `PostageContext`, and the validator checks, in order: the batch is usable (enough confirmations) and not expired, the stamp index and bucket bounds hold, and the signature recovers the batch owner. The signer check is nectar's `Stamp::verify` (an EIP-191 ecrecover compared against `batch.owner()`), reused verbatim; the digest and ecrecover are never re-implemented here. Because the owner is already known from the batch store, no side state is required. A per-process in-memory memoisation of the recovered key (recover-once, verify-the-rest) is noted in the code as a possible future optimisation but is not implemented; if ever added it must never be a persisted sidecar table.

### Typed errors, no String funnel

The reserve calls the synchronous `BatchStore` directly and maps its associated `BatchStore::Error` straight into `SwarmError::storage`, preserving the typed error rather than funnelling through a `String`. The old `block_on` bridge and its ad-hoc `BatchStoreSyncError` wrapper are gone.

### Put, restamp, and eviction

`put` validates on ingest, then in one transaction reads the `(batch, stampIndex)` slot and runs `decide` (newest-wins; equal-or-older is rejected). On admission it assigns the next bin sequence, writes the `Replay`, `Entry`, and `BatchGroup` rows, and bumps or inserts the refcounted payload. On a restamp (the slot held an older stamp) it first deletes the four index rows of the displaced entry and decrements its payload before writing the incoming entry; the displaced rows are deleted using the *displaced* address's own proximity order, since a restamp may re-point the slot at a different content address.

Eviction (`evict_furthest`, `evict_from_bin`, `evict_batch`) operates per entry: it removes an entry's index rows and the stamp-index slot it owns, decrements the shared payload, and drops the body only when the last entry referencing it goes. `evict_batch` documents the evict-then-remove ordering seam: an expiry handler must drain a batch's entries through `evict_batch(batch, None, ..)` before the batch is removed from the postage `BatchStore`, or its entries would be orphaned and inflate the committed radius. Enforcement of that ordering lands in the following car.

### Supporting changes

- The `ReserveStore` trait now documents the deliberate coincidence of `Bin` and `ProximityOrder` in a reserve, and the eviction seam crosses that boundary through named conversions rather than an unchecked `u8` pun.
- `IncomingStamp::entry` is made public so the reserve can write the arbiter slot inside its own transaction.
- `Reserve::set_count` is added so the per-entry size can be seeded from the persisted `Entry` table count at construction.

## Consensus notes

The reserve size is the per-stamped-entry count, and it is the quantity that drives the consensus-committed storage radius and committed depth, samples the redistribution game, and lets an inclusion proof carry the precise stamp a sample slot was won with, none of which an address-keyed store can represent. A chunk stored under N batches counts as N paid slots but stores exactly one refcounted body. Newest-wins arbitration is keyed on the full eight-byte `(batchID, stampIndex)` slot, and the read, `decide`, and write all run inside one redb transaction so the slot cannot change between the read and the conditional write. The exact admitting stamp is surfaced byte-for-byte by `get`.

## Testing

The consensus-load-bearing behaviours are covered by executable tests, each over an in-memory redb database with real batches in a batch store and real stamps signed by an `alloy-signer-local` wallet, so the validate-on-ingest admission path runs as in production:

- `reserve_size_counts_stamped_entries_not_addresses`: the same content under N batches leaves the size at N with one shared payload at refcount N.
- `newest_timestamp_wins_full_index_keying`: newest-wins over the full `(batchID, stampIndex)` slot, with equal-timestamp and older presentations rejected and a distinct index in the same bucket admitted as a separate slot.
- `refcounted_payload_survives_partial_eviction`: no body rewrite on a second batch, and the body dropped only when the last referencing entry goes.
- `restamp_to_different_address_cleans_displaced_rows`: a restamp onto a different content address leaves no orphaned `Entry`, `BatchGroup`, or `Replay` rows and no leaked payload refcount.
- `get_returns_exact_admitting_stamp`: the admitting stamp is surfaced byte-for-byte.
- `removal_fully_compacts_all_tables_no_tombstones`: full compaction of all six tables on removal, with the arbiter slot cleared.
- `bin_scan_replays_entries_in_insertion_order`: the per-bin `Replay` projection is returned in insertion order.
- `unknown_batch_and_stampless_puts_are_rejected`: ingest validation rejects stamps for unknown batches.

Key-codec ordering is pinned by `entry_key_round_trips_and_orders_proximity_major`, `batch_group_key_orders_batch_major_then_bin`, `replay_key_orders_bin_major_then_sequence`, and `bin_key_round_trips`. The admission validator's own tests (`admits_a_well_formed_stamp`, `rejects_a_foreign_signer`, `rejects_an_unusable_then_an_expired_batch`) cover the stateless per-stamp checks.

`sample_collapses_duplicate_transformed_address` is `#[ignore]`d: the sample-at-most-once collapse over transformed addresses is a sampler property, not a reserve one. The reserve supplies the per-entry, chunk-typed `Replay` projection the sampler consumes; the collapse itself belongs to the sampler car.

## Related issues

- #75: storer node components (storage, chunk sync, redistribution).
- #395: redistribution primitives not consensus-compatible with bee's reserve sampler and proof-of-entitlement; the per-entry reserve and the precise admitting stamp are prerequisites for that compatibility.
- #399: avoiding redundant per-chunk ecrecover in reserve sampling; the deferred per-process signer memoisation noted above is the in-reserve counterpart.

## AI assistance disclosure

This change was prepared with the assistance of an AI coding agent. The agent implemented the per-stamped-entry schema, the stateless admission validator, the direct synchronous batch-store calls, and the executable consensus coverage described above. All of it is submitted for human review.


crates/swarm/postage/src/admission.rs
```diff
@@ -0,0 +1,264 @@
...
+///
+/// The batch is loaded by the caller (the reserve holds a
+/// [`BatchStore`](nectar_postage::BatchStore) and reads it before entering its
+/// write transaction); passing it in keeps this validator free of any store
+/// dependency and lets it run inside a synchronous redb transaction.
+#[derive(Debug)]
+pub struct AdmissionValidator {
+    /// Minimum block confirmations for a batch to be usable.
+    confirmation_threshold: u64,
+}
```
mfw78, line 93: can file an issue for later that this could be done optimistically and only commit batches once the confirmed block is marked finalized.

crates/swarm/storer/src/db_reserve.rs
```diff
@@ -1,49 +1,83 @@
...
+//! `Database`.
 //!
-//! [`DbReserve`] is the storer's authoritative reserve: an always-stamped chunk
-//! store that adds the proximity axis ([`ReserveStore`]) and the per-bin
-//! insertion-order axis ([`BinCursorStore`]) on top of point access
-//! ([`SwarmLocalStore`]). It owns a [`DbChunkStore`] for the chunk values, a
-//! [`Reserve`] for the capacity counter and the local overlay, and five
-//! hand-maintained secondary tables:
+//! [`DbReserve`] is the storer's authoritative reserve: a bee-faithful,
+//! per-stamped-entry chunk store. It is the reworked successor of the original
```
mfw78: kill references to bee

crates/swarm/storer/src/db_reserve.rs
```diff
@@ -1,49 +1,83 @@
...
-//! - [`AddrIndex`]: `ChunkAddress -> BinSeqKey`, the reverse lookup from an
-//!   address to its `(bin, seq)` so a removal can compact the insertion-order
-//!   row (no tombstones, bounded growth).
-//! - [`BatchIndex`]: `(batch_id, bin, addr) -> ()`, grouping every chunk by its
-//!   postage batch (then bin, then address) so a whole batch, or a batch's
-//!   shallow bins, can be evicted ([`evict_batch`](ReserveStore::evict_batch))
-//!   with a single prefix cursor scan.
+//! - [`BinCounter`]: `bin -> u64`. The per-bin monotonically increasing
+//!   insertion sequence (the bin cursor). Never rewound on eviction, so
+//!   sequences are never reused (sync resumability).
```
mfw78: What happens when the bin counter reaches u64::MAX?

crates/swarm/storer/src/db_reserve.rs
```diff
@@ -58,57 +92,76 @@ use vertex_storage::{
...
 
-// Per-bin insertion-order index: `(Bin, u64) -> BinSeqEntry`.
+// Refcounted content payload: `addr -> (refcnt, typed_bytes)`.
+//
+// One row per distinct *address* (not per entry). The body is the type-tagged
+// `AnyChunk` encoding, shared by every stamped entry of that content; the
+// refcount is the number of live entries referencing it. Present iff refcnt >=
+// 1. Uncompressed: chunk bodies are arbitrary/encrypted, so compression costs
+// CPU without saving space.
+table!(pub(crate) Payload, "reserve_payload", ChunkAddress, PayloadValue, compressed = false);
```
mfw78: I think we can actually optimistically compress here as making the assumption that it doesn't save space is premature. This could actually be a configuration option (and allow porting from uncompressed to compressed for analysis purposes)

crates/swarm/storer/src/db_reserve.rs
mfw78, line 1: This file is simply too large and should likely be broken down into a db submodule with associated modules so that its not a huge monolith

crates/swarm/postage/src/admission.rs
mfw78: Good idea. I have filed https://github.com/nxm-rs/vertex/issues/406 to track this: admit optimistically once the batch's confirmation block is observed, and only commit the entries as durable once that block is marked finalised. It needs a provisional-then-finalised lifecycle and a hook on the chain follower's finality signal, so it is a cross-cutting follow-up rather than a change for this PR. Resolving this thread and tracking it there.

This reply was prepared with AI assistance and reviewed by me.

crates/swarm/storer/src/db_reserve.rs
mfw78: Done. The "bee-faithful" wording is gone: the module doc now describes the reserve as the storer's authoritative, per-stamped-entry chunk store, with no reference to bee. I grepped the whole reserve module and there are no other bee mentions.

This reply was prepared with AI assistance and reviewed by me.

crates/swarm/storer/src/db_reserve.rs
mfw78: At u64::MAX the previous `+ 1` would panic in debug and silently wrap to zero in release, reusing sequence 0 and overwriting the live Replay and Entry rows at binid 0. That breaks the "never rewound, never reused" cursor invariant and would corrupt a sync or redistribution consumer's resume point with a non-monotonic sequence, so a silent wrap is the worst outcome here.

In practice the counter is per bin and is bumped once per admitted or restamped put, never on eviction, so its value is the lifetime admission count for that single bin. Reaching u64::MAX needs more than 1.8e19 puts into one bin, which is roughly 585,000 years at a million puts per second, so it is unreachable in any node lifetime. Given that, saturating would be wrong (it would pin and then reuse the top sequence), so I have documented the invariant at the increment site and switched to `checked_add(1)`, surfacing the impossible case as a hard `DatabaseError` through the existing transaction error path rather than wrapping. No new error variant, no hot-path cost, behaviour identical for every reachable input.

This reply was prepared with AI assistance and reviewed by me.

crates/swarm/storer/src/db_reserve.rs
mfw78: Good point, and I agree the "compression will not save space" rationale is an untested assumption that is worth measuring rather than baking in.

The blocker for doing it here is that compression is not a runtime property: it is `Table::COMPRESS_VALUES`, a compile-time `const bool` selected by the `table!` macro and consumed by monomorphisation in the redb codec (`put`/`decode_value`). Turning it into a configuration option means threading a runtime flag through the generic read and write path instead of reading the trait const, and a clean uncompressed-to-compressed migration additionally needs an on-disk framing or version marker per value, since today a stored value is a raw byte string with nothing to tell a decoder whether it is zstd-framed. Both are storage-layer changes rather than reserve changes, and out of scope for this refactor PR.

I have filed https://github.com/nxm-rs/vertex/issues/407 to capture exactly this: config-selectable compression for the `Payload` table plus an uncompressed-to-compressed migration path for analysis. I have reworded the table's rationale to note the assumption is untested and tracked there. Resolving this thread.

This reply was prepared with AI assistance and reviewed by me.

crates/swarm/storer/src/db_reserve.rs
mfw78: Agreed, this file had grown into a monolith. I have broken it into a `db_reserve` submodule with cohesive associated modules, a pure structural refactor with identical behaviour:

- `db_reserve/mod.rs`: the module documentation, the shared `EvictTarget` identity, and the re-exports that keep the public surface intact (`pub use store::DbReserve`).
- `db_reserve/schema.rs`: the six-table `table!` definitions and every key and value codec (`BinKey`, `EntryKey`, `BatchGroupKey`, `ReplayKey`, and the value records), alongside the pure key-ordering tests that pin them.
- `db_reserve/tx.rs`: the per-entry transaction primitives (payload refcount bump and decrement, the entry-row and full-entry deletes, and the small decode and error helpers).
- `db_reserve/store.rs`: the `DbReserve` struct, its constructor and accessors, and the `SwarmLocalStore`, `ReserveStore`, and `BinCursorStore` implementations plus the bin-scan iterator.
- `db_reserve/consensus_spec.rs`: the consensus spec tests, moved verbatim.

The public API is unchanged: `DbReserve` is re-exported from the same `crate::db_reserve::DbReserve` and `crate::DbReserve` paths, so `lib.rs` and the downstream cars compile without edits. The consensus spec tests move verbatim and continue to pass. The only adjustments beyond the move are the small in-place fixes from the other threads on this PR (the wording cleanup, the bin-counter overflow note, and the payload-compression note).

This change was prepared with AI assistance and reviewed by me before submission.

PR #442, opened 2026-06-21
fix(swarm-bandwidth-pseudosettle): bound first-contact allowance and clamp over-acks

### LINKED ISSUE #435: fix(accounting-pseudosettle): first-contact allowance bypass grants unbounded free credit (C1/M6)
## What

Pseudosettle grants unbounded free soft credit on first contact (and after every reconnect), defeating the time-based allowance that is the only anti-free-ride brake in soft accounting.

In `PseudosettleService::calculate_acceptable` the elapsed interval is seeded from `now` when there is no prior `last_settlement` entry for the peer:

```rust
let elapsed = self.last_settlement.get(peer).map_or(now, |&last| now.saturating_sub(last));
let allowance = self.refresh_rate.checked_scale(elapsed).unwrap_or(Au::from_amount(u64::MAX));
```

On first contact `elapsed = now` (a Unix timestamp, ~1.75e9), so `refresh_rate * elapsed` overflows `i64` and saturates to `Au::MAX`. The accepted amount is then bounded only by `owed` and `requested`, i.e. the first pseudosettle from a peer (and the first after any disconnect that clears the in-memory `last_settlement`) is accepted up to the entire debt with no rate limit.

A related defect in the same path: on a `Sent` ack the service credits the raw acked amount (`au_from_wire(ack.amount)`) even when the peer acks more than we offered; it reports a violation but still applies the inflated credit (M6).

## Why

Pseudosettle moves no real money; the `refresh_rate * elapsed` cap IS the mechanism that prevents infinite free service. Bypassed on first contact and on every reconnect, a peer can drain to the disconnect threshold, send one settle, and repeat at zero real cost. This is the core free-riding hole and makes the chainless pseudosettle-default v1 unsafe on the live network.

## How to fix

- Seed `elapsed` from connection time, never `now`: no prior entry should mean ~0 elapsed, not "the whole epoch". Persist/seed `last_settlement` to connect time on first sight.
- Hard-cap the per-grant allowance to `refresh_rate * (now - connect_time)` and never to `Au::MAX`.
- Credit `min(ack.amount, offered)` on the `Sent` path; an ack must never exceed the offer.
- Add regression tests: first-contact grant is bounded; reconnect does not reset the cap; over-ack is clamped.

Found in the accounting/settlement red-team. Pre-existing; not introduced by the planned accounting reshape, but gates mainnet-safe v1.

mfw78:
## What

Bound the pseudosettle time-based allowance to genuine elapsed wall-clock, and clamp an over-ack to the amount we offered.

`PseudosettleService::calculate_acceptable` seeded the elapsed interval from `now` when a peer had no prior `last_settlement` entry, so on first contact (and after any reconnect that cleared the in-memory state) `refresh_rate * now` overflowed and saturated to `Au::MAX`, removing the rate cap that is the only brake on free soft credit. A new `first_seen` anchor records when we began accounting for a peer; the allowance now accrues from the last settlement, or from `first_seen`, never from the absolute timestamp. The existing-`last_settlement` path is unchanged.

The `Sent` ack path also credited the raw acked amount even when a peer acked more than we offered. It now credits `min(ack, offered)`; the violation is still reported.

## Why

Closes #435. Pseudosettle moves no money, so the `refresh_rate * elapsed` cap is the anti-free-ride mechanism. Bypassed on first contact and every reconnect, a peer could drain to the disconnect threshold, send one settle, and repeat at zero cost. This gates a mainnet-safe v1 (pseudosettle is the default chainless settlement). Pre-existing bug, found in the accounting and settlement red-team.

## Testing

`cargo test -p vertex-swarm-bandwidth-pseudosettle` (19 passed). `cargo clippy -p vertex-swarm-bandwidth-pseudosettle --all-targets -- -D warnings` clean. New regression tests: first-contact grant bounded by elapsed, reconnect does not reset to unbounded, over-ack clamped to offer.

## AI Assistance

Claude Code (Opus 4.8) used for the red-team analysis that found the bug and for implementing the fix and tests, under human review.

### DIFF
--- a/crates/swarm/bandwidth/pseudosettle/src/service.rs
+++ b/crates/swarm/bandwidth/pseudosettle/src/service.rs
@@ -55,6 +55,13 @@ pub struct PseudosettleService<A: SwarmBandwidthAccounting> {
     pending: HashMap<OverlayAddress, PendingSettlement>,
     /// Track last settlement time per peer (for rate limiting).
     last_settlement: HashMap<OverlayAddress, u64>,
+    /// First time we started accounting for a peer's inbound settlements.
+    ///
+    /// The time-based allowance accrues from this point, never from the Unix
+    /// epoch. On first contact, and after a reconnect that cleared
+    /// `last_settlement`, the grant is bounded by the genuine wall-clock elapsed
+    /// since this instant, not by the absolute timestamp.
+    first_seen: HashMap<OverlayAddress, u64>,
     /// Optional reporter feeding settlement violations into peer scoring.
     reporter: Option<Arc<dyn PeerReporter>>,
 }
@@ -76,6 +83,7 @@ impl<A: SwarmBandwidthAccounting + 'static> PseudosettleService<A> {
             refresh_rate,
             pending: HashMap::new(),
             last_settlement: HashMap::new(),
+            first_seen: HashMap::new(),
             reporter: None,
         }
     }
@@ -192,7 +200,10 @@ impl<A: SwarmBandwidthAccounting + 'static> PseudosettleService<A> {
                         self.report_violation(&peer);
                     }
 
-                    let accepted = au_from_wire(ack.amount);
+                    // Credit at most what we offered: an ack can never accept
+                    // more than the offer, so a violating over-ack is clamped to
+                    // the offer rather than inflating our credited balance.
+                    let accepted = au_from_wire(ack.amount).min(pending.amount);
 
                     // Credit our balance (we paid, debt reduced)
                     let handle = self.accounting.for_peer(peer);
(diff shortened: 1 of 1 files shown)


crates/swarm/bandwidth/pseudosettle/src/service.rs
```diff
@@ -265,15 +281,21 @@ impl<A: SwarmBandwidthAccounting + 'static> PseudosettleService<A> {
...
+        // below still bound the result.
         let now = current_timestamp();
-        let elapsed = self
+        let since = self
             .last_settlement
             .get(peer)
-            .map_or(now, |&last| now.saturating_sub(last));
+            .or_else(|| self.first_seen.get(peer))
+            .copied()
+            .unwrap_or(now);
```
mfw78, line 297: But doesn't this `unwrap_or(now)` allow this? If this is effectively moot, perhaps look at another method to do this without having this here to prevent potential footguns

mfw78: Good catch. The fallback yields `elapsed = now - now = 0` (no grant), so it is safe in effect, but I agree the `now` anchor is a footgun smell. Tracked as a non-blocking follow-up in #444 to enforce a present anchor by construction. Merging this now per discussion.

Issue #438, opened 2026-06-21
swap: complete on-chain cashout, deploy, credit-on-cash and solvency


mfw78:
## Problem

SWAP settlement exchanges, validates and signs cheques with bounded uncashed exposure, but the monetary parts that close the loop on chain are not complete. Without confirmed on-chain cashing, a peer's debt is only ever bounded by the uncashed-exposure guardrail, not actually cleared, and chequebook deploy and solvency checks are missing.

## Current behaviour

On-chain cashout encoding exists and is wired: `crates/swarm/accounting/swap/src/cashout.rs` (`Cashout::cash`) is invoked from the settlement path (`crates/swarm/accounting/swap/src/service.rs:424`), and the chequebook chain client deploys and cashes (`crates/swarm/accounting/chequebook/src/chain.rs`, `deploy`/`cash_cheque_beneficiary`). Deploy is threaded from config (`crates/swarm/builder/src/launch.rs:353`, `deploy: params.swap.deploy`).

Credit-on-cash is still absent: debt is bounded by the per-peer uncashed exposure cap (`DEFAULT_BOUNCE_LIMIT` in `crates/swarm/accounting/swap/src/constants.rs`, ten times the payment threshold), whose rustdoc still reads "Bounds free service while on-chain cashing is stubbed." There is no solvency/liquidity check at accept time and no exchange-rate/deduction handling on the swap headers.

## Proposed

- Credit-on-cash: only reduce a peer's debt once cashing is confirmed (supersedes the v1 bounded-exposure guardrail), with a chequebook solvency/liquidity check at accept time.
- On-chain cashout: redeem received cheques against the issuer chequebook contract, paying out to our beneficiary; replace the best-effort behaviour with confirmed settlement.
- Chequebook deploy: complete `--swap.deploy` end to end.
- Exchange-rate and deduction handling on the swap stream headers; cheque value = `amount*rate + deduction`.

## Scope

- Affected: `crates/swarm/accounting/swap`, `crates/swarm/accounting/chequebook`, `crates/swarm/builder`
- Out of scope: the accounting/settlement reshape itself, which has landed.

## Notes

Wire: keep the cheque body byte-identical to the live network (JSON-in-protobuf, PascalCase keys, decimal `CumulativePayout` as u256, base64 signature, EIP-712 domain `Chequebook`/`1.0`, chain id 100); migrate to protobuf alongside the upstream migration tracked in #183.

[...]

PR #445, opened 2026-06-21
fix(swarm-bandwidth-swap): bound uncashed cheque exposure and canonicalise signatures

### LINKED ISSUE #437: fix(accounting-swap): bound per-peer uncashed cheque exposure (C2/H5)
## What

Received cheques reduce a peer's debt on structural validation alone (beneficiary, issuer signature, strictly-increasing cumulative payout), while on-chain cashing is stubbed for v1. Nothing bounds how much uncashed cheque value we treat as settled, so a counterparty can pay for unlimited real service with cheques that may never cash (a worthless or under-funded chequebook), and our books treat the debt as cleared.

## Why

Crediting uncashed cheques without a bound is theft-of-service: the issuer gets real bytes for cheques that bounce. Until cashout is real (tracked separately), exposure must be bounded.

## How to fix (v1 guardrail: bounded exposure)

- Track per-peer uncashed cheque exposure (cumulative received payout not yet confirmed cashed).
- Cap it at a configurable per-peer bounce limit, mirroring bee's bounce-tolerance behaviour; refuse to credit beyond the limit and stop serving the peer on real (originated) debt once exposure is maxed.
- Canonicalise cheque signatures at accept time (reject non-canonical `v`, enforce low-s / EIP-2) so a cheque cannot change identity before cash (H5).
- Tests: exposure accumulates, is capped, and blocks further credit; signature canonicalisation rejects malleated cheques.

The real fix (credit only on confirmed on-chain cash, with a chequebook solvency check) lands with the on-chain swap work; this issue is the v1 bounded-exposure guardrail. Found in the accounting/settlement red-team.

mfw78:
## What

Bound per-peer uncashed SWAP cheque exposure, and reject malleable cheque signatures at accept time.

A received cheque settles debt on structural validation while on-chain cashing is stubbed for v1, so nothing bounded how much uncashed value we treated as settled. `accept_cheque` now tracks per-peer `received_uncashed` and caps the credited increment to the remaining headroom under a configurable `bounce_limit` (default ten times the payment threshold), refusing with `SwapSettlementError::ExposureLimit` once maxed. The cumulative payout advances only by the settled amount, so a remainder withheld by the cap stays claimable on a later cheque. A new `--swap.bounce-limit` knob threads through the builder.

`SignedCheque::ensure_canonical` enforces low-s (EIP-2) and a canonical recovery byte, run before signer recovery, so a peer cannot present a second wire-distinct cheque for the same payout.

## Why

Closes #437. While cashout is stubbed, crediting uncashed cheques without a bound is theft-of-service: a counterparty buys real service with cheques that may never cash. The real fix (credit only on confirmed cash, with a chequebook solvency check) is tracked in #438; this is the v1 bounded-exposure guardrail. Pre-existing, found in the accounting and settlement red-team.

## Interop

No wire-visible change. Cheque body, JSON encoding, and EIP-712 signing/recovery are untouched. `ensure_canonical` only validates inbound bytes; `last_received_payout` (now advancing by the settled amount) is internal accounting state, never serialised. One behaviour shift: with a finite bounce limit, an inbound cheque can no longer reach the `AmountOverflow` path (the cap pre-empts it); this is strictly safer and the overflow path remains reachable and tested for conversion uses.

## Testing

`cargo test`/`cargo clippy -D warnings` clean for: vertex-swarm-bandwidth-swap (also `swap-chequebook`), vertex-swarm-bandwidth-chequebook (also `swap-chequebook`), vertex-swarm-node (`cli`), vertex-swarm-builder (`swap`). New tests cover exposure accumulation and cap, blocked-once-maxed, normal in-limit credit, and rejection of high-s and non-canonical-v signatures.

## AI Assistance

Claude Code (Opus 4.8) used for the red-team analysis that found the bug and for implementing the fix and tests, under human review.

### DIFF
--- a/crates/swarm/bandwidth/chequebook/src/cheque.rs
+++ b/crates/swarm/bandwidth/chequebook/src/cheque.rs
@@ -114,16 +114,9 @@ impl ChequeExt for Cheque {
 pub struct SignedCheque {
     /// The unsigned cheque data.
     pub cheque: Cheque,
-    /// The raw signature payload (canonically 65 bytes: r[32] + s[32] + v[1]).
-    ///
-    /// Kept as opaque [`Bytes`] rather than a typed
-    /// [`alloy_primitives::Signature`] so the payload round-trips whatever a peer
-    /// sends; it is parsed into a `Signature` only when recovering the signer
-    /// (see [`Self::recover_signer`]), where alloy validates the length and `v`.
-    /// A malleated twin recovers to the same issuer for the same cumulative
-    /// payout, so the per-peer monotonicity guard on the accept path rejects it
-    /// on replay; strict canonicalisation is enforced on chain at cash time
-    /// (#438).
+    /// Raw 65-byte signature (r, s, v), kept opaque so it round-trips the peer's
+    /// bytes; parsed into a `Signature` only on recovery (see
+    /// [`Self::recover_signer`]).
     pub signature: Bytes,
 }
 
--- a/crates/swarm/bandwidth/chequebook/src/cheque.rs
+++ b/crates/swarm/bandwidth/chequebook/src/cheque.rs
@@ -121,7 +121,9 @@ pub struct SignedCheque {
     /// length at 65 bytes, rejects non-canonical `v` values, and renormalizes
     /// `v` to `27 + parity` on re-encode. The opaque payload preserves whatever
     /// a peer sends and is parsed into a `Signature` only when verifying or
-    /// recovering a signer (see [`Self::recover_signer`]).
+    /// recovering a signer (see [`Self::recover_signer`]). The accept path calls
+    /// [`Self::ensure_canonical`] before crediting, so the preserved bytes never
+    /// settle debt in a malleable form.
     pub signature: Bytes,
 }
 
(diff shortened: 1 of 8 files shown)


crates/swarm/bandwidth/chequebook/src/cheque.rs
```diff
@@ -184,6 +186,29 @@ impl SignedCheque {
             .map_err(|e| ChequeError::SignatureRecovery(format!("invalid signature: {e}")))
     }
 
+    /// Reject a malleable signature.
+    ///
+    /// ECDSA signatures are malleable: for any valid `(r, s)` the pair
```
mfw78: Shouldn't this actually just be handled by alloy? Why are we reproducing this here?

mfw78: Agreed. Pushed e5f907ed: removed `ensure_canonical` entirely. The raw v-byte check reproduced what alloy's `Signature::try_from` parse already validates, and the low-s check, while it used alloy's `normalize_s`, was not load-bearing here: a malleated twin recovers to the same issuer for the same cumulative payout, so the per-peer monotonicity guard rejects it on replay regardless of signature bytes. We now lean on alloy's parse on the recovery path plus that guard; strict canonicalisation belongs on the on-chain cash path (#438). The bounded uncashed exposure (the actual C2 fix) is unchanged.

PR #455, opened 2026-06-21
feat(storer): persist reserve epoch for pullsync cursor handshake

mfw78:
## What

Persist a stable per-reserve `epoch` marker (a creation timestamp in a single-key metadata table) and expose it via an inherent `DbReserve::reserve_epoch()`.

## Why

Part of the pullsync protocol (#77). The cursor handshake returns per-bin cursors alongside an `epoch`; a pulling peer uses the epoch to detect that a reserve was wiped or recreated and to invalidate its cached cursors. Foundation crate, independent of the rest of the pullsync stack.

## Testing

`cargo test -p vertex-swarm-storer` (adds epoch-is-nonzero, survives-reopen, distinct-reserves); clippy clean.

## AI Assistance

Claude Code (Opus 4.8) used for implementation and verification.

### DIFF
--- a/crates/swarm/storer/src/db_reserve/store.rs
+++ b/crates/swarm/storer/src/db_reserve/store.rs
@@ -4,6 +4,7 @@
 
 use std::sync::Arc;
 use std::sync::atomic::{AtomicU8, Ordering};
+use std::time::{SystemTime, UNIX_EPOCH};
 
 use alloy_primitives::B256;
 use nectar_primitives::{Bin, ChunkAddress, ProximityOrder};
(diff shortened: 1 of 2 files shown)


crates/swarm/storer/src/db_reserve/store.rs
```diff
@@ -67,14 +70,24 @@ impl<DB: Database, BS: BatchStore> DbReserve<DB, BS> {
...
+            tx.ensure_table(ReserveMetadata::NAME)?;
+            if let Some(e) = tx.get::<ReserveMetadata>(EPOCH_KEY)? {
+                Ok(e)
+            } else {
+                let e = SystemTime::now()
+                    .duration_since(UNIX_EPOCH)
+                    .unwrap_or_default()
+                    .as_nanos() as u64;
+                tx.put::<ReserveMetadata>(EPOCH_KEY, e)?;
+                Ok(e)
```
mfw78, line 89: This should use our system time helper, and not this.

PR #475, opened 2026-06-21
refactor(swarm): consolidate handshake signing onto the node identity

### LINKED ISSUE #411: Consolidate handshake signer/network_id/nonce onto the node identity (follow-up to #392)
## Summary

Follow-up to the receipt-signing refactor in #392. There, `Receipt::sign` was changed to take an identity-shaped handle (the new `ReceiptSigner` trait in `vertex-swarm-net-pushsync`) that both signs the chunk address and supplies the two overlay-derivation inputs (`network_id` and `nonce`), instead of threading the signer, `network_id` and `nonce` as three loose parameters. This consolidates them onto the node's single identity, which is their real source of truth.

The same `signer` plus `network_id` plus `nonce` creep exists in the handshake path and is out of scope for #392. This issue tracks consolidating it there too.

## Sites

- `crates/swarm/peers/peer/src/swarm_peer.rs`: the `SwarmPeer` signing constructor takes `signer: &S`, `network_id` and `nonce` as separate parameters and recombines them to derive the overlay. These three are facets of one identity, exactly as on the receipt path.
- `crates/swarm/net/handshake/src/codec/synack.rs`: the synack encode path pulls `identity.signer()`, `spec.network_id()` and `identity.nonce()` apart and threads them down individually, rather than passing the identity handle.

## Proposed direction

Mirror the #392 approach: have the handshake signing path take an identity-like handle that signs and reports `network_id` and `nonce`, so the overlay-derivation inputs travel with the key that owns them and cannot be passed inconsistently. The exact placement of any shared trait should respect the existing crate layering (the handshake crates sit below `vertex-swarm-api`, so a minimal trait would live where both sides already share a crate, as `ReceiptSigner` does in pushsync).

This is a pure plumbing consolidation: the signed bytes and the derived overlay must stay byte-identical, as no wire format or consensus artifact changes.

## Constraints

- Pure refactor: no change to the signed message or to the derived overlay. Handshake acks and peer records are consensus-relevant, so the signed bytes must round-trip unchanged.
- No dependency inversion: the low-level handshake and peer crates must not gain a dependency on `vertex-swarm-api`.

---

This issue was drafted with AI assistance.

mfw78:
## What

`SwarmPeer::sign` took the signer, overlay, `network_id` and `nonce` as four loose parameters and recombined them to derive the overlay. Those are four facets of one node identity, so this threads the identity itself: `sign` now takes `&impl SwarmIdentity` and derives the overlay internally the same way `parse` recovers it. The two production handshake callers (`prepare_local_peer`, `sign_self_record`) collapse to one-liners.

`StorerCapability::new` gets the same treatment, taking the identity in place of the erased `signer` plus `network_id` plus `nonce` triple, so both `enable_storage` call sites stop pulling the identity apart.

This reuses the existing `vertex_swarm_api::SwarmIdentity` rather than minting a new trait: `vertex-swarm-peer` already depends on `vertex-swarm-api`, so no new dependency edge and no dependency inversion. (The receipt path needed its own `ReceiptSigner` trait only because pushsync sits below the api crate; handshake does not.)

## Why

Closes #411, the follow-up to the receipt-signing consolidation in #392. The overlay-derivation inputs now travel with the key that owns them and cannot be passed inconsistently.

This is a pure plumbing refactor: no wire-format or consensus artefact changes. Both production callers already passed `identity.overlay_address()`, which is by definition `compute_overlay(signer.address(), network_id, nonce)`, so the signed bytes and the derived overlay stay byte-identical. The one test that deliberately signed a mismatched overlay (`invalid_overlay_rejected`) is rerouted to fabricate that inconsistency on the wire instead of through `sign`, so the `InvalidOverlay` rejection stays covered.

## Testing

- `cargo clippy -p vertex-swarm-peer -p vertex-swarm-net-handshake -p vertex-swarm-client-behaviour -p vertex-swarm-node --all-targets --all-features -- -D warnings`: clean.
- `cargo test -p vertex-swarm-peer`: 22 unit tests plus the 3 pinned handshake conformance vectors in `tests/interop.rs` pass unchanged, which is the byte-identity guarantee.
- `cargo test -p vertex-swarm-net-handshake` and `-p vertex-swarm-client-behaviour`: pass.
- `cargo test -p vertex-swarm-node`: passes, including `responsible_storer_stores_and_signs_a_receipt` (asserts the minted receipt's storer overlay recovers to the identity's own overlay).

## AI Assistance

Claude Code (Opus 4.8) used for the codebase audit, the design and adversarial-QA review of the approach, the implementation and the verification, and this description. A human author is accountable and has reviewed it.

### DIFF
--- a/crates/swarm/client-behaviour/src/storer.rs
+++ b/crates/swarm/client-behaviour/src/storer.rs
@@ -9,7 +9,7 @@
 use std::sync::Arc;
 
 use alloy_signer::SignerSync;
-use vertex_swarm_api::ReserveStore;
+use vertex_swarm_api::{ReserveStore, SwarmIdentity, SwarmSpec};
 use vertex_swarm_net_pushsync::ReceiptSigner;
 use vertex_swarm_primitives::{NetworkId, Nonce};
 
@@ -30,17 +30,18 @@ pub struct StorerCapability {
 }
 
 impl StorerCapability {
-    pub fn new(
-        reserve: Arc<dyn ReserveStore>,
-        signer: Arc<dyn SignerSync + Send + Sync>,
-        network_id: NetworkId,
-        nonce: Nonce,
-    ) -> Self {
+    /// Build the capability from the node identity. The signer is erased to
+    /// `dyn SignerSync`; `network_id` and `nonce` are read off the identity so
+    /// the minted receipt's overlay derives from the same key.
+    pub fn new(reserve: Arc<dyn ReserveStore>, identity: &impl SwarmIdentity) -> Self {
+        // `I::Signer: Signer + SignerSync + Send + Sync`, so erasing to
+        // SignerSync is sound.
+        let signer: Arc<dyn SignerSync + Send + Sync> = identity.signer();
         Self {
             reserve,
             signer,
-            network_id,
-            nonce,
+            network_id: identity.spec().network_id(),
+            nonce: identity.nonce(),
         }
     }
 }
--- a/crates/swarm/client-behaviour/src/storer.rs
+++ b/crates/swarm/client-behaviour/src/storer.rs
@@ -8,31 +8,34 @@
 
 use std::sync::Arc;
 
-use alloy_signer::SignerSync;
+use alloy_primitives::{Address, B256, ChainId, Signature};
 use vertex_swarm_api::{ReserveStore, SwarmIdentity, SwarmSpec};
-use vertex_swarm_net_pushsync::ReceiptSigner;
-use vertex_swarm_primitives::{NetworkId, Nonce};
+use vertex_swarm_primitives::{NetworkId, Nonce, OverlaySigner, SignerSync};
 
 /// Reserve plus the node's receipt-signing identity, shared into each handler.
 ///
-/// `network_id` and `nonce` are the overlay-derivation inputs a forwarder
-/// recovers from the minted receipt; implementing [`ReceiptSigner`] lets the
-/// handler pass this straight to [`vertex_swarm_net_pushsync::Receipt::sign`].
-/// `put` admits unconditionally; capacity is held out of band by the eviction
-/// primitives ([`ReserveStore::evict_from_bin`] /
-/// [`evict_batch`](ReserveStore::evict_batch)), not by ingest back-pressure.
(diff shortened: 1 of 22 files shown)


crates/swarm/client-behaviour/src/storer.rs
```diff
@@ -30,17 +30,18 @@ pub struct StorerCapability {
...
+        // `I::Signer: Signer + SignerSync + Send + Sync`, so erasing to
+        // SignerSync is sound.
+        let signer: Arc<dyn SignerSync + Send + Sync> = identity.signer();
         Self {
             reserve,
             signer,
-            network_id,
-            nonce,
+            network_id: identity.spec().network_id(),
+            nonce: identity.nonce(),
```
mfw78: But given that the storer has specific consensus based information that comes from the spec, and also deals with the identity, couldn't the storer be made generic over or use those traits for this information and streamline?

mfw78 (commented):
Can we be more targeted with use of the swarm spec / identities and the way the traits are composed?

crates/swarm/client-behaviour/src/storer.rs
```diff
@@ -8,48 +8,61 @@
...
 #[derive(Clone)]
 pub struct StorerCapability {
     pub(crate) reserve: Arc<dyn ReserveStore>,
     signer: Arc<dyn SignerSync + Send + Sync>,
+    // `dyn SignerSync` omits `address()` (it is on the async `Signer`), so the
+    // overlay-derivation address is captured at construction.
+    address: Address,
     network_id: NetworkId,
     nonce: Nonce,
 }
```
mfw78, line 27: Why can't this just take the OverlaySigner?

mfw78: Done in 8fbd5444. `StorerCapability` now holds a single `Arc<dyn OverlaySigner + Send + Sync>` and `new` takes exactly that, so the four loose fields (erased signer, address, network_id, nonce) are gone. It has to be the owned `Arc` rather than `&impl OverlaySigner`, since the capability needs to store a signable handle and you cannot extract an owned signer from a borrow; the node erases its identity into the handle once at `enable_storage`.

PR #480, opened 2026-06-21
refactor(accounting): remove unused builder apply combinators

### LINKED ISSUE #484: refactor(accounting): dedup builder apply/apply_if combinators
Define apply/apply_if once as a blanket BuilderExt and drop the duplicate. Parent: #476.

mfw78:
## What

Remove the `apply`/`apply_if` builder combinators entirely (from both `AccountingBuilder` and the swarm builder), rather than relocating them.

## Why

Reviewing the call sites, these combinators have no production callers; the only use anywhere was a single self-referential test (`apply_if(true, |b| b)`). Rather than move a generic, unused utility into (or out of) the swarm-domain crate path, both copies are deleted. If a generic builder combinator is wanted later it belongs in a generic node or util crate, not a `vertex-swarm-*` crate. Closes #484.

## Testing

`cargo clippy --all-targets --all-features` and `cargo test` on `vertex-swarm-accounting` and `vertex-swarm-builder`; workspace grep confirms no remaining `BuilderExt` reference.

## AI Assistance

Claude Code used for the audit, the call-site analysis, and the removal.

### DIFF
--- a/crates/swarm/accounting/core/src/builder.rs
+++ b/crates/swarm/accounting/core/src/builder.rs
@@ -80,22 +80,6 @@ impl<C, P> AccountingBuilder<C, P> {
         self
     }
 
-    /// Apply a transformation function.
-    pub fn apply<F>(self, f: F) -> Self
-    where
-        F: FnOnce(Self) -> Self,
-    {
-        f(self)
-    }
-
-    /// Apply a transformation function if condition is true.
-    pub fn apply_if<F>(self, cond: bool, f: F) -> Self
-    where
-        F: FnOnce(Self) -> Self,
-    {
-        if cond { f(self) } else { self }
-    }
-
     /// Get a reference to the config.
     pub fn config(&self) -> &C {
         &self.config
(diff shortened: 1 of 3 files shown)


crates/swarm/accounting/core/src/builder.rs
```diff
@@ -10,6 +10,21 @@ use vertex_swarm_api::{
...
+    }
+
+    /// Apply a transformation function only if `cond` holds.
+    fn apply_if<F: FnOnce(Self) -> Self>(self, cond: bool, f: F) -> Self {
+        if cond { f(self) } else { self }
+    }
+}
+
+impl<T: Sized> BuilderExt for T {}
+
```
mfw78: Aren't these something more generic and better suited to sit in some vertex-node crate or similar, not in the vertex-swarm-* path?

mfw78: Good catch. Checking the call sites, apply/apply_if had no production callers; the only use anywhere was a no-op self-test (`apply_if(true, |b| b)`). Rather than relocate a generic util that nothing uses, I removed the combinator from both crates entirely. If we want a builder combinator later it should live in a generic node or util crate, not the swarm-domain path.

Issue #425, opened 2026-06-20
feat(swarm): wasm Client-with-SWAP chain provider and IndexedDB cache backend


mfw78:
RFC step 11 (depends on the public seams, with_swap, and the wasm core convergence). Confirm-and-extend the shared alloy Provider (per docs/design/chain-service.md) with a browser-reachable RPC transport so swap/chain compile and run on wasm (feature + provider impl gated, not target_arch gated). Add the IndexedDB `SwarmLocalStore` impl as the wasm persistent cache backend. Together these enable a wasm Client+SWAP acting as a cache/forwarder.

See the design RFC: `docs/design/node-type-api.md` (PR #414).

mfw78:
## Sharpened scope (the wasm primitives floor under #487)

With the #424 client-core convergence merged, this issue is the wasm primitives floor that the wasm SWAP client (#487) builds on. It is two independent halves; the chain-transport half blocks #487's on-chain cashout, the IndexedDB half is orthogonal and can land in parallel.

### A. Browser-reachable chain-provider transport

`docs/design/chain-service.md` already sanctions a wasm alloy `Provider` over a browser transport. The construction seam `build_chain_provider` (`crates/swarm/builder/src/chain.rs`) has a transport-portable body but lives in the native-only `vertex-swarm-builder`, which never enters the wasm cone. The work:

- Relocate the constructor to a wasm-buildable crate (`vertex-chain` is the natural home; it is already wasm-CI-green and uses `alloy-provider` with `default-features = false`).
- Target-split `alloy-provider` there exactly as `crates/swarm/accounting/chequebook/Cargo.toml` already does: native = `["reqwest", "reqwest-native-tls"]`, wasm32 = `["reqwest"]` (browser fetch; TLS via the browser). The only errant unconditional `reqwest-native-tls` is `crates/swarm/builder/Cargo.toml:102-105`, moot once the seam relocates.
- Carry the `getrandom = ["js"]` wasm32 table (mirror `crates/chain/Cargo.toml`) for the k256 signer path, or the wasm build fails with a no-backend error.

No launcher changes here. Output: a `SharedChainProvider` constructible on `wasm32`.

### B. IndexedDB SwarmLocalStore backend

[...]

PR #502, opened 2026-06-22
feat(chain): make the chain-provider transport wasm-reachable

mfw78:
What

Relocate the shared chain-provider constructor (`build_chain_provider` and `SharedChainProvider`) out of the native-only `vertex-swarm-builder` into `vertex-chain`, behind a new `provider` feature, and target-split its alloy transport so the constructor builds for `wasm32-unknown-unknown`. The builder keeps the config and spec resolution (`build_node_chain_provider`) and calls into the relocated constructor.

The `provider` feature pulls the alloy HTTP transport (target-split: native HTTP over reqwest with system TLS on native, browser fetch on wasm32, mirroring `crates/swarm/accounting/chequebook`), the local wallet signer, and a single tracing line. Without the feature `vertex-chain` stays the pure, transport-free layer (the `Provider` extension trait, the error taxonomy, the address book), so the default wasm cone is unaffected. The wasm32 `getrandom = ["js"]` table the k256 signer path needs was already present on the crate.

The builder `chain` feature now enables `vertex-chain/provider`; its now-errant direct `alloy-provider` dependency and the unconditional `reqwest-native-tls` are removed, and `build_node_chain_provider` maps the relocated `ProviderBuildError` onto `SwarmNodeError::Chain`.

This revision also strips the spec-derived address book out of `SharedChainProvider`. The handle previously bundled a `ChainConfig` copy (chequebook factory, BZZ token, price oracle, settlement chain), which is pure spec-derived state. It now carries only the live `DynProvider`: `SharedChainProvider` is a single-field newtype with a `provider()` accessor and no `addresses()`. The validation chain comes from `spec.chain` passed into `build_chain_provider(rpc_url, signer, expected: Chain)`, and the contract addresses are read at the edge by each consumer. The only consumer, the swap cashout wiring, now resolves `ChainConfig::from_swarm(spec.swarm())` at the call site and feeds it to `Cashout::new`; the launch path still resolves the address book solely for the no-deployment fast-fail.

Part of #425 (half A: chain transport). This is the wasm chain-transport floor the wasm SWAP client builds on for on-chain cashout. The IndexedDB half (B) is orthogonal and lands separately.

Why

Closes the chain-transport half of #425. The construction seam had a transport-portable body but lived in a native-only crate that never enters the wasm cone, so a browser SWAP client could not construct a `SharedChainProvider`. Reviewing the relocation surfaced that the handle carried a `ChainConfig` copy that duplicates spec-derived state; the handle should carry only the live connection, with the chain id and contract addresses sourced from the `Spec`.

Testing

- `cargo fmt --all`
- `cargo clippy -p vertex-chain -p vertex-swarm-builder -p vertex-swarm-accounting-swap --all-targets --all-features -- -D warnings` (clean)
- `cargo test -p vertex-chain -p vertex-swarm-builder -p vertex-swarm-accounting-swap --all-features` (all pass; the offline construction-seam test now validates against `Chain::sepolia()`)
- wasm proof: `cargo build --target wasm32-unknown-unknown -p vertex-chain --features provider` builds the constructor over the browser fetch transport; `--tests` links the `#[cfg(target_arch = "wasm32")]` wasm-bindgen test asserting `ProviderBuilder::new().connect_http(url)` constructs without panicking (no live RPC). The default (no-`provider`) wasm build and the chequebook `swap-chequebook` wasm build both still compile.
- CI: the `wasm` job builds `vertex-chain --features provider` and its wasm tests.


crates/chain/src/build.rs
```diff
@@ -0,0 +1,166 @@
...
+    }
+}
+
+impl std::fmt::Debug for SharedChainProvider {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("SharedChainProvider")
+            .field("chain", &self.addresses.chain)
+            .finish_non_exhaustive()
+    }
+}
```
mfw78, line 67: I don't get why this is concretely required. Aren't the addresses just be default coming from the spec, and that the spec should be threaded through, in which case the last left is the provider, and why is it needed like this? It looks overengineered?

mfw78: Agreed - this was caching spec-owned state in the handle. Fixed in d7e3fa38: SharedChainProvider is now a single-field newtype over DynProvider (provider only, no ChainConfig, no addresses() accessor). build_chain_provider takes expected: Chain for the fail-fast chain-id check, fed spec.chain at the call site, and the one cashout consumer reads ChainConfig::from_swarm(spec.swarm()) from the threaded spec rather than a copy on the handle. #503 has been rebased onto this so its relocated swap wiring consumes the provider-only shape; there is no SharedChainProvider::addresses() usage anywhere in the tree.

crates/chain/Cargo.toml
```diff
@@ -19,25 +30,41 @@ nectar-swarms = { workspace = true }
...
 strum = { workspace = true }
 thiserror = { workspace = true }
+# `provider` feature: a single chain-connected log line in the construction seam.
+tracing = { workspace = true, optional = true }
+
+# Native uses the reqwest transport with native-tls (system OpenSSL). The rustls
+# path drags in a root-certs crate under a license outside the workspace
+# allow-list, so native TLS goes through system OpenSSL.
+[target.'cfg(not(target_arch = "wasm32"))'.dependencies]
+alloy-provider = { workspace = true, features = ["reqwest-native-tls"] }
```
mfw78, line 57: Shouldn't we be using rustls throughout for native?

mfw78: Verified by switching the native target to reqwest-rustls-tls and regenerating the lock: cargo deny check licenses fails on webpki-root-certs v1.0.8 (CDLA-Permissive-2.0), pulled via rustls-platform-verifier from reqwest 0.13's rustls feature (which selects aws-lc-rs, not the ring already in the tree). CDLA-Permissive-2.0 is not in deny.toml's allow-list, so native-tls is genuinely forced today. The comment is imprecise though, so I will reword it to name the real blocker rather than a vague "root-certs crate". Going rustls throughout (this crate plus chequebook, the only other native-tls/openssl consumer) is a clean follow-up once we add CDLA-Permissive-2.0 to the allow-list; that also drops openssl from the lock entirely. It is an OSI-permissive, copyleft-free data licence covering the bundled CA cert set. Happy to do that sweep if you want it.

PR #528, opened 2026-06-26
feat(vertex): gate the storer cone behind a cargo feature so the default build is a bare client

### LINKED ISSUE #527: Build a bare-client vertex by default, gate the storer cone behind a cargo feature
## What

The default `cargo build -p vertex` should produce a bare client: it must compile neither the storer code cone (the persisting reserve, the neighbourhood puller, the batch store, redistribution, and the `StorerNode` composite) nor, by extension, the chain and swap stacks.

The full storer node is selected with `--features storer`, which expands to the storer cone plus `chain` and `swap`. The `swap` and `chain` features stay opt-in on their own, so a swap-capable client without the storer cone is also buildable. A default binary asked to run a storer at runtime should return an error advising a rebuild, not panic.

## Why

The bare client is the wasm and lightweight-embedding target. Compiling the storer reserve, puller, and on-chain settlement into every default build bloats the binary and the dependency tree for users who only read and write chunks.

It is also the prerequisite for the cross-platform client binary and Docker release pipeline, which ships the default bare client and is tracked separately.

## Acceptance

`cargo tree -p vertex` proves the storer crates (`vertex-swarm-storer`, `vertex-swarm-puller`, `vertex-swarm-redistribution`, `vertex-swarm-storer-behaviour`) are absent from the default tree and present under `--features storer`. Both cones build and lint clean. A cone guard enforces the exclusion in CI.

mfw78:
## What

The default `vertex` binary is now a bare client: it compiles neither the storer code cone (the persisting reserve, the neighbourhood puller, the batch store, redistribution, and the `StorerNode` composite) nor, by extension, the chain and swap stacks.

The full storer node is selected with `--features storer`, which expands to the new `storer-core` feature plus `chain` and `swap`. A default binary asked to run a storer at runtime returns an error telling the operator to rebuild, rather than panicking. The `swap` and `chain` features still add their stacks on their own, so a swap-capable client without the storer cone is buildable.

`vertex-swarm-builder` gains a `storer-core` feature that makes the four storer crates optional dependencies and cfg-gates the storer launch, config, and builder paths. `vertex-swarm-node` gains a `storer` feature that gates the `StorerNode` module and its native storer-behaviour and puller dependencies, and its `cli` feature no longer unconditionally pulls redistribution. A justfile cone guard and a new CI `features` job assert the default tree never resolves the storer crates and that both shipped cones build clean.

## Why

The bare client is the wasm and lightweight-embedding target. Compiling the storer reserve, puller, and on-chain settlement into every default build bloated the binary and the dependency tree for users who only read and write chunks. Gating the storer cone keeps the default client lean and gives operators an explicit `--features storer` opt-in for the full node.

It is also the prerequisite for the cross-platform client binary and Docker release pipeline, which ships this default bare client.

Closes #527.

## Note on the second commit

A separate `chore(deps)` commit bumps `quinn-proto` to 0.11.15 to clear RUSTSEC-2026-0185 (a remote memory-exhaustion advisory published on 2026-06-22 against 0.11.14, a transitive dependency). The advisory is unrelated to the feature-gating but blocks the required `audit` check on every branch, so the minimal lockfile bump rides along here to keep this PR green. It can be split into its own PR if preferred.

## Testing

`cargo tree -p vertex` proves the five storer crates are absent from the default tree and present under `--features storer`. Built and checked `-p vertex` default, `--features storer`, `--features swap`, and `--all-features` (the last under `nix develop`). Clippy `-p vertex -p vertex-swarm-builder -p vertex-swarm-node --all-targets -- -D warnings` is clean both default and with `--features storer`, as is the all-features clippy run. Tests for the builder and node crates pass with and without `--features storer` (the storer cone adds nine gated builder tests). `just check-cone` and `cargo fmt --all --check` are clean. The quinn-proto bump is a minimal lockfile change verified against the crates.io index checksum and `cargo metadata --locked`.

## AI Assistance

Claude Code used for the design and implementation of the feature-gating and the advisory bump.


mfw78:
Extended this PR with the feature/cfg architecture work (the client/storer/FFI organisation), per a feature-architecture red-team:

- **Feature and cfg contract** added to AGENTS.md: `default = []` is the load-bearing bare client; features are for capabilities never node types (node type is the runtime `SwarmNodeType`); `#[cfg(feature)]` only at composition roots, never in domain crates; platform boundaries use `target_arch`; FFI is a crate (cdylib), not a feature; and workspace members must not unconditionally enable `chain`/`swap`/`storer` (the unification footgun).
- **CI cone matrix** pinned in the `features` job: a `--no-default-features` footgun guard, plus builds of the swap-client and FFI cones, and a new `vertex-ffi` storer-cone guard alongside the existing default-vertex one.
- **FFI** stays a crate boundary (no `ffi` feature; that verdict was unanimous on idiomacy grounds). It keeps the builder + redb path documented as the native analogue of the browser IndexedDB cache for a future mobile embedder; the storer cone is already excluded from `vertex-ffi`.

The `swap` cfg-proliferation (~48 gates) is tracked separately in #532. No `.rs` changed in these two commits; they are docs + CI + justfile only.

mfw78:
Folded in the figment config-load fix (addresses #535): `swap.bounce_limit` (and any figment-loaded u128 config field) now serialises as a string so figment can round-trip it. The binary previously could not boot with the default config (`invalid type: found integer ... as u128`) - a pre-existing bug on main, missed by CI because no job boots the binary and the old `test_default_config` used a stub with no u128. Now `vertex node` boots through the shell to the event loop. Added a regression test (`bin/vertex/tests/config_load.rs`) that loads the real `ProtocolConfig` default and a quoted-u128 config file. Closes #535.

Issue #526, opened 2026-06-25
epic: light-client retrieval conformance and hardening


mfw78:
## Problem

A clean-sheet re-implementation of the retrieval fixes prototyped on `feat/node-ip-and-allowance-fixes`, which drove the in-browser light client through real mainnet downloads and surfaced a set of distinct root problems. That spike proves the diagnoses and carries reference fixes but is not the shape to merge; this epic re-implements them against `main`, one focused PR per problem.

The spine is v1 reference-interop: the serving node debits a requester per request and, when debt crosses its disconnect line (1,687,500 AU for a light node), resets and blocklists it. The open-loop client crossed that line under any sustained download and got dropped, so it could not acquire users on the live network.

## Success criteria

- [ ] Protocol compliance: the client settles its pseudosettle debt and is no longer blocklisted by serving peers under sustained download.
- [ ] No self-inflicted peer churn: shared-IP cohorts, blameless transport resets, and fan-out overruns no longer evict serving peers.
- [ ] Download path is usable: streaming to disk, out-of-order delivery, pipelined prefetch, and a non-saturated wasm thread.

## Sub-tasks

Protocol compliance (the reference was disconnecting us):

- [x] #512 Light client never settles its pseudosettle debt, so serving peers blocklist it
- [x] #513 Client settles only on delivery, so carried debt opens the next request over the line
- [x] #514 Client debt view lags bee's shadow-reserved view, crossing bee's line under wide prefetch
- [x] #515 Accounting bans serving peers for our own debt (a debtor banning its creditor)
- [x] #516 Nothing bounds per-peer unsettled debt under sustained load

Self-inflicted peer churn:

- [x] #517 Shared-IP storer cohorts are misflagged as Sybil identity-cycling and evicted
- [x] #518 Blameless transport-reset closes are penalized as early disconnects, evicting serving peers
- [x] #519 Instant re-dial of a transport-reset peer reloads it into the reset (bin-0 churn cascade)
- [x] #520 Retrieval fan-out overruns the peer set and per-connection stream budget

Download performance and memory:

- [x] #521 Whole file is buffered in browser memory, so large downloads exhaust the heap
- [x] #522 Ordered download delivery lets one slow chunk gate everything behind it
- [x] #523 Level-synchronous prefetch serializes deep chunk trees
- [x] #524 WebGL globe saturates the single wasm thread and starves retrieval
- [x] #525 Inline sink writes stall the fetch stream on the single thread

## Non-goals

[...]

PR #559, opened 2026-06-28
fix(swarm-node): bound the debtor settle with a timeout so a withholding peer cannot pin the dedup guard

### LINKED ISSUE #555: accounting: bound the debtor settle with a timeout so a withholding peer cannot pin the dedup guard
## Problem

A debtor-initiated settle to a peer is deduped by a per-peer in-flight guard (one settle per peer at a time), cleared when the settle future completes. But the pseudosettle and swap settle awaits have no timeout: a peer that accepts the settle stream and then never acks parks the settle future forever, so the guard never clears and the peer is pinned "settle in flight" for the process lifetime. No further settle to that peer can start, its debt climbs to the disconnect line, and the peer is dropped, while settlement to it is dead in the interim.

## Current behaviour

The node-side settle trigger spawns the settle and clears the per-peer guard on completion (success, error, or cancellation) via an RAII drop. The provider settle resolves on the network ack with no upper bound, so a withholding peer keeps the future (and the guard) alive indefinitely.

## Proposed

Bound the spawned settle future with a `vertex_tasks::time` timeout, so the guard always drops within a bounded window and the peer can be settled again. The timeout is best-effort (the next request retries); it only needs to release the dedup guard.

## Scope

- Affected: `crates/swarm/node` (the settle trigger), possibly `crates/swarm/accounting/{pseudosettle,swap}`

## Notes

Pre-existing latent bug surfaced by the accounting-seam correctness red-team; not introduced by the seam reshape. Part of the client retrieval hardening spine (epic #526). Test the release via the timeout, not via a panicking task (CI nextest rejects spawned-task panics).

mfw78:
## What

Bounds the debtor-initiated settle with a timeout so a withholding peer cannot pin the per-peer single-in-flight dedup guard.

The settle trigger keeps at most one settle in flight per peer, clearing the dedup entry when the spawned settle future completes or drops. But the pseudosettle/swap settle await is unbounded: a peer that accepts the settle stream and never acks parks the future forever, so the guard never drops and that peer is pinned "settle in flight" for the process lifetime, starving its settlement while its debt climbs. The spawned settle is now wrapped in a `vertex_tasks::time` timeout (15s, native-backed by tokio, wasm-safe), so the guard always drops within a bounded window and a later trigger can start a fresh settle. The settle is best-effort, so a timed-out settle simply retries on the next trigger.

## Why

Closes #555.

A pre-existing latent bug surfaced by the accounting-seam correctness red-team: the single-in-flight dedup is also the per-peer rate limit, so one hung settle disabled all future settlement to that peer. Part of the client retrieval hardening spine (epic #526).

## Testing

`cargo nextest run -p vertex-swarm-node --all-features`: 129 pass. A new test parks a never-acking peer's settle and advances tokio paused time past the timeout, asserting the peer leaves the in-flight set so a fresh trigger starts a new settle; the existing single-in-flight dedup test still passes. `cargo clippy --all-targets --all-features -- -D warnings` clean plus a swap-off pass.

## AI Assistance

Claude Code used for the analysis and implementation.

### DIFF
--- a/crates/swarm/node/src/selection.rs
+++ b/crates/swarm/node/src/selection.rs
@@ -24,6 +24,7 @@ use std::collections::HashMap;
 use std::future::Future;
 use std::pin::Pin;
 use std::sync::Arc;
+use std::time::Duration;
 
 use futures::FutureExt;
 use futures::future::Shared;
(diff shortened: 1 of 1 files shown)


mfw78:
Superseded by #562. Red-team found the debtor settle does not actually lack a bound: the pseudosettle/swap settle substreams already carry the handler's 30s Config::timeout, but a substream failure was mapped to a no-op ClientEvent::ProtocolError, so the service leaked its pending settle and handle.settle() hung. #562 propagates that existing libp2p bound to the settle instead of layering a second SETTLE_TIMEOUT (which also removes the wasm spawn_local cfg-gate this PR needed).

Issue #514, opened 2026-06-25
retrieval: reserve per-chunk price at dispatch for debt parity


mfw78:
## Problem

Origin retrieval debits a peer only on delivery, so the in-flight reservation stays at zero between dispatch and delivery.

The serving creditor (the reference) shadow-reserves each in-flight leg at request time and meters debt as `balance + shadowReserved`. Under a wide prefetch fan-out the creditor crosses its disconnect line while our delivered-only view still looks safe, and it disconnects us.

Our view of per-peer debt must match the creditor's, or any debt gating we add is computed against the wrong number.

## Current behaviour

`crates/swarm/net/retrieval` debits the price on delivery only; no reservation is held across the in-flight leg.

## Proposed

- Reserve the per-chunk price at dispatch, hold it across the in-flight permit, apply on delivery, and release on failure (mirroring the forwarder two-leg reserve/apply/drop).
- Pace the throttle against `balance + threshold - reserved`.

## Scope

- Affected: `crates/swarm/net/retrieval`, `crates/swarm/accounting`

## Notes

Light disconnect line is 1,687,500 AU.

Prototype fix on `feat/node-ip-and-allowance-fixes`: `c9b2cbf2`, test `8df86acd`. Context: `docs/development/light-client-retrieval-hardening.md` (P3) on that branch.

Verified: on `origin/main` (37baadc1) the reserve-at-dispatch path is absent; this is part of the unmerged light-client retrieval hardening spine (epic #526).

Issue #569, opened 2026-06-29
accounting: fork-gated balance reconciliation to maximise the free pseudosettle rate


mfw78:
## Problem

The reference and vertex book mutual debt asymmetrically: the server credits itself when it serves a chunk, the client debits itself when it receives one. The two ledgers therefore diverge, worst under reset churn, and there is no wire mechanism for two peers to agree on their mutual balance. To stay safe a client must book pessimistically and keep headroom below the disconnect line, which caps the usable **free** pseudosettle rate below the true line. SWAP chequebooks similarly carry no in-protocol reconciliation of the agreed cumulative payout.

## Current behaviour

Each side keeps its own ledger; reconciliation is implicit and lossy (a lost in-flight delivery is charged by one side and not the other). v1 mitigates this locally with pessimistic booking plus headroom (#568), trading throughput for safety, but the desync margin is pure overhead on the free rate.

## Proposed

A fork-gated (`SwarmHardfork` / `ForkDigest`) wire mechanism for two peers to periodically exchange and reconcile their view of the mutual balance, and with SWAP the chequebook cumulative-payout state, so both agree on the true debt. With an agreed balance a client can run pseudosettle and settlement up to the actual line with no desync margin, maximising the free rate, and reconcile chequebook state cleanly.

Must be fork-gated because it is a wire-visible divergence from the reference, and interop-safe: a peer that has not adopted the fork stays on the current asymmetric model and continues to function.

## Scope

- A new or extended wire message (pseudosettle, or a dedicated balance-reconciliation protocol), a `SwarmHardfork` variant with `ForkDigest` selection at handshake, and the accounting plus swap/chequebook layers.

## Related

- Removes the need for the v1 local mitigation #568. Guard from #515 (do not penalise a peer for our own debt). Resets that amplify the desync are #519.

## Notes

Post-v1, wire-fork: must not ship in v1, which stays reference-conformant. This is the durable fix that the v1 pessimistic-booking mitigation approximates locally.

## AI Assistance

AI Assistance: Claude Code (Opus 4.8) used for the diagnosis that motivated this and for drafting the issue.

mfw78:
Refinement of the "new or extended wire message" in Proposed: reconcile at connect by piggybacking each peer's believed mutual state onto the `AnnouncePaymentThreshold` message (pricing protocol), so the two peers converge their ledgers at the reconnect boundary (a resync handshake), not only periodically.

Design constraints from red-teaming this:

[...]

Issue #515, opened 2026-06-25
accounting: stop scoring peers for our own unpaid debt


mfw78:
## Problem

When our own debt to a peer crosses our disconnect threshold, accounting scores that peer as an `AccountingViolation` and drives its score to the ban line.

Under a sustained download the closest storers reach this line purely because the download outruns the rate at which they forgive our debt, not because they misbehave. So the client scores and bans the very storers it depends on, shrinking its own neighbourhood and amplifying churn. A debtor cannot blame its creditor for its own unpaid debt.

## Current behaviour

`crates/swarm/accounting/core/src/accounting/mod.rs` (around line 136): when the projected balance crosses `-disconnect_threshold`, the receive path reports `SwarmScoringEvent::AccountingViolation` against the peer.

## Proposed

- Refuse the receive at the disconnect line so the caller routes the chunk elsewhere while a background settle drains the debt, but never score the peer for our own debt.
- Keep the over-acceptance and malformed-chunk scoring paths intact.

## Scope

- Affected: `crates/swarm/accounting`

## Notes

Evidence (first 60s of a 60 MB download, after the fix): auto-bans of serving storers 9 to 0; local score-driven disconnect-line crossings 30 to 0; close-neighbourhood cuts 71 to 0; connected footprint 95 to 123.

Prototype fix on `feat/node-ip-and-allowance-fixes`: `faac9950`. Context: `docs/development/light-client-retrieval-hardening.md` (P4) on that branch.

Verified: on `origin/main` (37baadc1) the receive path still emits `AccountingViolation` on own-debt threshold crossing.

PR #570, opened 2026-06-29
fix(swarm-node): book the origin retrieval debit at dispatch

### LINKED ISSUE #519: retrieval: stop peer overrun from collapsing bin-0 depth mid-download
## Status

Root cause addressed: both overrun brakes have merged to main. The debt brake (#514 shadow-reserve at dispatch, #516 admission gate, #513 pre-pay carried debt) landed with the light-client retrieval hardening spine; the stream brake landed as #544 (the per-peer in-flight skip-busy cap, which superseded the closed #355). Residual scope is now the demo-side fast re-dispatch (coordinated with #509) plus verifying the acceptance outcomes under a sustained browser download.

## Problem

Under a sustained browser-client download, bin 0 (our closest neighbourhood, which the retrieval forwarding hot path loads hardest) drains below saturation and routing depth collapses mid-download. The same close peer transport-resets several times in quick succession until its bin empties, turning a single honest reset into a self-reinforcing cascade.

## Current behaviour

A serving peer io_resets the connection. The close is correctly blameless after #542 (a productive `RemoteClose` is neither scored nor backed off), so the replenishment dialler re-selects it on the next sub-second dial tick and reloads it straight back into the reset. Observed: reset-to-redial gap median 0.09s, the same peer resetting up to 6x, bin 0 down to 0-1 connected.

## Root cause

The reset is not a peer-side timing quirk; it is backpressure against our own overrun. The serving reference node resets us for two reasons, both ours to fix:

- Stream-budget overrun: a wide prefetch fan-out opens more concurrent retrieval substreams to one chunk-holder than it tolerates. Bounded by #544 (merged; the per-peer in-flight skip-busy cap that superseded the closed PR #355).
- Debt-line crossing: our per-peer unsettled debt crosses the remote economic disconnect line faster than pseudosettle drains it, so the creditor resets and blocklists us. This is the load-bearing conformance brake, now closed by #516 (admission gate), with #513 (pre-pay carried debt before the next request) and #514 (shadow-reserve the per-chunk price at dispatch so our debt view matches the creditor's). All three merged.

Re-dialling a peer we just overran simply repeats the offence, which is why it churns. Pacing the re-dial treats the symptom; bounding the overrun removes the cause and keeps the peer serving us rather than evicting it from the neighbourhood.

## Proposed

The durable fix is the overrun brakes above (#544 for streams, #516/#513/#514 for debt). Those close this issue's root cause and have now all merged.

This issue's residual scope narrows to recovering the work a reset loses, without re-routing retrieval away from the close peers (which regressed earlier: it concentrated the deep-leaf tail onto a smaller live set and provoked more resets):

- Demo-side fast re-dispatch: distinguish a connectivity failure from a not-found in the prefetch retry. A connectivity failure re-dispatches the in-flight chunk to a fresh live peer at once; a not-found keeps the existing backoff so ancestors can warm first. Coordinate with #509, which reworks the demo download path.

Explicitly dropped: the non-penalizing re-dial cooldown prototyped on `feat/node-ip-and-allowance-fixes`. Holding a useful bin-0 peer out of selection is at best a band-aid, is self-contradictory (it removes from bin 0 the very peer whose loss we are trying to prevent), and is redundant now that the overrun brakes have landed. We are not pursuing it.

## Acceptance

Outcome-based and mechanism-agnostic, to verify against current main now that the brakes have merged:

- [ ] Depth holds through a sustained download; no bin-0 collapse.
- [ ] A full download completes byte-exact at or near line rate.
- [ ] Per-peer reset count falls toward at most one per peer as a consequence of the overrun brakes, not of re-dial pacing.

## Scope

- Affected: `bin/swarm-demo` retrieval re-dispatch; verification of depth and throughput. Root-cause work lived in #544 (streams) and #516/#513/#514 (debt), all merged.
- Not in scope: the re-dial cooldown (dropped); routing retrieval away from close peers (regressed).

## Notes

Part of the light-client retrieval hardening epic #526.

Supersedes the earlier framing that treated this as a blameless re-dial timing problem. #542 (disconnect attribution) confirmed the productive reset is blameless on our scoring side, which is precisely why the fix must bound the overrun rather than penalize or pace the peer.

Prototype context on `feat/node-ip-and-allowance-fixes`: the demo fast re-dispatch (`c700a8ac`) is the part worth keeping; the topology cooldown (`9f495880`) is dropped.

### LINKED ISSUE #568: accounting: commit the origin debit on a lost delivery, release only on not-found
## Problem

A chainless client is disconnected by serving storers mid-download. An instrumented live test plus the reference source confirm these are reference accounting-disconnects at the light line (1,687,500), driven by a debt-view desync, not stream or resource limits.

The reference charges us when it **serves** a chunk and keeps the charge through a connection reset. vertex reserves the per-chunk debit at dispatch but **releases it unbooked** when the delivery is lost (reset/timeout/cancel). So a lost-in-flight delivery is charged by the server and not by us: the server's view of our debt exceeds ours, it crosses its line while our admission band (reading our lower view) still admits, and it disconnects us, dropping more in-flight deliveries and cascading on the reference's ~4 second blocklist cycle.

## Evidence

- Instrumented 36 second gRPC download: committed debt to hot peers up to 1,220,000 (near the 1,687,500 line), 7 band Refuses, and 18 reservations released unbooked concentrated on the same peers that reset us; 111 whole-connection resets.
- Reference source: serve path calls `PrepareDebit` (shadow-reserved) before sending, then `debit.Apply()` after the muxer write returns (buffered, not received), and ghosts rather than rolls back on a failed send; its peer-debt view = balance + shadow-reserved + ghost, deliberately kept on the high side. It only prepares the debit after a store hit or successful forward, so a genuine not-found returns before charging and surfaces as a delivery-error: the only signal the server did not charge.

## Current behaviour

`retrieve_chunk` (`crates/swarm/node/src/client_service.rs`) applies the held reservation only on `result.is_ok()` and drops (releases) it on any failure. `push_chunk` has the same uploader-credits-storer asymmetry.

## Proposed

Mirror the reference's pessimistic booking: on a post-dispatch failure where bytes may have flowed (reset/timeout/cancel), **commit** the held reservation; release it **only** on a confirmed not-found / delivery-error.

- Needs a `ChunkTransferError` discriminant distinguishing "confirmed not-found" from "lost in flight" (`vertex-swarm-client-protocol`); the commit primitive `Reservation::apply` already exists.
- Mirror the change in `push_chunk`.
- The commit must feed **only the ledger, never `report_peer`** (#515): a reset caused by our own debt must not score down or back off the serving peer.
- Phantom debt (over-counting a delivery the server never served, i.e. a reset before it prepared the debit) is bounded to one chunk price per ambiguous reset, and is cheap under pseudosettle (time-refresh: it becomes surplus on the peer, not unclearable). Tighten the release predicate when SWAP is enabled, where cheques cost real value.

## Scope

- `crates/swarm/node/src/client_service.rs` (the `retrieve_chunk` and `push_chunk` failure arms)
- `crates/swarm/client-protocol` (the confirmed-not-found discriminant)

## Related

- Sibling of the closed #514 (which did the dispatch-reserve half; this does the failure half). Guarded by #515. Resets amplified by #519 (reducing them shrinks the desync source).

## AI Assistance

AI Assistance: Claude Code (Opus 4.8) used for the live diagnosis, the reference-source validation, and this issue.

mfw78:
## What

Book the per-chunk origin retrieval debit **at dispatch** instead of on delivery, so a chainless client's committed debt-view stays at or above the serving peer's at all times. This is what stops the mid-download neighbourhood collapse.

- `reserve_origin` commits the price the moment we send a request (gated by the admission band, so we never commit past the disconnect line), and returns the committed price.
- A request is **refunded** only when it provably reached no charge: a confirmed not-found, a not-connected peer, or a send that never left the node. Every other outcome (success, reset, timeout, cancel, and crucially a **dropped losing race leg**) keeps the commit.
- New `BandwidthDebit::refund_received` (the inverse of the dispatch commit, a pure ledger credit). Origin dispatch routes through the one-step `debit_received`, retiring the now-unused `BandwidthReserve`/`HeldReceive` reserve-and-hold path.

## Why

Closes #568. Closes #519. The reference server books our debt when it **serves** a chunk and keeps it through a connection reset; a client that books only on delivery under-counts a delivery lost in flight (the server charged, we did not), so the server's view leads ours over its disconnect line and it drops us, collapsing the neighbourhood.

Booking at dispatch gives the guarantee: every chunk the server serves corresponds to a request we already committed, so `server-view <= our-committed <= line`. Our band refuses before the server's line, so the server can never debt-disconnect us.

## Live validation (mainnet, controlled gRPC download)

| build | resets | routing depth |
|---|---|---|
| pre-fix baseline (~48s) | 111 | 8 -> 0 (collapsed) |
| commit-on-failure (first attempt, ~56s) | 165 | 8 -> 0 (still collapsed) |
| **book-at-send (sustained ~60s)** | **3** | **9, held the entire run (45/45 depth samples)** |

Depth never dropped under 60s of continuous download load, versus collapsing to 0 in both prior runs. An earlier commit-on-failure attempt fixed the desync but still collapsed (it released a dropped race leg's reservation); book-at-send commits synchronously before the await, so a dropped leg cannot un-book. That was the gap.

## Testing

- `cargo nextest run -p vertex-swarm-node -p vertex-swarm-client-protocol -p vertex-swarm-accounting --all-features`: 175 passed. New cases: dispatch books immediately, refund on not-found / not-connected / send-failure, commit retained on reset/timeout/cancel and on a **dropped race-loser future**, commit-and-refund are ledger-only, and an accounting-core `refund_received` inverse test.
- `cargo clippy ... --all-targets --all-features -- -D warnings` (incl. api, builder): clean. `cargo build --release -p vertex` and `cargo check --target wasm32-unknown-unknown`: clean.
- Adversarial QA: the `refund_received` sign is correct (`+price`, no double-debit), `record(_, Upload)` is a pure ledger op (no threshold/settlement/scoring side-effects), gating refuses before the line, no double-refund, phantom debt strictly in the safe (over-count) direction. Rust-idiomacy: collapsed onto the existing one-step `debit_received`, byte-identical semantics, vestigial traits retired.

## Notes

- **Ledger-only (the #515 guard):** commit and refund touch only the reserved balance; `OriginGate` carries no peer reporter, so a reset driven by our own debt never docks the serving peer.
- **Bounded phantom debt:** committing on ambiguous loss (cancelled race-losers, timeouts where the server never served) over-counts by at most one chunk price per event, in the safe direction, and is cheap under pseudosettle (time-refresh: it becomes surplus, not unclearable). Tighten the release predicate when SWAP is enabled, where cheques cost real value. The throttle this induces is the correct chainless pacing to the free rate.
- The durable fix that removes the desync entirely (fork-gated balance reconciliation, letting the free rate run to the true line) is #569. Builds on #514 (the dispatch-reserve groundwork).

## AI Assistance

AI Assistance: Claude Code (Opus 4.8) used for the live diagnosis, reference-source validation, implementation, adversarial QA and idiomacy review, the live mainnet validation, and this description.

### DIFF
--- a/crates/swarm/accounting/core/src/accounting/mod.rs
+++ b/crates/swarm/accounting/core/src/accounting/mod.rs
@@ -363,6 +363,29 @@ mod tests {
     }
 
     #[test]
+    fn test_refund_received_credits_back_a_committed_receive() {
+        // The inverse of the dispatch commit: commit a receive debit, then refund
+        // it and watch the balance return to zero. Reserved stays cleared.
+        use vertex_swarm_api::BandwidthDebit;
+
+        let accounting = test_accounting();
+
+        accounting
+            .prepare_receive(test_peer(), au(1000), true)
+            .expect("within threshold")
+            .apply();
+
+        let handle = accounting.for_peer(test_peer());
+        assert_eq!(handle.balance(), au(-1000));
+        assert_eq!(handle.state.reserved_balance(), au(0));
+
+        accounting.refund_received(test_peer(), au(1000));
+
+        assert_eq!(handle.balance(), au(0));
+        assert_eq!(handle.state.reserved_balance(), au(0));
+    }
+
+    #[test]
     fn test_prepare_receive_dropped() {
         let accounting = test_accounting();
 
--- a/crates/swarm/api/src/components/bandwidth.rs
+++ b/crates/swarm/api/src/components/bandwidth.rs
@@ -187,6 +187,11 @@ pub trait SwarmBandwidthAccounting: Send + Sync {
 pub trait BandwidthDebit: Send + Sync {
     /// Debit `peer` by `price` for a received chunk, committing immediately.
     fn debit_received(&self, peer: OverlayAddress, price: Au, originated: bool) -> SwarmResult<()>;
+
+    /// Credit back a receive debit committed at dispatch. Inverse of the
+    /// dispatch commit: the balance moves back, the reservation is already
+    /// cleared. Pure ledger op, never peer scoring.
+    fn refund_received(&self, peer: OverlayAddress, price: Au);
 }
 
 impl<B: SwarmBandwidthAccounting> BandwidthDebit for B {
@@ -194,56 +199,9 @@ impl<B: SwarmBandwidthAccounting> BandwidthDebit for B {
         self.prepare_receive(peer, price, originated)
             .map(Commit::apply)
     }
-}
 
-/// Object-safe held receive reservation: reserve at dispatch, commit on
-/// delivery, release on drop.
-///
-/// The origin retrieval and pushsync legs reserve the price before sending and
-/// carry the hold across the in-flight leg behind a trait object (the dispatch
-/// site cannot name the concrete `ReceiveAction`). Calling [`apply`](Self::apply)
-/// commits the debit; dropping the box releases the reservation, so any
-/// non-delivery exit (failure, timeout, a dropped request future) never leaks.
-pub trait HeldReceive: Send {
-    /// Commit the reserved receive debit.
-    fn apply(self: Box<Self>);
-}
-
-impl<T: Commit + 'static> HeldReceive for T {
-    fn apply(self: Box<Self>) {
-        Commit::apply(*self);
-    }
-}
-
-/// Object-safe reserve-at-dispatch for callers that hold accounting behind a
-/// trait object.
-///
-/// Reserves the receive leg and returns the hold to carry across the in-flight
-/// leg. An `Err` is the band refusing at the disconnect line (the same hard gate
-/// [`SwarmBandwidthAccounting::prepare_receive`] applies): sending would cross
-/// the creditor's line, so the caller routes elsewhere.
-pub trait BandwidthReserve: Send + Sync {
-    /// Reserve `price` for an in-flight receive leg from `peer`. Apply the
-    /// returned hold on delivery; drop it to release.
-    fn reserve_received(
-        &self,
-        peer: OverlayAddress,
-        price: Au,
-        originated: bool,
-    ) -> SwarmResult<Box<dyn HeldReceive>>;
-}
-
-impl<B: SwarmBandwidthAccounting> BandwidthReserve for B
-where
-    B::ReceiveAction: 'static,
-{
-    fn reserve_received(
-        &self,
-        peer: OverlayAddress,
-        price: Au,
-        originated: bool,
-    ) -> SwarmResult<Box<dyn HeldReceive>> {
-        Ok(Box::new(self.prepare_receive(peer, price, originated)?))
+    fn refund_received(&self, peer: OverlayAddress, price: Au) {
+        self.for_peer(peer).record(price, Direction::Upload);
     }
 }
 
--- a/crates/swarm/api/src/components/mod.rs
+++ b/crates/swarm/api/src/components/mod.rs
@@ -9,9 +9,9 @@ mod reserve;
 mod topology;
 
 pub use self::bandwidth::{
-    BandwidthDebit, BandwidthReserve, Commit, CommitOnWrite, Direction, HeldReceive,
-    SwarmAccountingConfig, SwarmBandwidthAccounting, SwarmClientAccounting, SwarmPeerBandwidth,
-    SwarmPeerState, SwarmSettlementProvider,
+    BandwidthDebit, Commit, CommitOnWrite, Direction, SwarmAccountingConfig,
+    SwarmBandwidthAccounting, SwarmClientAccounting, SwarmPeerBandwidth, SwarmPeerState,
+    SwarmSettlementProvider,
 };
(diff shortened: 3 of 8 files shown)


mfw78:
Live mainnet check (controlled gRPC download): this PR fixes the debt-view **desync** (our ledger now tracks the server's), but on its own it does **not** stop the neighbourhood collapse under sustained download. Depth still collapses 8 -> 0 with the fix in place, and resets are comparable to the pre-fix baseline.

Reason: commit-on-failure tightens *our* admission band for *future* requests, but it does not lower the server's view and does not slow our request rate, and the server already crossed its line at serve time for the in-flight deliveries before we detect the failure. So this is necessary accounting hygiene (correct ledger, validated by unit tests + adversarial QA), but the live collapse is driven by requesting faster than the free rate so debt climbs to the line regardless. That is the #519 axis (bound retrieval fan-out width/rate, protect bin-0), which must land alongside this for the client to actually hold the neighbourhood.

Recommend merging this on its accounting-correctness merits, but not treating it as the fix for the mid-download disconnects.

PR #594, opened 2026-06-30
perf(swarm-node): hold a peer off settle for the creditor rate-gate window

### LINKED ISSUE #593: accounting: pace per-peer settle retries to the creditor rate-gate window
## Problem

Under a bulk download the closest peers to each chunk gate first, and the admission band refuses the whole close set. Each banded retrieval request to a gated peer triggers a settlement. The creditor's pseudosettle rate gate works in whole Unix seconds, so a settle offered again in the same second as the last accepted one is answered with a zero-amount no-op ack: there is nothing new to forgive yet.

The in-flight dedup map keeps at most one settle per peer running, but a no-op settle completes almost instantly and clears the entry, so the next banded request immediately triggers another. Live mainnet instrumentation under bulk concurrency measured this re-trigger loop spawning settles at roughly 79k/s, each one a fresh task the creditor instantly rejects, spinning the single poll thread for no economic effect.

## Current

`AccountingSettlement::trigger_settlement` skips a peer only while a settle is already in flight. Because a no-op settle drains nothing and returns at once, the dedup gap is effectively zero and a flood of gated requests re-triggers a settle on every request.

## Proposed

Keep the at-most-one-in-flight rule, and after a settle finishes leave a one-second cooldown on the peer matching the creditor's whole-second gate. A fresh trigger inside the cooldown is suppressed; once the interval passes a genuinely-needed settle still fires. Only the no-op re-trigger spin is removed. The cooldown timestamp uses the wasm-safe monotonic clock, and expired cooldowns are pruned when a new settle starts so the map stays bounded.

## Scope

`crates/swarm/node/src/selection.rs`, the `AccountingSettlement` settlement trigger only. No wire change: settlement frames are unchanged, only the debtor-side retry pacing. Legitimate single-settle behaviour and the in-flight dedup are preserved.

## Notes

The cooldown is configurable through `with_min_settle_interval`; a zero gap restores the bare in-flight dedup and is used by the dedup test in isolation.

mfw78:
## What

Hold a peer off settlement for a one-second cooldown after each settle attempt, matching the creditor's whole-second pseudosettle rate gate. The at-most-one-settle-per-peer-in-flight rule is kept; on top of it, when a settle finishes the peer's map entry is left as a cooldown for the inter-attempt gap rather than cleared, so a flood of gated retrieval requests cannot re-trigger a no-op settle the creditor would only reject. A genuinely-needed settle still fires once the interval elapses. The cooldown uses the wasm-safe monotonic clock and expired entries are pruned when a new settle starts, so the map stays bounded.

## Why

Closes #593.

Under a bulk download the closest peers to each chunk gate first and the admission band refuses the whole close set, so every banded request triggers a settlement. The creditor's pseudosettle rate gate works in whole Unix seconds: a settle offered again in the same second is answered with a zero-amount no-op ack. The in-flight dedup alone does not pace these, because a no-op settle completes almost instantly and clears the entry, so the next request immediately re-triggers. Live mainnet instrumentation measured this re-trigger loop spawning settles at roughly 79k/s, each a fresh task the creditor instantly rejects, spinning the single poll thread for no economic effect. The cooldown collapses that spin to the rate the creditor would actually accept.

## Testing

`cargo fmt --all --check`, `cargo clippy --all-targets --all-features -- -D warnings`, `cargo test -p vertex-swarm-node -p vertex-swarm-accounting --all-features`, and `cargo build --target wasm32-unknown-unknown -p vertex-swarm-node` all clean.

The in-flight dedup test (`trigger_settlement_runs_one_settle_per_peer_in_flight`) still passes: the spam guard does not change single-settle behaviour. A new test (`trigger_is_suppressed_during_the_cooldown_then_fires_after_it`) asserts re-triggers inside the cooldown spawn no settle and that a settle fires again once the gap elapses, and the guard test (`in_flight_guard_leaves_a_cooldown_on_drop`) asserts the in-flight entry becomes a cooldown on drop, including on panic or cancellation.

## AI Assistance

Claude Code used for the live-mainnet diagnosis, the implementation, and the benchmark.

### DIFF
--- a/crates/swarm/node/src/selection.rs
+++ b/crates/swarm/node/src/selection.rs
@@ -38,12 +38,36 @@ use vertex_swarm_api::{
 use vertex_swarm_primitives::OverlayAddress;
 use vertex_swarm_topology::TopologyHandle;
 use vertex_tasks::TaskExecutor;
-use vertex_tasks::time::Instant;
+use vertex_tasks::time::{Duration, Instant};
 
 /// A shareable handle that resolves when an in-flight settle for a peer
 /// completes (the settle's [`InFlightGuard`] fires it on drop), so a waiter can
 /// re-evaluate admission as soon as the network settle acks.
 type SettleCompletion = Shared<oneshot::Receiver<()>>;
+
+/// Minimum wall-clock gap between settle attempts to one peer.
+///
+/// The creditor's pseudosettle rate gate works in whole Unix seconds: a settle
+/// offered in the same second as the last accepted one is answered with a
+/// zero-amount ack (nothing new to forgive yet). With no debtor-side gap a gated
+/// retrieval re-triggers a settle on every banded request, each spawning a task
+/// that the creditor instantly no-ops, spinning the single poll thread. Holding a
+/// peer off for one second after an attempt collapses that spin to the rate the
+/// creditor would actually accept, while a genuinely-needed settle still fires
+/// once the interval passes.
+const MIN_SETTLE_INTERVAL: Duration = Duration::from_secs(1);
+
+/// Per-peer settlement slot in the dedup map: a settle is running, or one
+/// finished recently and the peer is cooling down.
+enum SettleSlot {
+    /// A settle is in flight; its completion lets a waiter await the ack and a
+    /// concurrent trigger dedup against it.
+    InFlight(SettleCompletion),
+    /// The last settle to this peer finished at this instant. A fresh trigger is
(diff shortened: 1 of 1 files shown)


mfw78:
Superseded by #600 (epic #606, Phase 2), which removes the settle-spam at the source rather than pacing retries. Recommend closing this PR; see #593 for the full rationale and the one residual case worth re-checking against post-#600 metrics.

Issue #667, opened 2026-07-02
networking: tighten the behaviour hot paths (wake, queues, plumbing)


mfw78:
**Epic.** Tighten the libp2p behaviour hot paths: wake correctness, queue policy, and the collapse of mechanical event plumbing across the topology, client and storer behaviours. The dispatch engine above these paths was unified in #639; this epic is the layer below it. The survey found the per-message compute already lean (snapshot-cached `closest_to`, atomic depth, batch-drained event loop, lock-free handler polls) and the residual debt concentrated in plumbing: two drains that never wake, one drop policy that can eat a booked debit, four enum vocabularies wrapping one payload, three copies of peer identity, and three copies of the handler scaffold.

## Ground truth

Wake gaps: the topology poll drains gossip actions and peer lifecycle events (including bans) with `try_recv` and no waker, so enforcement waits for the next unrelated wake or the dial tick on a quiet node. Queue policy: the client behaviour's single drop-on-full policy treats consumer events and accounting-bearing commands alike; a dropped retrieve command leaves its dispatch-booked debit unrefunded by design. Plumbing: `ClientCommand -> HandlerCommand -> HandlerEvent -> ClientEvent` are 1:1 rewraps of shared payloads; overlay-to-PeerId truth lives in three places; the pending-queue and inbound-drain handler skeleton is hand-copied three times.

Defended as already right, not to be churned: `closest_to`'s generation-cached snapshot and top-k partition, the batch-drained node event loop, the hand-rolled client multiplexer's substream and timeout machinery (load-bearing per `crates/swarm/AGENTS.md`), and wakerless `pending_events` pops (the standard libp2p callback pattern).

## Phases

- [x] #668 register wakers for the gossip and lifecycle drains
- [x] #669 never silently drop accounting-bearing commands
- [x] #670 collapse the layered event and command vocabularies (coordinate #620)
- [x] #671 share one peer identity view across behaviours (absorbs #320; verify it against the post-#654 wiring)
- [x] #672 extract the shared handler scaffold as building blocks (multiplexer untouched)
- [x] #673 profile then sample the per-poll behaviour timing metrics
- [x] hardening (existing issues): #573 connection caps, #241 hive multiaddr cap, #126 rate-limit inbound ECDSA-bearing messages

[...]

PR #717, opened 2026-07-02
refactor(swarm-client-behaviour,swarm-client-protocol): collapse the layered command and event vocabularies

### LINKED ISSUE #670: client-behaviour: collapse the layered event and command vocabularies
Part of #667. Builds on the wake and queue fixes.

## Problem

One request crosses four enum vocabularies (`ClientCommand`, `HandlerCommand`, `HandlerEvent`, `ClientEvent`) with roughly two hundred lines of mechanical match-and-rewrap, so every behavioural change lands in four places.

## Current

`crates/swarm/client-behaviour/src/behaviour.rs` (`on_command`, the handler-event mapping) and `handler.rs` (`on_inbound_outcome`) are 1:1 wrapping layers over shared payloads.

## Proposed

Shared payload structs carried by every layer with `From` impls doing the wrapping once; keep the distinct enums where the layers genuinely diverge (per-connection versus per-overlay addressing), delete the arms that only relabel.

## Scope

`crates/swarm/client-behaviour/src/{behaviour.rs,handler.rs}`, `vertex-swarm-client-protocol`. No wire change, no handler-multiplexer reshape.

## Notes

Respects the rule that the multiplexer stays hand-rolled; this is vocabulary, not substream machinery. Coordinate with #620, whose per-operation error types touch the same enums.

## AI Assistance

AI Assistance: Claude Code (Fable 5) used for the analysis and drafting this issue.

mfw78:
## What

Collapse the four enum vocabularies a single client request crossed into shared payloads defined once.

`vertex-swarm-client-protocol` now defines `PeerEvent` and `PeerCommand`, the per-peer protocol payloads. The four outer enums keep only the variants where the layers genuinely diverge: `ClientEvent` becomes `{ Peer { peer, peer_id, event }, PeerActivated, PeerDisconnected, ProtocolError }`, `ClientCommand` becomes `{ ActivatePeer, Peer { peer, command } }`, and the handler's `HandlerEvent`/`HandlerCommand` shrink to their lifecycle plus a `Peer` envelope. `on_command` drops to two arms, `on_handler_event` to three, and `on_inbound_outcome` becomes a `From<InboundOutcome>` conversion. The wire-to-domain payment-ack conversions move from the behaviour into the handler, so the wire `PaymentAck` type is now handler-confined.

## Why

Closes #670 (part of epic #667). One request crossed four vocabularies with roughly 350 lines of mechanical match-and-rewrap and duplicated variant definitions, so every behavioural change had to land in four places. Sharing the payload removes the relabel arms and the duplication.

Behaviour-preserving: the same events are emitted, the two cap-exemption sets are preserved verbatim and deliberately kept distinct (they are not homogenized), the refusal check order is unchanged (capacity before the peer lookup), the settlement tee still runs for every pseudosettle and swap signal, and all metrics and operator-facing log lines are carried over. The only observability delta is that a handful of per-command success and unknown-peer debug logs fold into one generic drop log. No wire change: the codecs, upgrades, and protocol IDs are untouched.

## Testing

- `cargo clippy --all-targets --all-features -- -D warnings` (full workspace, at implementation time)
- `cargo nextest run` across the six touched crates, all features (224/224), including new tests pinning the cap-exemption set through the new envelope, the refusal order, and the moved ack round-trip
- The `behaviour_tests` and `timeout_repro` libp2p-swarm integration suites pass with construction-only changes (the no-lost-wakeup and no-lost-event proof)
- `just check-cone` and `cargo build --target wasm32-unknown-unknown` for `vertex-swarm-client-behaviour` and `vertex-swarm-node`

## Notes

AI Assistance: Claude Code (Fable 5 for design and QA, Opus 4.8 for implementation) used for the refactor and this description.


crates/swarm/AGENTS.md
```diff
@@ -16,8 +16,8 @@ Root `/AGENTS.md` applies here too (multiaddrs-not-underlay, no inline reference
...
+- `client-protocol`: the command and event contract for the client behaviour. A request or signal crosses one shared per-peer payload, not four parallel vocabularies: `PeerEvent` and `PeerCommand` carry the 17 per-peer signals and the per-peer commands with no addressing field, and the outer `ClientEvent`/`ClientCommand` add only what a layer genuinely needs (`ClientEvent::Peer { peer, peer_id, event }` enriches with the libp2p identity; `ClientCommand::Peer { peer, command }` addresses an overlay; `ClientEvent::{PeerActivated,PeerDisconnected,ProtocolError}` and `ClientCommand::ActivatePeer` stay distinct). Also `FailureKind`, the `RetrievalResponseTx`/`PushResponseTx` channels, `ChunkTransferError`, `RetrievalResult`, the domain `PseudosettleAck`, and the extracted `PseudosettleEvent`/`SwapEvent`. Sits below both `node` and the settlement crates so neither depends up on the other: `node` re-exports every item from its original paths, `accounting/{pseudosettle,swap}` import the contract directly. Pure data, no `NetworkBehaviour` or `Swarm`; uses libp2p only for `PeerId` in the `ClientEvent` envelope.
```
mfw78, line 19: IMO, an `overlay` should never be assigned to a variable `peer` and the variable `peer` should be reserved for types libp2p::PeerId, and an overlay should bare the variable name `overlay` (we should be able to reason about it easily when reading the code)

